PR #146, opened 2016-08-25
Add access with apikey

Rikuoja:
Thorough checking should be done because of the refactoring of permission validation I had to do to incorporate API key support.

As per DRF conventions, permission validation should take place in Serializer, not in the Viewset perform_create or perform_update methods.


events/api.py
```diff
@@ -718,30 +726,52 @@ class EventSerializer(LinkedEventsSerializer, GeoModelAPIView):
     publication_status = EnumChoiceField(PUBLICATION_STATUSES)
     external_links = EventLinkSerializer(many=True, required=False)
     offers = OfferSerializer(many=True, required=False)
+    id = serializers.CharField(required=False)
```
juyrjola: As a convention, the `id` field should be placed as the first field in the class.

events/api.py
```diff
@@ -718,30 +726,52 @@ class EventSerializer(LinkedEventsSerializer, GeoModelAPIView):
...
 
     def __init__(self, *args, skip_empties=False, **kwargs):
         super(EventSerializer, self).__init__(*args, **kwargs)
         # The following can be used when serializing when
         # testing and debugging.
         self.skip_empties = skip_empties
 
+        # for post and put methods, user information is needed to restrict permissions at validate
+        self.method = self.context['request'].method
+        if self.context['request'].method in ['PUT', 'POST']:
```
juyrjola: `self.method` didn't seem to be used anywhere.

It's better to architect the code like this:
```
from rest_framework.permissions import SAFE_METHODS

if self.method in SAFE_METHODS:
    return
...
```

events/api.py
```diff
@@ -718,30 +726,52 @@ class EventSerializer(LinkedEventsSerializer, GeoModelAPIView):
...
+            self.user = self.context['request'].user
+            self.api_key = self.context['request'].query_params.get('api_key')
```
juyrjola: This doesn't seem like the sort of stuff that should live in `EventSerializer`. Maybe store the auth information in `request.auth`?

events/api.py
```diff
@@ -718,30 +726,52 @@ class EventSerializer(LinkedEventsSerializer, GeoModelAPIView):
...
+
+            # api_key takes precedence over user
+            if isinstance(self.user, ApiKeyUser):
+                self.data_source = self.context['request'].auth['authenticated_data_source']
```
juyrjola: Oh, looks like you're doing it already. :)

events/api.py
tuomas777: It seems that this allows modifying existing object's ID, which most probably is a bad idea.

events/api.py
```diff
@@ -718,30 +726,52 @@ class EventSerializer(LinkedEventsSerializer, GeoModelAPIView):
...
         self.skip_empties = skip_empties
 
+        # for post and put methods, user information is needed to restrict permissions at validate
+        self.method = self.context['request'].method
+        if self.context['request'].method in ['PUT', 'POST']:
+            self.user = self.context['request'].user
+            self.api_key = self.context['request'].query_params.get('api_key')
+
+            # api_key takes precedence over user
+            if isinstance(self.user, ApiKeyUser):
```
juyrjola: Check `isinstance(self.context['request'].auth, ...)` instead

events/api.py
```diff
@@ -769,6 +799,33 @@ def to_internal_value(self, data):
...
+        # validate id permissions
+        if 'id' in data:
+            if not data['id'].split(':', 1)[0] == self.data_source.id:
+                raise serializers.ValidationError(
+                    {'id': "Setting id to " + str(data['id']) +
+                           " is not allowed for your organization. The id"
+                           " must be left blank or set to " + str(self.data_source) + ":desired_id"})
+        # validate data source permissions
+        if 'data_source' in data:
+            if data['data_source'] != self.data_source:
```
juyrjola: Implement a test to expose this problem

events/api.py
```diff
@@ -769,6 +799,33 @@ def to_internal_value(self, data):
...
+                raise serializers.ValidationError(
+                    {'data_source': "Setting data_source to " + str(data['data_source']) +
```
juyrjola: Add translations to error messages

events/api.py
```diff
@@ -777,7 +843,9 @@ def validate(self, data):
 
         # require the publication status
         if 'publication_status' not in data:
-            raise serializers.ValidationError({'publication_status': 'You must specify whether you wish to submit a draft or a public event.'})
+            raise serializers.ValidationError({'publication_status':
+                                                   _(
+                                                       "You must specify whether you wish to submit a draft or a public event.")})
```
tuomas777: Formatting looks weird.

Rikuoja: Oh, it does. PyCharm apparently does not know how to implement PEP 8 recommendations here.

events/api.py
```diff
@@ -850,16 +918,25 @@ def validate_event_status(self, value):
...
             raise serializers.ValidationError(_('POSTPONED and RESCHEDULED statuses cannot be set directly.'
-                                    'Changing event start_time or marking start_time null'
-                                    'will reschedule or postpone an event.'))
+                                                'Changing event start_time or marking start_time null'
+                                                'will reschedule or postpone an event.'))
 
     def create(self, validated_data):
+        # if id was not provided, we generate it upon creation:
+        if 'id' not in validated_data:
+            validated_data['id'] = generate_id(self.data_source)
```
tuomas777: What if an already existing `id` is provided, will it result in HTTP 500?

events/models.py
```diff
@@ -57,6 +57,8 @@ class Meta:
 class DataSource(models.Model):
     id = models.CharField(max_length=100, primary_key=True)
     name = models.CharField(verbose_name=_('Name'), max_length=255)
+    api_key = models.CharField(max_length=128, null=True, blank=True)
```
tuomas777: Django recommends avoiding `null=True` on `CharField`.

events/api.py
Rikuoja: A good point. Will add an error that yields HTTP 400.

events/models.py
Rikuoja: An excellent point.

events/api.py
```diff
@@ -871,14 +959,27 @@ def create(self, validated_data):
...
 
+        validated_data['last_modified_by'] = self.user
         # Update event_status if a PUBLIC SCHEDULED or CANCELLED event start_time is updated.
         # DRAFT events will remain SCHEDULED up to publication.
         # Check that the event is not explicitly CANCELLED at the same time.
         if (instance.publication_status == PublicationStatus.PUBLIC and
                     validated_data.get('event_status', Event.Status.SCHEDULED) != Event.Status.CANCELLED):
+            # if the instance was ever RESCHEDULED or POSTPONED, it may never be SCHEDULED again
+            if instance.event_status != Event.Status.SCHEDULED:
+                validated_data['event_status'] = instance.event_status
```
juyrjola: I think it's better to fail with an error instead of silently fixing the data.

events/api.py
Rikuoja: Fixed.

events/api.py
Rikuoja: Fixed.

events/tests/test_event_put.py
```diff
@@ -148,6 +175,72 @@ def test__cancel_an_event_with_put(api_client, complex_event_dict, user):
...
+
+    # remove the start_time
+    data2 = response.data
+    data2['start_time'] = None
+
+    # update the event
+    event_id = data2.pop('@id')
+    response2 = update_with_put(api_client, event_id, data2)
+    print('updated the event')
+    print(response2.data)
```
juyrjola: leftover prints?

Rikuoja: Removed.

events/api.py
Rikuoja: Fixed.

PR #153, opened 2016-09-02
Remove DocumentedRouter

### LINKED ISSUE #143: Refactor DocumentedRouter for DRF 3.4 compatibility

tuomas777:
A template containing the base description is used instead. Own
function to change the root title from `Api Root` is needed also.

Closes #143


Rikuoja:
I would add to the PR events/templates/rest_framework/api.html file with the contents
```{% include "rest_framework/description.html" %}```. This way, when a new user installs linkedevents, he will get the standard LE description even if he has installed the Helsinki template.

The Helsinki template was added separately exactly because we want the code to be reusable, not Helsinki specific, so any non-Helsinki specific documentation should be visible without the Helsinki template.

tuomas777:
Yeah, good point.

tuomas777:
Fixed

PR #154, opened 2016-09-02
Add field photographer_name to Image

### LINKED ISSUE #150: Add photographer name field in Image model
Required by the CC BY 4.0 license.

tuomas777:
Closes #150


events/models.py
```diff
@@ -110,6 +110,7 @@ class Image(models.Model):
     url = models.URLField(verbose_name=_('Image'), max_length=400, null=True, blank=True)
     cropping = ImageRatioField('image', '800x800', verbose_name=_('Cropping'))
     license = models.ForeignKey(License, verbose_name=_('License'), related_name='images', default='cc_by')
+    photographer_name = models.CharField(verbose_name=_('Photographer name'), max_length=255, blank=True)
```
juyrjola: Undefined `photographer_name` fields should be exposed as `null`s over the API.

tuomas777: added `null=True` to be consistent with existing `CharField`s

PR #163, opened 2016-11-14
Bulk operations

### LINKED ISSUE #125: Allow POSTing repeating events with superevents
Required by UI issue #126.

tuomas777:
* Allow bulk POST and PUT in event endpoint
* Exclude recurring super events that don't have subevents

Closes #125


Rikuoja:
Looking good, but let's make extra tests for POSTin subevents to a superevent just to be on the safe side.

events/api.py
```diff
@@ -799,6 +799,7 @@ class EventSerializer(LinkedEventsSerializer, GeoModelAPIView):
                                      view_name='language-detail', many=True, queryset=Language.objects.all())
     audience = JSONLDRelatedField(serializer=KeywordSerializer, view_name='keyword-detail',
                                   many=True, required=False, queryset=Keyword.objects.all())
+    is_recurring_super = serializers.BooleanField(required=False)
```
Rikuoja: Actually, we decided to make this field more usable. It will allow using the superevents for different purposes, i.e. changing the event model field to something like
```
-    is_recurring_super = models.BooleanField(default=False)
+    super_event_type = models.CharField(null=True, blank=True, choices=SUPER_EVENT_TYPES)
``` 
with
```
+    SUPER_EVENT_TYPES = (
+        ('recurring', _('Recurring')),
+        # Other types include e.g. a festival
+    )
```

Rikuoja:
Excellent! Another feature request while we're at it :D Now that we have super_event_type available in the api, it would be more than beneficial to be able to filter by the field. I'm talking api calls of the type `/v1/event/?super_event_type=none` , `/v1/event/?super_event_type=recurring` and whatever else comes along?

PR #164, opened 2016-11-30
Allow images to be posted with an api_key as well

Rikuoja:


events/api.py
```diff
@@ -385,6 +385,27 @@ def __init__(self, instance=None, files=None,
             if 'disable_camelcase' in request.query_params:
                 self.disable_camelcase = True
 
+        # for post and put methods, user information is needed to restrict permissions at validate
+        if context is not None:
```
juyrjola: It's better to do:
```python
if context is None:
    return
```

PR #200, opened 2017-03-16
Expand management command to only update changed event numbers

Rikuoja:
When events are changed, only "n_events_changed" in Place and Keywords are touched. The management command will do the actual updates.


events/models.py
```diff
@@ -488,10 +493,14 @@ def save(self, *args, **kwargs):
         super(Event, self).save(*args, **kwargs)
 
         # needed to cache location event numbers
-        if self.location:
-            recache_n_events_in_location(self.location)
+        if not old_location and self.location:
+            self.location.n_events_changed = True
+            self.location.save(update_fields=("n_events_changed",))
```
akx: This will cause that position/divisions save again.

events/models.py
```diff
@@ -488,10 +493,14 @@ def save(self, *args, **kwargs):
...
         if old_location and old_location != self.location:
-            recache_n_events_in_location(old_location)
+            old_location.n_events_changed = True
+            old_location.save(update_fields=("n_events_changed",))
```
akx: Ditto.

events/models.py
```diff
@@ -488,10 +493,14 @@ def save(self, *args, **kwargs):
...
+            self.location.n_events_changed = True
+            self.location.save(update_fields=("n_events_changed",))
```
akx: Ditto.

events/management/commands/update_n_events.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def add_arguments(self, parser):
+        parser.add_argument('model', nargs='?', default=False)
+        parser.add_argument('--all',
+                            default=False,
+                            action='store_true',
+                            dest='all',
+                            help='Recalculate everything from scratch')
+
+    def handle(self, *args, **kwargs):
```
akx: Just unload `model` and `all` here in the named args. No need to bother with `kwargs.get()`.

events/management/commands/update_n_events.py
```diff
@@ -0,0 +1,40 @@
...
+        model = kwargs.get('model')
+        if model and model not in self.models:
+            raise CommandError("Model %s not found. Valid models: %s" % (model, self.models))
+        if not model or self.models[model] is Keyword:
+            if kwargs.get('all'):
```
akx: Refactor this if body into `self.handle_keywords(all)`.

events/management/commands/update_n_events.py
```diff
@@ -0,0 +1,40 @@
...
+                keywords = Keyword.objects.all()
+            else:
+                keywords = Keyword.objects.filter(n_events_changed=True)
+            recache_n_events(keywords)
+            print("Updated %s keyword event numbers." % ('all' if kwargs.get('all') else 'changed'))
+            print("A total of %s keywords updated." % (str(keywords.count())))
+        if not model or self.models[model] is Place:
+            if kwargs.get('all'):
```
akx: Refactor this if body into `self.handle_places(all)`.

events/models.py
```diff
@@ -255,12 +256,12 @@ def recache_n_events(keywords):
...
-        n_events = count_map.get(keyword.id, 0)
-        if n_events != keyword.n_events:
-            keyword.n_events = n_events
-            keyword.save(update_fields=("n_events",))
+    if not keywords:
+        return
+    with transaction.atomic():
+        keywords.update(n_events=0)
+        for keyword_id, n_events in count_events_for_keywords((k.id for k in keywords)).items():
+            keywords.filter(id=keyword_id).exclude(n_events=n_events).update(n_events=n_events)
```
akx: Add `, n_events_changed=False)` to the update? :)

events/models.py
```diff
@@ -343,17 +345,20 @@ def save(self, *args, **kwargs):
...
-        # Use the queryset update API to skip the position/divisions save
-        Place.objects.filter(id=place.id).update(n_events=n_events)
+    # The helper function has to exist outside the model, because it is used in migration.
+    # Django apps cannot serialize unbound instance functions when saving model history during migration.
+    for place in locations:
+        n_events = place.events.count()
+        if n_events != place.n_events:
+            place.n_events = n_events
+            # Use the queryset update API to skip the position/divisions save
+            Place.objects.filter(id=place.id).update(n_events=n_events)
```
akx: Add `, n_events_changed=False)` to the update? :)

events/management/commands/update_n_events.py
```diff
@@ -0,0 +1,47 @@
...
+    def handle_places(self, update_all=False):
+        if update_all:
+            places = Place.objects.all()
+        else:
+            places = Place.objects.filter(n_events_changed=True)
+        recache_n_events_in_locations(places)
+        print("Updated %s place event numbers." % ('all' if update_all else 'changed'))
+        print("A total of %s places updated." % (str(places.count())))
+
+    def handle(self, *args, model=None, update_all=False, **kwargs):
```
akx: Since you don't actually handle `args`, don't allow them here either. `(self, model, update_all, **kwargs)` is enough.

events/management/commands/update_n_events.py
```diff
@@ -0,0 +1,47 @@
+from django.core.management import BaseCommand, CommandError
+from django.db import transaction
+
+from events.models import Keyword, Place
+from events.utils import recache_n_events_in_locations, recache_n_events
+
+
+class Command(BaseCommand):
+    help = "Update keyword and place event numbers"
+    models = {'keyword': Keyword,
```
akx: The models dict is unnecessary complexity now.

events/management/commands/update_n_events.py
```diff
@@ -0,0 +1,47 @@
...
+        else:
+            places = Place.objects.filter(n_events_changed=True)
+        recache_n_events_in_locations(places)
+        print("Updated %s place event numbers." % ('all' if update_all else 'changed'))
+        print("A total of %s places updated." % (str(places.count())))
+
+    def handle(self, *args, model=None, update_all=False, **kwargs):
+        if model and model not in self.models:
+            raise CommandError("Model %s not found. Valid models: %s" % (model, self.models))
+        if not model or self.models[model] is Keyword:
```
akx: Just `if not model or model == 'keyword':`

events/management/commands/update_n_events.py
```diff
@@ -0,0 +1,47 @@
...
+            self.handle_keywords(update_all=update_all)
+        if not model or self.models[model] is Place:
```
akx: Just `if not model or model == 'place':`

events/management/commands/update_n_events.py
```diff
@@ -0,0 +1,47 @@
...
+        if update_all:
+            places = Place.objects.all()
+        else:
+            places = Place.objects.filter(n_events_changed=True)
+        recache_n_events_in_locations(places)
+        print("Updated %s place event numbers." % ('all' if update_all else 'changed'))
+        print("A total of %s places updated." % (str(places.count())))
+
+    def handle(self, *args, model=None, update_all=False, **kwargs):
+        if model and model not in self.models:
```
akx: `if model and model not in ('keyword', 'place')` if you get rid of `self.models`.

events/models.py
```diff
@@ -488,10 +459,11 @@ def save(self, *args, **kwargs):
         super(Event, self).save(*args, **kwargs)
 
         # needed to cache location event numbers
-        if self.location:
-            recache_n_events_in_location(self.location)
+        if not old_location and self.location:
+            Place.objects.filter(id=self.location.id).update(n_events_changed=True)
         if old_location and old_location != self.location:
-            recache_n_events_in_location(old_location)
+            Place.objects.filter(id=old_location.id).update(n_events_changed=True)
```
akx: `Place.objects.filter(id__in=(old_location.id, self.location.id))...` to only use 1 query.

events/utils.py
```diff
@@ -38,3 +44,38 @@ def update(d, u):
...
+    """
+
+    # The helper function has to exist outside the model, because it is used in migration.
+    # Django apps cannot serialize unbound instance functions when saving model history during migration.
+    if not keywords:
+        return
+    with transaction.atomic():
+        keywords.update(n_events=0)
+        for keyword_id, n_events in count_events_for_keywords((k.id for k in keywords)).items():
+            keywords.filter(id=keyword_id).exclude(n_events=n_events).update(n_events=n_events,
```
akx: Hmm, now that you're also updating the changed flag, take the `exclude` clause out.

akx: Also, it's not a good idea to use `keywords.filter(id=keyword_id)`, since `keywords` could be a queryset that has 1000 other filter clauses already (and you'll end up making an SQL query that looks like `UPDATE ... WHERE id=... AND 1000-other-clauses`.

Just use `Keyword.objects.filter()`.

events/utils.py
```diff
@@ -38,3 +44,38 @@ def update(d, u):
         else:
             d[k] = u[k]
     return d
+
+
+def recache_n_events(keywords):
+    """
+    Recache the number of events for the given keywords.
+
+    :type keywords: Iterable[Keyword]
```
akx: Since there's `keywords.update()` there now, `Iterable[Keyword]` is a lie; it has to be a `QuerySet[Keyword]`.

akx: I'd honestly suggest changing the signature of the function to `keyword_ids` (and obviously how the function works as well), so you never have to wrangle querysets or actual keyword objects within here, since you don't need them at all.

That way the function would be

```python
def recache_n_events(keyword_ids):
    """
    Recache the number of events for the given keywords (by ID).
    :type keywords: Iterable[str]
    """

    if not keywords:
        return

    with transaction.atomic():
        Keyword.objects.filter(id__in=keyword_ids).update(n_events=0)
        for keyword_id, n_events in count_events_for_keywords(keyword_ids).items():
            Keyword.objects.filter(id=keyword_id).update(n_events=n_events, n_events_changed=False)
```
and the caller would be free to pass in a `.values_list('id', flat=True)` or a set of keyword ids from a cache or whatever without having to worry about loading the full objects -- and if they _do_ have keyword objects, they can pass in `(k.id for k in keyword)`.

events/management/commands/update_n_events.py
Rikuoja: How exactly?

events/management/commands/update_n_events.py
```diff
@@ -0,0 +1,45 @@
...
+            places = Place.objects.all()
+        else:
+            places = Place.objects.filter(n_events_changed=True)
+        recache_n_events_in_locations((k.id for k in places), all=update_all)
+        print("Updated %s place event numbers." % ('all' if update_all else 'changed'))
+        print("A total of %s places updated." % (str(places.count())))
+
+    def handle(self, model=None, update_all=False, **kwargs):
+        if model and model not in ('keyword', 'place'):
+            raise CommandError("Model %s not found. Valid models: %s" % (model, str('keyword', 'place')))
```
akx: I think `str('keyword', 'place')` doesn't do what you expect it to. :)

events/sql.py
```diff
@@ -1,11 +1,12 @@
...
 
 
-def count_events_for_keywords(keyword_ids=()):
+def count_events_for_keywords(keyword_ids=(), all=False):
     """
     Get the actual count of events using the given keywords.
 
-    :param keyword_ids: set of keyword ids; pass an empty set to get the data for all keywords
+    :param keyword_ids: set of keyword ids
+    :param all: count all keywords instead
```
akx: This should be after `:type keyword_ids:` (and it's missing a `:type:` clause too).

events/sql.py
```diff
@@ -34,4 +35,41 @@ def count_events_for_keywords(keyword_ids=()):
...
+    """
+    # sorry for the non-DRY-ness; would be easier with an SQL generator like SQLAlchemy, but...
+
+    place_ids = tuple(set(place_ids))
+    with connection.cursor() as cursor:
+        if place_ids:
+            cursor.execute('''
+            SELECT t.location_id, COUNT(DISTINCT t.id)
+            FROM (
+              SELECT location_id, id FROM events_event WHERE location_id IN %s
```
akx: Place event counts don't need a subquery, since they need to pick up data from a single table only.

```sql
SELECT e.location_id, COUNT(*)
FROM events_event e
-- possible where clause here
GROUP BY e.location_id;
```

events/utils.py
```diff
@@ -38,3 +44,42 @@ def update(d, u):
...
+
+def recache_n_events(keyword_ids, all=False):
+    """
+    Recache the number of events for the given keywords (by ID).
+
+    :param all: recache all keywords instead
+    :type keyword_ids: Iterable[str]
+    """
+
+    # needed so we don't empty the blasted iterator
```
akx: Well, you _do_ empty the iterator...

Rikuoja: Where is the facepalm emoji when you need one?

Rikuoja: Thank you.

Issue #204, opened 2017-03-28
Allow to filter by multiple publishers


hugovk:
I'd like to be able to fetch events from 11 different publishers, and ideally would like to do it in a single query. 

For example, single queries:
* https://api.hel.fi/linkedevents/v1/event/?publisher=ahjo:46104 52 results
* https://api.hel.fi/linkedevents/v1/event/?publisher=ahjo:46111 21 results

But these don't work:
* https://api.hel.fi/linkedevents/v1/event/?publisher=ahjo:46104,ahjo:46111 0 results
* https://api.hel.fi/linkedevents/v1/event/?publisher=ahjo:46104+ahjo:46111 0 results
* https://api.hel.fi/linkedevents/v1/event/?publisher=ahjo:46104&publisher=ahjo:46111 21 results

---

If we check the [source](https://github.com/City-of-Helsinki/linkedevents/blob/master/events/api.py#L1363-L1365), we can see that `_filter_event_queryset()` supports only one publisher: 

```python
    val = params.get('publisher', None)
    if val:
        queryset = queryset.filter(publisher__id=val)
```

However filtering by `data_source` [supports multiple data sources](https://github.com/City-of-Helsinki/linkedevents/blob/master/events/api.py#L1318-L1322
):
```python
    # Filter by data source, multiple sources separated by comma
    val = params.get('data_source', None)
    if val:
        val = val.split(',')
        queryset = queryset.filter(data_source_id__in=val)
```

So it looks like changing `publisher` to accept multiple values should be an easy task and shouldn't affect current behaviour.

Thank you!

PR #205, opened 2017-03-28
Allow to filter by multiple publishers

hugovk:
Resolves https://github.com/City-of-Helsinki/linkedevents/issues/204

Allow to filter by multiple comma-separated `publisher` parameters, in the same way as `data_source`.

Note: untested, but CI passes.

### DIFF
--- a/events/api.py
+++ b/events/api.py
@@ -1360,8 +1360,10 @@ def _filter_event_queryset(queryset, params, srs=None):
         cond = 'end_time - start_time >= %s :: interval'
         queryset = queryset.extra(where=[cond], params=[str(dur)])
 
+    # Filter by publisher, multiple sources separated by comma
     val = params.get('publisher', None)
     if val:
+        val = val.split(',')
         queryset = queryset.filter(publisher__id=val)
 
     return queryset


juyrjola:
Looks good, thanks!

hugovk:
Great! Do you know when this will likely be deployed?

eg. https://api.hel.fi/linkedevents/v1/event/?publisher=ahjo:46104,ahjo:46111

events/api.py
```diff
@@ -1360,8 +1360,10 @@ def _filter_event_queryset(queryset, params, srs=None):
         cond = 'end_time - start_time >= %s :: interval'
         queryset = queryset.extra(where=[cond], params=[str(dur)])
 
+    # Filter by publisher, multiple sources separated by comma
     val = params.get('publisher', None)
     if val:
+        val = val.split(',')
         queryset = queryset.filter(publisher__id=val)
```
akx: Shouldn't this be `publisher_id__in` now?

juyrjola:
There was a small bug in the patch, but I fixed that now and deployed it to production.

events/api.py
juyrjola: Yup, it should.

akx:
```python
>>> from events.models import Event
>>> qs = Event.objects.filter(publisher__id=['ee', 'qq', 'yy'])
>>> print(qs.query)
SELECT LOTS_OF_COLUMNS FROM "events_event" WHERE "events_event"."publisher_id" = ['ee', 'qq', 'yy'] ORDER BY "events_event"."tree_id" ASC, "events_event"."lft" ASC
>>> qs = Event.objects.filter(publisher_id__in=['ee', 'qq', 'yy'])
>>> print(qs.query)
SELECT LOTS_OF_COLUMNS FROM "events_event" WHERE "events_event"."publisher_id" IN (ee, qq, yy) ORDER BY "events_event"."tree_id" ASC, "events_event"."lft" ASC
```

hugovk:
Thanks all for the quick response!

PR #206, opened 2017-03-28
Add coverage tracking configuration to .travis.yml

akx:


akx:
Coverage doesn't look all that good at present, but hey, the only way is up from here :)

PR #223, opened 2017-08-28
Added support for creating, updating and deleting places thru API.

hkurhinen:
Added support for creating, updating and deleting places thru API.
Fixed parsing geometric shapes when creating places.


Rikuoja:
An excellent feature, we haven't gotten around to implementing this yet so it should make a good addition.

However, a few observations:
1) Noticing your need for a separate `create_keyword` method, I just made a PR #226 so that keywords would have the same base fields as the other models. This means the separate method should no longer be needed once PR #226 is merged.
2) The keyword and place ids are namespaced. Therefore, to prevent non-namespaced ids to be inserted, I suggest you add a `validate_id` method to the place and keyword serializers, similar to the one in the event serializer: https://github.com/City-of-Helsinki/linkedevents/blob/master/events/api.py#L999
3) Tests should be added for creating and editing keywords and places. We are currently trying to get the test coverage up :)
4) As the `update` and `create` (and perhaps `validate_id` too?) methods will be unnecessarily duplicated in all serializers, I suggest you refactor the code to have common elements in the common superclass for events, places, images and keywords.

Hope this will help other cities get up and running with Linkedevents!

.gitignore
```diff
@@ -13,3 +13,4 @@ media/
 *.iml
 *.kate-swp
 data/
+venv
```
akx: File missing last newline. (You might want to check if your editor supports auto-adding it?)

events/api.py
```diff
@@ -529,6 +529,23 @@ def create(self, validated_data):
...
+            self.user = None
+        validated_data['created_by'] = self.user
+        validated_data['last_modified_by'] = self.user
+        try:
+            instance = super().create(validated_data)
+        except IntegrityError as error:
+            if 'duplicate' and 'pkey' in str(error):
+                raise serializers.ValidationError({'id':_("An object with given id already exists.")})
+            else:
+                raise error
```
akx: Use a bare `raise` instead of `raise error`, so you don't lose the stacktrace.

events/api.py
```diff
@@ -529,6 +529,23 @@ def create(self, validated_data):
...
+        # no django user exists for the api key
+        if isinstance(self.user, ApiKeyUser):
+            self.user = None
+        validated_data['created_by'] = self.user
+        validated_data['last_modified_by'] = self.user
+        try:
+            instance = super().create(validated_data)
+        except IntegrityError as error:
+            if 'duplicate' and 'pkey' in str(error):
+                raise serializers.ValidationError({'id':_("An object with given id already exists.")})
```
akx: You might want to use  `raise ... from error` so the root cause of the error is not lost.

events/api.py
```diff
@@ -529,6 +529,23 @@ def create(self, validated_data):
...
+    def create_keyword(self, validated_data):
+        if 'data_source' not in validated_data:
+            validated_data['data_source'] = self.data_source
+        # no django user exists for the api key
+        if isinstance(self.user, ApiKeyUser):
+            self.user = None
+        validated_data['created_by'] = self.user
+        validated_data['last_modified_by'] = self.user
+        try:
+            instance = super().create(validated_data)
```
akx: Just `return super()...` instead of storing into a local you immediately return.

events/api.py
```diff
@@ -529,6 +529,23 @@ def create(self, validated_data):
                 raise error
         return instance
 
+    def create_keyword(self, validated_data):
+        if 'data_source' not in validated_data:
```
akx: Feels kind of icky that a `create`... function is mutating data that's passed in.

events/api.py
```diff
@@ -529,6 +529,23 @@ def create(self, validated_data):
...
+        validated_data['created_by'] = self.user
+        validated_data['last_modified_by'] = self.user
+        try:
+            instance = super().create(validated_data)
+        except IntegrityError as error:
+            if 'duplicate' and 'pkey' in str(error):
+                raise serializers.ValidationError({'id':_("An object with given id already exists.")})
+            else:
+                raise error
+        return instance
```
akx: . . . Wait, this function is almost exactly the same as `create()` above. What's up with that?

events/api.py
```diff
@@ -568,20 +585,59 @@ def _clean_qp(query_params):
...
     alt_labels = serializers.SlugRelatedField(slug_field='name', read_only=True, many=True)
+    data_source = serializers.PrimaryKeyRelatedField(queryset=DataSource.objects.all(),
+                                                     required=False, allow_null=True)
 
     class Meta:
         model = Keyword
         exclude = ('n_events_changed',)
+        
+    def update(self, instance, validated_data):
+        if 'data_source' in validated_data and not validated_data['data_source']:
```
akx:
```python
if not validated_data.get('data_source'):
  validated_data.pop('data_source', None)
```
would be a more succinct way to put this.

events/api.py
```diff
@@ -568,20 +585,59 @@ def _clean_qp(query_params):
...
+        if 'data_source' not in validated_data:
+            validated_data['data_source'] = self.data_source
+        return super().update(instance, validated_data)
+        
+    def create(self, validated_data):
+        # if id was not provided, we generate it upon creation:
+        if 'id' not in validated_data:
+            validated_data['id'] = generate_id(self.data_source)          
+        if 'data_source' not in validated_data or not validated_data['data_source']:
+          raise ValidationError({'data_source': _('Datasource is required field')})
```
akx: Shouldn't this be done as part of validation?

events/api.py
```diff
@@ -620,6 +676,9 @@ def get_queryset(self):
             queryset = queryset.filter(name__icontains=val)
         return queryset
 
+    @atomic
```
akx: This doesn't feel like the right place to stuff `@atomic`...

events/api.py
```diff
@@ -721,13 +784,44 @@ def filter_division(queryset, name, value):
...
     view_name = 'place-detail'
     divisions = DivisionSerializer(many=True, read_only=True)
-
+    
     class Meta:
         model = Place
         exclude = ('n_events_changed',)
+        
+    def update(self, instance, validated_data):
+        if 'data_source' in validated_data and not validated_data['data_source']:
```
akx: These should be done as part of validation, if you ask me...

events/api.py
```diff
@@ -721,13 +784,44 @@ def filter_division(queryset, name, value):
...
+        if 'publisher' in validated_data and not validated_data['publisher']:
+          del validated_data['publisher']
+        if 'data_source' not in validated_data:
+            validated_data['data_source'] = self.data_source
+        if 'publisher' not in validated_data:
+            validated_data['publisher'] = self.publisher
+        return super().update(instance, validated_data)
+        
+    def create(self, validated_data):
+        # if id was not provided, we generate it upon creation:
```
akx: These should be done as part of validation, if you ask me...

hkurhinen:
It is true that this PR was done in quite serious hurry. I'll close this one and refactor the code as soon as I have time and open several new ones, each for one specific issue.

PR #231, opened 2017-09-18
Fix/update django

Rikuoja:
* Old migrations squashed to get rid of unsupported old `hstore`.
* All needed dependencies updated to current versions.
* `ApiKeyUser` model added, because all `django-reversion` versions compatible with current Django require user model (as opposed to an `AnonymousUser`) actually being present when saving.


requirements.txt
```diff
@@ -4,68 +4,68 @@
...
-git+https://github.com/City-of-Helsinki/django-helusers@v0.1#egg=django-helusers==0.1
-git+https://github.com/City-of-Helsinki/munigeo#egg=django-munigeo
-astroid==1.4.8            # via pylint
-bleach==1.5.0
-click==6.6                # via pip-tools
-contextlib2==0.5.4        # via raven
-defusedxml==0.4.1         # via python3-openid
-django-allauth==0.28.0
+-e git+https://github.com/City-of-Helsinki/django-helusers@v0.1#egg=django-helusers
+-e git+https://github.com/City-of-Helsinki/munigeo#egg=django-munigeo
```
tituomin: Munigeo is available in pip now (though maybe not the HEAD of master). Could you at least pin a version here?

PR #236, opened 2017-09-25
Added possibility to create, update and remove keywords and places using API

hkurhinen:
Added support for creating, editing and removing keywords and places using API.

Removed deprecated GeoModelSerializer and added rest_framework_gis ( as advised at: https://github.com/djangonauts/django-rest-framework-gis/blob/master/README.rst ) to settings.INSTALLED_APPS since serializing geojson always resulted null.

Added tests for this functinality.


hkurhinen:
@Rikuoja is there something wrong with this PR or would it be possible to merge it into the core project.

events/api.py
```diff
@@ -586,8 +587,12 @@ def update(self, instance, validated_data):
                 raise PermissionDenied()
         else:
             # without api key, the user will have to be admin
-            if not instance.is_user_editable() or not instance.can_be_edited_by(self.user):
-                raise PermissionDenied()
+            if hasattr(instance, 'can_be_edited_by'):
```
Rikuoja: Actually, instead of disregarding `can_be_edited_by`, the method `can_be_edited_by` should be implemented in `models.py` for `Keyword` and `Place` instances.

Essentially, use the same method as in the `Event` model. They just use the `is_admin` method.

If you don't implement `can_be_edited_by`, the added `is_admin` methods in `Place` and `Keyword` are rather useless :)

events/models.py
```diff
@@ -346,6 +352,20 @@ def save(self, *args, **kwargs):
         else:
             self.divisions.clear()
 
+    def is_admin(self, user):
```
Rikuoja: Also implement `can_be_edited_by` method that uses this method.

events/models.py
```diff
@@ -235,6 +235,12 @@ class Keyword(BaseModel):
     def __str__(self):
         return self.name
 
+    def is_admin(self, user):
```
Rikuoja: Also implement `can_be_edited_by` method that uses this method.

anttileppa:
@Rikuoja merged the upstream and fixed all issues mentioned in comments, would it now be possible to merge this?

Rikuoja (changes requested):
Hm, I still couldn't find tests or permission checks for destroying objects?

Didn't test this, but currently the code looks like any admin user could delete places and keywords, even if the data source is *not* user_editable. This is not the desired outcome.

So, we need tests verifying that places and keywords cannot be soft_deleted unless the data_source is explicitly user_editable and the user has the rights.

anttileppa:
@Rikuoja true, completely forgot those. Should be fixed now

anttileppa:
@Rikuoja is there still something wrong or could this PR be merged?

Rikuoja (approved):
Looks like everything is in order :) Thanks for your contribution!

PR #238, opened 2017-10-12
Flake8

hukka:
Almost passes. Two bugs left, will make new PRs.


Rikuoja:
Rebase needed to get rid of Espoo.

Issue #235, opened 2017-09-22
Create hierarchical admin rights model on top of #234


Rikuoja:
When the hierarchical Organization structure is in place, an extension on User admin rights is required. This means that Organization *admin* Users will have admin rights on all events POSTed by an Organization or its suborganizations.

Also, the User model should be extended to include *non-admin* Users that have the right to POST draft Events  (`PublicationStatus.DRAFT`) for an Organization, but no rights to publish Events. Any *admin* User of the Organization (or its super-Organizations) may then see the draft Events and approve their publication.

Rikuoja:
![image](https://user-images.githubusercontent.com/9367712/31170007-34ac007a-a903-11e7-9f64-0facbd05a419.png)

Rikuoja:
User may have rights to

T1) POST draft events for Organization, or
T2) POST live events for Organization and approve draft events (PUT) for Organization, or
T3) POST live events for Organization and its suborganizations, and approve draft events (PUT) for Organization and its suborganizations.

Rikuoja:
Essentially, I believe only two kinds of organization memberships are required:

1) If a User is *non-admin*, he may only POST and PUT drafts for Organization and any suborganizations. This would be user T1.
2) If a User is *admin*, he may POST and PUT events (including drafts made by non-admins) for Organization and any suborganizations. This would be users T2 and T3, where T2 is member of a suborganization and T3 a member of the superorganization.

Rikuoja:
![image](https://user-images.githubusercontent.com/9367712/31170361-71594c8e-a904-11e7-8d4f-2a025b3059e0.png)

Rikuoja:
Also, probably we want to include this user rights model to the django-organizations or django-helusers app, not directly into linkedevents.

A similar user rights management based on the organization structure and admin/non-admin dichotomy might be needed in other of our projects as well.

Rikuoja:
@juyrjola What is your opinion on the model? If we create a two-tiered user rights model, should it

1) be included in django-helusers, or
2) should it be included in a Helsinki organizations importer and app, because the organization structure tells us which suborganizations the user can admin?

Rikuoja:
Use organization data model from https://github.com/City-of-Helsinki/paatos/blob/master/decisions/models/organization.py , possibly with hot-swappable Datasource etc. project specific models?

User rights (admin/non-admin) are linkedevents-specific, so let's add them there for the time being.

Rikuoja:
[...]

Issue #234, opened 2017-09-22
Create standalone django-organizations app that can be reused across projects


Rikuoja:
Essentially, separate the Organization model into a separate pluggable django app.

The app should contain
* Organization model
* Organization viewset and serializer
* Organization importer that automatically imports and updates the Helsinki organization structure from current data at https://api.hel.fi/paatos/v1/organization/

Rikuoja:
In addition to hierarchical Suborganizations, an Organization must have "affiliated" organizations which are essentially the same as suborganizations, but separated from the Helsinki organization structure.

Add simple django-admin view to the current admin for creating affiliated organizations to an organization.

Rikuoja:
This one is in use as of https://github.com/City-of-Helsinki/linkedevents/commit/5909ea0b3488fb46666aace900da6baf226ddf26

PR #240, opened 2017-11-27
Orghierarchy

mingfeng:
This PR integrates the `django_orghierarchy` app to `linkedevents` to replace the existing Organization model.

#### Backgrounds
The original idea of the `django-orghierarchy` app is to extract the Organization related models into a reusable app so that it can be used in different projects (see details in #234). The app defines its own `Organization`, `OrganizationClass` and `DataSource` model. The `DataSource` is a swappable model, and `events.DataSource` is used to replace the default one in `django-orghierarchy` when integrating to linkedevents. The app also provides an importer that allows importing organization data from different sources.

Later on we will extend existing permission check based on this new app (see details in #235).

Refs: #234, #235


events/models.py
```diff
@@ -200,7 +200,7 @@ class Organization(MPTTModel, BaseModel):
     parent = TreeForeignKey('self', null=True, blank=True, on_delete=models.SET_NULL,
                             related_name='children')
     admin_users = models.ManyToManyField(
-        User, blank=True, related_name='admin_organizations'
+        User, blank=True, related_name='admin_old_organizations'
```
mingfeng: This is because the `Organization` model in `django_orghierarchy` has the same field with the same related_name.

mingfeng:
Update:

Specified the fields for `OrganizationSerializer`. Hide permission related fields.

PR #242, opened 2017-11-29
 Allow fetching objects published by replaced/replacing organization

mingfeng:
When filtering the objects with publishers, the REST API will also return objects that are published by a replaced organization when a replacing organization is provided. Also it will return objects for replacing organization if a replaced organization is provided.

### DIFF
--- a/events/api.py
+++ b/events/api.py
@@ -134,10 +134,6 @@ def get_authenticated_data_source_and_publisher(request):
         publisher = data_source.owner
         if not publisher:
             raise PermissionDenied(_("Data source doesn't belong to any organization"))
-
-        if publisher.replaced_by:
-            # using new organization that replaces current organization
-            publisher = publisher.replaced_by
     else:
         # objects created by api are marked coming from the system data source unless api_key is provided
         # we must optionally create the system data source here, as the settings may have changed at any time
--- a/events/api.py
+++ b/events/api.py
@@ -147,6 +147,38 @@ def get_authenticated_data_source_and_publisher(request):
     return data_source, publisher
 
 
+def get_publisher_query(publisher):
+    """Get query for publisher (Organization)
+
+    Some organizations can be replaced by a new organization.
+    We need to return objects that reference on replaced
+    organization when querying the new organization, and vice
+    versa.
+
+    :param publisher: a or a list of filtering organizations
+    :type publisher: str, Organization, list
+    :return: the query that check both replaced and new organization
+    """
+    if isinstance(publisher, list):
+        q = Q(
+            publisher__in=publisher,
+        ) | Q(
+            publisher__replaced_by__in=publisher,
+        ) | Q(
+            publisher__replaced_organization__in=publisher,
+        )
+    else:
+        q = Q(
+            publisher=publisher,
+        ) | Q(
+            publisher__replaced_by=publisher,
+        ) | Q(
+            publisher__replaced_organization=publisher,
+        )
+
+    return q
+
+
 class JSONLDRelatedField(relations.HyperlinkedRelatedField):
     """
     Support of showing and saving of expanded JSON nesting or just a resource
@@ -946,7 +978,9 @@ class ImageViewSet(viewsets.ModelViewSet):
         publisher = self.request.query_params.get('publisher', None)
         if publisher:
             publisher = publisher.lower().split(',')
-            queryset = queryset.filter(publisher__in=publisher)
+            q = get_publisher_query(publisher)
+            queryset = queryset.filter(q)
+
         data_source = self.request.query_params.get('data_source')
         # Filter by data source, multiple sources separated by comma
         if data_source:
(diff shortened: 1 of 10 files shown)


events/api.py
```diff
@@ -147,6 +147,38 @@ def get_authenticated_data_source_and_publisher(request):
...
+
+    :param publisher: a or a list of filtering organizations
+    :type publisher: str, Organization, list
+    :return: the query that check both replaced and new organization
+    """
+    if isinstance(publisher, list):
+        q = Q(
+            publisher__in=publisher,
+        ) | Q(
+            publisher__replaced_by__in=publisher,
```
Rikuoja: What will happen if the replaced organization is replaced?

This probably does not happen at the moment (and I don't know if it ever will), but what if we have a chain of organizations that replace one another? I believe we should display the objects of all the previous organizations there.

mingfeng: Good point, I think it should be possible to have multiple mptt tree (and multiple tree manager) in a model, I think it will require changing some existing code how we use the mptt methods (e.g. providing `tree_id` for the methods). But not sure if it's necessary to introduce it now, maybe better we introduce it in the time when we need it as there's possibility that we ain't gonna need it.

mingfeng: Another option would to recursively query the parents for each provided publisher. This will not affect the code in other places, but will sacrifice the performance quite a bit.

events/api.py
```diff
@@ -1521,7 +1556,7 @@ def filter_queryset(self, queryset):
             if 'show_all' in self.request.query_params:
                 # Show all events for this organization,
                 # along with public events for others.
-                auth_filters |= Q(publisher=self.organization)
+                auth_filters |= get_publisher_query(self.organization)
```
Rikuoja: Here, `self.organization` is actually defined by `get_authenticated_data_source_and_publisher`, and there it is generated from the `User` model `get_default_organization`.

I believe that when we are doing this change, we already might want to change that method to not return replaced organizations? Even if this filter works because it checks the replacing organization, fixing the organization to be non-replaced will fix the `v1/user` endpoint as well, I believe.

events/api.py
Rikuoja: Hm. Yeah, thinking about it again, perhaps it is good that we do not allow downright ancient organizations to be used in queries. If there is a new replacement, I think the users should be aware of that or at least the previous replacement by that point :)

events/api.py
mingfeng: Make sense

mingfeng:
Updates:

Return new organization if the data source owner organization is replaced for `get_authenticated_data_source_and_publisher` and `get_default_organization`.

helevents/models.py
```diff
@@ -9,4 +9,8 @@ def get_display_name(self):
         return '{0} {1}'.format(self.first_name, self.last_name).strip()
 
     def get_default_organization(self):
-        return self.admin_organizations.order_by('created_time').first()
+        org = self.admin_organizations.order_by('created_time').first()
```
Rikuoja: I think here we should just filter out the replaced organizations. That way we don't accidentally return a replaced organization. This will also make sure that the user has at least one current (non-replaced) organization, so it's not an old user.

mingfeng: Just want to make sure, that means those users that assigned to `replaced_organization.admin_users` will return None for this method, which will break backward compatibility unless the admin manually copy the users from old organization to new one?

Rikuoja: @mingfeng I believe that is the only transition that we should make automatically. While the old objects may remain published by the old organizations, I think we should create an admin command for copying admin_users to the new organizations.

Because we do not want to keep the users scattered around different organizations, I think it is safer to have all the people who may edit data listed under the new organization directly, so we can check permissions without having to check replaced organizations separately.

mingfeng: Will do the auto data transition in `django-orghierarchy`, we can leave this PR open until the changes is done there.

Rikuoja: Perhaps we should do the same for `DataSource`. `DataSource` and `User` are the only two external models referred to in `django-orghierarchy`, and it is better that they are not linked to replaced organizations. That way, also the api keys for data sources automatically map to the new organizations.

Rikuoja: However, there is the problem that datasource is linkedevents-specific. I'm referring to the datasource `owner` field here. To map the apikeys correctly, that would have to be done in linkedevents :(

mingfeng: That's true, it has to be done in linkedevents then.

mingfeng:
Updates:

Add signals to update Organization `admin_users` and `regular_users`, as well as related data source `owner` if an old organization is replaced by a new one.

events/api.py
```diff
@@ -134,6 +134,10 @@ def get_authenticated_data_source_and_publisher(request):
         publisher = data_source.owner
         if not publisher:
             raise PermissionDenied(_("Data source doesn't belong to any organization"))
+
+        if publisher.replaced_by:
```
Rikuoja: This one should not be needed, if we map all api keys to non-replaced organizations?

PR #245, opened 2017-12-05
Extended version of access control to events

mingfeng:
This is an extended version of access control to events based on the specs described in #235.

The user rights can be briefly described as:

Admin users:
- admin users of an organization can create event (of the default organization) with publication status PUBLIC and DRAFT
- admin users of an organization can modify and delete events of its own organizations
- admin users of an organization can modify and delete events of their sub-organizations
- admin users of an organization can modify and delete events of affiliated organization (e.g. if organization A has been assigned a responsible organization B, then the admin users of B can modify and delete events of A)

Note that the permissions between sub-organizations and affiliated organizaitons are not recursively inherited.

Regular users:
- regular users of an organization (of users' default organization ) can create event with publication status DRAFT
- regular users of an organization can modify/delete events with publication status DRAFT

The user rights not specified above works the same way as before.

Refs: #235


events/api.py
```diff
@@ -1555,37 +1562,78 @@ def filter_queryset(self, queryset):
...
+                    # along with public events for others.
+                    auth_filters |= get_publisher_query(self.organization)
+            queryset = queryset.filter(auth_filters)
+        else:
+            # prevent changing events user does not have write permissions (for bulk operations)
+            queryset = self.request.user.get_editable_events(self.organization, queryset)
+
         queryset = _filter_event_queryset(queryset, self.request.query_params,
                                           srs=self.srs)
         return queryset.filter()
```
Rikuoja: So what is the exception that the API will return for POST and PUT if the event is not in the queryset after filtering, i.e. the user has no edit rights? I.e. does the API return HTTP 401, 403 or 404?

events/tests/test_api.py
```diff
@@ -229,3 +232,217 @@ def test_event_list_with_publisher_filters(self):
             self.assertEqual(response.status_code, status.HTTP_200_OK)
             # note that org-2 is replaced by org-1
             self.assertEqual(len(response.data['data']), 2)  # event-1 and event-2
+
+    def test_random_user_create_event_denied(self):
```
Rikuoja: I guess for completeness' sake, we should add the same tests for a random user trying to PUT or DELETE any events?

events/tests/test_api.py
```diff
@@ -229,3 +232,217 @@ def test_event_list_with_publisher_filters(self):
...
+        data = minimal_event_dict(self.data_source, self.org_1, location_id)
+        data['publication_status'] = 'public'
+        self.client.force_authenticate(self.user)
+        response = self.client.put(url, data, format='json')
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+        event.refresh_from_db()
+        self.assertEqual(event.publication_status, PublicationStatus.PUBLIC)
+
+    def test_regular_user_create_public_event_denied(self):
```
Rikuoja: I guess for completeness' sake, we should add the same tests for a regular user trying to PUT or DELETE any public events?

events/api.py
Rikuoja: Ah, apparently the tests specify it is 403, that's good. I just don't get where the logic for showing 403 is hidden after filtering has been done :D

events/tests/test_api.py
mingfeng: Make sense

helevents/tests/test_models.py
```diff
@@ -23,3 +47,41 @@ def test_get_default_organization_org_replaced(self):
...
+        is_admin = self.user.is_admin(self.child_org)
+        self.assertTrue(is_admin)
+
+        # test for some other organization
+        is_admin = self.user.is_admin(self.org_1)
+        self.assertFalse(is_admin)
+
+    def test_is_regular_user(self):
+        self.org.regular_users.add(self.user)
+
```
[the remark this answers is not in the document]
mingfeng: That's a very good point, I'll add those asserts.

events/api.py
mingfeng: That's a very good question. In general, if a object is neither readable nor editable by a user, then it's better to return a 404 for all requests, but if the object is readable but not editable by user, then it should return 403 when user tries to edit the object.

Normally we want the `get_object` method to call the `filter_queryset` method to filter out unwanted objects first and then tries to get the wanted object from the filtered queryset, so it's not a good  idea to return different querysets for read requests (GET) and write requests (PUT, POST) for `filter_queryset` method as it means we can get 200 for GET but 404 for PUT and POST on the same object URL.

In our code, we are using a custom `get_object` method which does not use `filter_queryset` method at all, which means the `filter_queryset` is only used multi-objects operations (list and bulk updates). Thus we do not have the problem above.

Another options if to put this logic to the `perform_update` method, but we will write more lines of code. Cannot say which is better, just choose one that I thought simpler :smile:

mingfeng:
@Rikuoja That's true, I'll add tests for those bulk methods.

mingfeng:
Updates:

- Replace `is_admin` method for Image and Event model with `can_be_edited_by`, which is used to check if a user can edit target object.  This is because none admin (or regular) users can also edit event now.
- Add tests for bulk REST API interfaces
- Remove obvious duplicated auth-related tests. `test_api` contains more thorough tests that covers different permission check scenarios

events/tests/test_event_auth_data.py
```diff
@@ -1,48 +0,0 @@
-# -*- coding: utf-8 -*-
-import pytest
-from .utils import versioned_reverse as reverse
-
-from .test_event_get import get_detail
-
-
-@pytest.mark.django_db
-def test__user_sees_auth_fields(api_client, event, user, user2, organization, organization2):
```
Rikuoja: I guess these two asserts (`assert response.data.get('publication_status') == "public"` and `assert response.data.get('publication_status') is None`) should be included somewhere in the new tests, because they test field visibility for admins and random users?

Rikuoja: Oh, there they are, the diff was so large it wasn't loaded! My bad :) So how about removing `test__a_non_admin_cannot_update_an_event` in `test_event_put.py` still?

mingfeng: True that, I'll remove it.

Issue #248, opened 2017-12-11
Add support for #234 and #235 in Admin UI


aleksisalonen:
Add support for the organization model (#234) and admin rights (#235) in the admin UI:

https://api.hel.fi/linkedevents/admin/events/organization/

The admin UI should have support for:
1) Viewing all levels of the new hierarchy
2) Adding new affiliated organizations and editing them
3) User rights management for both admin-level and basic user-level

Mockup for the organizations-view:
![admin-ui-new-orgs-11122017](https://user-images.githubusercontent.com/16101544/33833899-da9a36a8-de89-11e7-8a3e-9ffd9c524235.jpg)

Mockup for user rights management:
![admin-ui-new-orgs-add-users-view-11122017](https://user-images.githubusercontent.com/16101544/33833961-0ac8ed2e-de8a-11e7-825c-e8f4fcba16d6.jpg)

Mockup for adding affiliated organization:
![admin-ui-new-orgs-add-aff-organization-view-11122017](https://user-images.githubusercontent.com/16101544/33833978-1938b006-de8a-11e7-81b5-4b562528577c.jpg)

mingfeng:
Regarding the drop down options (Toiminto),  the options are supposed to be bulk operations and user can check multiple objects for the operation. I think it's better not to add single object actions (add affiliated organization to an organization or edit affiliated organization) because in such case we'll need to hack the Django admin UI, for example, disable single object action options when multiple objects are selected.

I think we can move those operations into single organization editing page, and allow user to add/remove/edit affiliated organizations there.

mingfeng:
Regarding display sub-organization and affiliated organization in a tree structure in list admin. Since the affiliated organizations are not part of the tree structure from our Organization model, I think it will be quite tricky to force this information into a tree structure.

If we really need to display affiliated organizations as part of the tree, I'd suggest we refactor the model to make it part of the hierarchical tree. The changes to the model will be like:
* remove `responsible_organization` field
* add `internal_type` field, default value is `normal`

When we need to add an affiliated organization to an organization, we create an new organization and set its `parent` field to the target organization (the same as `responsible_organization` in current model) and set the `internal_type` to `affiliated`. 

This way the affiliated organizations are also part of the tree and will be quite trivial to display it in tree structure, and with the `internal_type` information, we can display it in different color based on this field.

This will require some changes on permission checking code, but we have quite good  code isolation in the new hierarchical permissions, so I think the code change on that will be quite small.

[...]

PR #249, opened 2017-12-20
Corresponding changes in Organization REST API and permission check for Organization model

mingfeng:
The `responsible_organization` field is removed from Organization model, and the affiliated organizations are part of the organization hierarchical tree. We need to change the REST API and permission check accordingly.

Refs: https://github.com/City-of-Helsinki/linkedevents/issues/248

### DIFF
--- a/events/api.py
+++ b/events/api.py
@@ -904,10 +904,15 @@ class OrganizationSerializer(LinkedEventsSerializer):
     )
     sub_organizations = serializers.HyperlinkedRelatedField(
         queryset=Organization.objects.all(),
-        source='children',
         view_name='organization-detail',
         many=True,
     )
+    affiliated_organizations = serializers.HyperlinkedRelatedField(
+        queryset=Organization.objects.all(),
+        view_name='organization-detail',
+        many=True,
+    )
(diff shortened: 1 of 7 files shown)


events/api.py
```diff
@@ -904,22 +904,30 @@ class OrganizationSerializer(LinkedEventsSerializer):
...
+    is_affiliated = serializers.SerializerMethodField()
 
     class Meta:
         model = Organization
         fields = (
             'id', 'data_source', 'origin_id',
             'classification', 'name', 'founding_date',
             'dissolution_date', 'parent_organization',
-            'sub_organizations', 'responsible_organization',
+            'sub_organizations', 'affiliated_organizations',
```
Rikuoja: I believe we should test these two fields, `sub_organizations` and `affiliated_organizations`, in the `TestOrganizationSerializer`. Currently, both fields have the same queryset so I wasn't sure that `sub_organizations` only lists proper suborganizations and `affiliated_organizations` only lists affiliated ones.

mingfeng:
Update:

Add tests for sub_organization and affiliated_organization fields for organization REST API

PR #252, opened 2018-01-08
Make importer logging more neater

vikoivun:


Rikuoja:
Flake8 needed!

vikoivun:
This actually make no sense whatever

PR #258, opened 2018-05-08
Filter events by address locality

hkurhinen:
Added addess_locality_fi, addess_locality_sv and addess_locality_en query parameters which allow filtering events based on their address localities in different languages.


anttileppa:
@Rikuoja filter_division relies on having source system for divisions (e.g. https://kartta.hel.fi) and most Finnish cities do not have such system or the system is not integrated to LinkedEvents API so could you reconsider merging this feature?

PR #263, opened 2018-05-18
Cleanup importer logging

vikoivun:
All importers are changed to use logging.* functions for
logging instead of prints. Also removed the superfluous
logger created in base.py. Django has already created a
logger for us, therefore we can just use `logging`.

Default log format has been changed to include severity
and module name.

the little-used verbosity flag usage has been removed
from importers. Instead messages are tagged with log
level. If verbosity flag is desired, it can be used to
control logging level.

### DIFF
--- a/events/importer/harrastushaku.py
+++ b/events/importer/harrastushaku.py
@@ -65,7 +65,7 @@ class HarrastushakuImporter(Importer):
         self.tprek_data_source = DataSource.objects.get(id='tprek')
         self.ahjo_data_source, _ = DataSource.objects.get_or_create(id='ahjo', defaults={'name': 'Ahjo'})
         self.organization, _ = Organization.objects.get_or_create(origin_id='u48040030',
-                                                               data_source=self.ahjo_data_source)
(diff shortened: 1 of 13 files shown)


vikoivun:
This has now been rebased on master, to make sure no funny stuff happens during merge.

events/importer/harrastushaku.py
```diff
@@ -64,7 +65,10 @@ def setup(self):
         self.data_source, _ = DataSource.objects.get_or_create(id=self.name, defaults={'name': 'Harrastushaku'})
         self.tprek_data_source = DataSource.objects.get(id='tprek')
         self.ahjo_data_source, _ = DataSource.objects.get_or_create(id='ahjo', name='Ahjo')
-        self.organization = Organization.objects.get(origin_id='u48040030', data_source=self.ahjo_data_source)
+        try:
```
Rikuoja: I would just use `get_or_create` instead, the same way that the other importers do this.

This way, each importer can be run even without running the whole `import_organizations` thing, if the user doesn't want hierarchical organization structure quite yet.

PR #265, opened 2018-05-29
Dockerize project's development setup

igordavydsson:
Refs #266

### DIFF
--- a/docker/django/Dockerfile
+++ b/docker/django/Dockerfile
@@ -1,0 +1,18 @@
+FROM python:3.6
+
+ENV PYTHONUNBUFFERED 0
+
+RUN apt-get update \
+    && apt-get install --no-install-recommends -y gdal-bin python-gdal python3-gdal
+
+COPY requirements.txt /usr/src/
+
+RUN pip install --upgrade pip \
+    && pip install pip-tools \
+    && pip install --src /usr/src -r /usr/src/requirements.txt
+
+RUN mkdir /code
+
+WORKDIR /code
+
+COPY . /code
--- a/docker/postgres/Dockerfile
+++ b/docker/postgres/Dockerfile
@@ -1,0 +1,11 @@
+FROM postgres:9.5
+
+RUN apt-get update \
+    && apt-get install --no-install-recommends -y \
+        postgis postgresql-9.5-postgis-2.4 postgresql-9.5-postgis-2.4-scripts
+
+RUN localedef -i fi_FI -c -f UTF-8 -A /usr/share/locale/locale.alias fi_FI.UTF-8
+
+ENV LANG fi_FI.UTF-8
+
+COPY ./docker-entrypoint.sh /docker-entrypoint-initdb.d/docker-entrypoint.sh
--- a/docker/postgres/docker-entrypoint.sh
+++ b/docker/postgres/docker-entrypoint.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+set -e
+
+psql -v ON_ERROR_STOP=1 --username "$POSTGRES_USER" --dbname "linkedevents" <<-EOSQL
+	CREATE EXTENSION IF NOT EXISTS postgis;
+	CREATE EXTENSION IF NOT EXISTS hstore;
+EOSQL
--- a/local_settings.py.template
+++ b/local_settings.py.template
@@ -1,0 +1,85 @@
+DEBUG = True
+
+DATABASES = {
+    'default': {
+        'ENGINE': 'django.contrib.gis.db.backends.postgis',
+        'NAME': 'linkedevents',
+        'USER': 'linkedevents',
+        'PASSWORD': 'linkedevents',
+        'HOST': 'linkedevents-db'
+    }
+}
+
+CUSTOM_MAPPINGS = {
+    'autosuggest': {
+        'search_analyzer': 'standard',
+        'index_analyzer': 'edgengram_analyzer',
+        'analyzer': None
+    },
+    'text': {
+        'analyzer': 'default'
+    }
+}
+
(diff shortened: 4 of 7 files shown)


Rikuoja:
Asking @jussiarpalahti for comments, as he's been doing containerization of our projects lately :)

Rikuoja:
And @vikoivun will comment as our infrastructure and documentation person :P

local_settings.py.template
```diff
@@ -0,0 +1,70 @@
...
+    'default': {
+        'ENGINE': 'django.contrib.gis.db.backends.postgis',
+        'NAME': 'linkedevents',
+        'USER': 'linkedevents',
+        'PASSWORD': 'linkedevents',
+        'HOST': 'linkedevents-db'
+    }
+}
+
+LANGUAGES = (
```
Rikuoja: I wouldn't have a template with different languages than the repo default, I guess this means there will be extra languages in linkedevents migrations that are not used by the code? Don't know if extra unused fields in the db will cause any trouble, though.

igordavydsson: I could copypaste settings from other languages, maybe it would be ok for now. :D

jussiarpalahti:
@igordavydsson Have you done Travis integration with docker compose before?
We have github -> Travis -> Docker Hub pipeline in the works. I'd would be nice to incorporate these changes and not do two separate build configs (the other one for dev).

docker/postgres/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM postgres:10.1
```
jussiarpalahti: We are currently stuck on Ubuntu 16.04 Postgres 9.5 so containers should reflect that.

docker/django/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+RUN apt-get update \
+    && apt-get install -y gdal-bin python-gdal python3-gdal
+
+# Let's install git packages from requirements.txt here
+RUN mkdir -p /src
+
+COPY requirements.txt /src/
+
+RUN pip install --upgrade pip \
+    && pip install --upgrade typing \
```
vikoivun: Why add typing? Especially through dockerfile?
Python 3.5 (and thus 3.6) will bring it along also…

docker/django/Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM python:3.4
+
+ENV PYTHONUNBUFFERED 0
+
+RUN apt-get update \
+    && apt-get install -y gdal-bin python-gdal python3-gdal
+
+# Let's install git packages from requirements.txt here
+RUN mkdir -p /src
```
vikoivun: Instead of this, you can shorten the dockerfile a bit and use /usr/src

docker/django/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+
+COPY requirements.txt /src/
+
+RUN pip install --upgrade pip \
+    && pip install --upgrade typing \
+    && pip install pip-tools \
+    && pip install -r /src/requirements.txt
```
vikoivun: Use `pip install --src /usr/src -r requirements.txt` to install in /usr/src. Nifty! Saves three lines!

docker/postgres/docker-entrypoint.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+set -e
+
+psql -v ON_ERROR_STOP=1 --username "$POSTGRES_USER" --dbname "linkedevents" <<-EOSQL
+	CREATE EXTENSION postgis;
```
vikoivun: CREATE EXTENSION postgis IF NOT EXISTS (and the same for hstore)
is a bit more robust.

docker/postgres/Dockerfile
```diff
@@ -0,0 +1,10 @@
...
+
+RUN apt-get update \
+    && apt-get install -y postgis
+
+RUN localedef -i fi_FI -c -f UTF-8 -A /usr/share/locale/locale.alias fi_FI.UTF-8
```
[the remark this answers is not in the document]
vikoivun: Ah. I see, whether it is or not is irrelevant. The environment set for the container is used to create the initial database. I guess that is sensible.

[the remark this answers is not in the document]
igordavydsson: yeah, making sure that the database uses the right locale

docker/postgres/Dockerfile
igordavydsson: alright! switchin' to postgres 9.5 image, which is based on `debian-slim`, I guess that should be okay

docker/django/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM python:3.4
```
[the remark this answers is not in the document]
igordavydsson: ah, sure, will do

docker/django/Dockerfile
igordavydsson: oh, yeah, good idea! will use that 🙂

docker/django/Dockerfile
igordavydsson: nifty indeed!

docker/postgres/docker-entrypoint.sh
igordavydsson: 👍

igordavydsson:
@jussiarpalahti @vikoivun thanks for the comments! tried to address all of them.

@jussiarpalahti I haven't done Travis integration before, but I've seen a few tutorials on how to do that. Could be done for this project as well, but would take some time and I don't have it right now during this sprint. Maybe later there would be a chance to implement it. Anyway, I think such integration could be later based on this docker setup with some tweaks and adjustments, of course.

Issue #266, opened 2018-05-29
Extend the project to LinkedCourses


igordavydsson:
**TODO**:
- [x] Setup the project (by adding Docker)
- [x] Add new fields to Event model

PR #268, opened 2018-05-29
Fix Django language code for Chinese

igordavydsson:
Django does not know `zh` code.

Refs #266


Rikuoja:
Interestingly, I couldn't figure out whether to use ´zh-cn´ or ´zh-hans´, as the former is deprecated (https://github.com/django/django/commit/e5e044da87800feb6ef63fef1765d8c05022d926). Therefore, I just used ´cn´, but apparently I shouldn't have :D

igordavydsson:
@Rikuoja: Fixed the bug in `TranslatedModelSerializer`. Seems to work fine now. :)

Rikuoja:
Ah, one more thing: the production database http://api.hel.fi/linkedevents/v1/language/ currently already has `zh` as a language, albeit with no translations. I'm guessing we should also do a data migration for that, so we don't end up with separate `zh` and `zh-hans` language objects, only the latter having translation available.

igordavydsson:
@Rikuoja Wouldn't this `migrations.RenameField` migrate all the existing data (if it finds any) to the new column?

igordavydsson:
@Rikuoja added data migration. not sure if it should be `zh_hans` or `zh-hans` though... 🤔 💭

Rikuoja:
Ah! It actually wouldn't work in the first place, I didn't try it yet, because langdetect uses `zh-cn` :D So we'll have to map `zh-hans` to `zh-cn` somewhere. This is getting stupid.

events/migrations/0055_fix_chinese_language_code.py
```diff
@@ -0,0 +1,148 @@
...
+from django.db import migrations, models
+
+
+def fix_chinese_language_object(apps, schema_editor):
+    Language = apps.get_model('events', 'Language')
+
+    try:
+        chinese_lang = Language.objects.get(id="zh")
+        chinese_lang.id = "zh_hans"
+        chinese_lang.save()
```
Rikuoja: Also, what happens here is that at least for me, the django ORM creates a copy of Chinese with the new primary key. The original object remains, and all foreign key references still point to the old `zh` primary key and not the new one :(

Rikuoja: Also, for some reason, both `zh` and `zh_hans` have `translation_available` as false, even though `ar` and `ru` list it as true :(

Rikuoja: Probably something with how modeltranslate serializes the `translation_available` field.

igordavydsson: I fixed 'def get_translation_available()' last time I ammended my commit. So, translation available shows `True` for me.
I will fix the foreign key links and extra copies and all that ORM jazz.

igordavydsson:
@Rikuoja Me thinks me has fixed this.

igordavydsson:
Due to [this bug](https://code.djangoproject.com/ticket/25012), I also need to change ForeignKey fields on other models, if I change `max_length` on the original model's pk field...

igordavydsson:
@Rikuoja alright! now everything should be in order.

PR #269, opened 2018-06-04
Add new fields to event model

igordavydsson:
Refs #266


igordavydsson:
@Rikuoja got a suggestion from @tuomas777 to put stuff into [this `custom_data` field](https://github.com/City-of-Helsinki/linkedevents/blob/master/events/models.py#L55). I think that `provider_contact_info` would fit there nicely. what do you think?

`typical_age_range`, however, should stay sas a new field IMO.

Rikuoja:
Hm. Yeah, any data that will not be needed in core Linkedevents or won't be used for filtering could, of course be stored in `custom_data`.

So, the question is, what are the other extra fields going to be like, do we need filtering based on them,  or do we just store them as strings. In that case, hstore could fit the bill.

On the other hand, we have `location` and `location_extra_info` in each event, so `provider` and `provider_contact_info` would make sense in the event itself too.

igordavydsson:
agreed with @tuomas777 that the implementation will take a slightly different approach.

**WIP**

events/models.py
```diff
@@ -440,6 +442,11 @@ class SuperEventType:
     has_start_time = models.BooleanField(default=True)
     has_end_time = models.BooleanField(default=True)
 
+    # In accordance with this convention schema.org/typicalAgeRange
```
akx: It could have a validator to enforce the convention.
Is indexing the field necessary?

events/tests/conftest.py
```diff
@@ -394,6 +394,8 @@ def complex_event_dict(data_source, organization, location_id, languages):
         'headline': {'en': TEXT, 'sv': TEXT, 'fi': TEXT},
         'short_description': {'en': TEXT, 'sv': TEXT, 'fi': TEXT},
         'provider': {'en': TEXT, 'sv': TEXT, 'fi': TEXT},
+        'provider_contact_info': TEXT,
+        'typical_age_range': TEXT
```
akx: Trailing comma missing.

igordavydsson:
@akx Thanks for the comments! PR is now updated with a slightly different implementation.

Rikuoja:
Hm, really depends on what you guys will need in the courses API. In my opinion, one field is indeed better than two, if you can live with the age range format!

tuomas777:
IMO that one combined field does not really make sense, it just makes the API harder to use, but if we want to use schema.org fields then I guess there is no way around it.

Rikuoja:
Yeah. If you feel you need two fields for the UI to use the API sensibly, then that is the way to go.

igordavydsson:
You mean I need to change it back to just two fields?

image

Rikuoja:
IMO the range looks better in the single field. So it's really up to whether your UI is able to handle the format or not :)

Rikuoja:
Mind that separate filters for min and max ages will be needed anyway in the `event` endpoint, regardless of how the API displays the range, so don't know what would be the problem with processing the range string?

igordavydsson:
@Rikuoja @tuomas777 well, here is the update for this PR.

spent a few days trying to implement a `through` model on the relation between `Event` and `Image`, but the current serializing logic is just too hard to overcome. had to pull back on that idea and just implement a direct m2m.

helevents/templates/rest_framework/event_list.html
```diff
@@ -100,6 +100,13 @@ <h3 id="event-text">Event publisher</h3>
 <pre><code>event/?publisher=ytj:0586977-6
 </code></pre>
 <p><a href="?publisher=ytj:0586977-6" title="json">See the result</a></p>
+<h3 id="event-text">Audience min and max age</h3>
```
Rikuoja: I would not include this in the main linkedevents documentation at the moment. That is because none of our actual event data have these ages now, so these filters are useless for linkedevents users. It is okay to include the fields in the code, though, because in the future they may come in handy.

This is one point in which you might need separate linked courses documentation that is geared towards courses users specifically, as only courses currently have this data available?

igordavydsson:
@Rikuoja alright! got rid of the documentation 🙂

igordavydsson:
@Rikuoja so, do I merge this now? there will maybe be some conflicts for the PR you just opened, when I merge this...

igordavydsson:
on a second thought, it should be fine. 🙂

PR #291, opened 2018-09-03
Fix image assignment during different imports

igordavydsson:
BaseImporter is used by multiple child importers. Only Events have
multiple images though, so we need to check what are we importing
at the moment and assign either a single ForeignKey image or
ManyToMany images.


igordavydsson:
Resolved by #295

Rikuoja:
Am I correct in assuming that linkedcourses has this fixed with the Harrastushaku base importer refactoring, i.e. this PR may be closed?

PR #295, opened 2018-10-04
Harrastushaku importer

### LINKED ISSUE #276: Implement harrastushaku importer

tuomas777:
Added [Harrastushaku](http://nk.hel.fi/harrastushaku/) importer which imports courses and places. 

The PR includes also some changes outside the importer that were required to get the whole thing working. Details are in the commit messages.

Closes #276

### DIFF
--- a/events/importer/base.py
+++ b/events/importer/base.py
@@ -9,11 +9,13 @@ from django.conf import settings
 from rest_framework.exceptions import ValidationError
 from django.contrib.gis.geos import Point, Polygon
 from django.contrib.gis.gdal import SpatialReference, CoordTransform
+
+from events.importer.sync import ModelSyncher
 from .util import separate_scripts
 
 from modeltranslation.translator import translator
 
-from events.models import Image, Language, Event, License, Offer, EventLink, Place
+from events.models import Image, Language, Event, Offer, EventLink, Place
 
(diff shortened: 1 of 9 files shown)


tuomas777:
Merged master to linked-courses and rebased this on top of that.

events/importer/base.py
```diff
@@ -74,23 +79,83 @@ def _set_multiscript_field(string, event, languages, field):
...
+        image._changed = True
+        image._created = True
+
+        return image
+
+    def _update_image(self, image, image_data):
+        if not hasattr(image, '_changed'):
+            image._changed = False
+
+        self._set_field(image, 'publisher', self.organization)
```
Rikuoja: This line here reveals a bug in the underlying `_set_field` method (or actually the `Image` model `is_user_edited` method at https://github.com/City-of-Helsinki/linkedevents/blob/master/events/models.py#L144) that is only revealed when trying to import all tprek locations with the `tprek` importer. It fails with 

```
    return bool(self.data_source.user_editable and self.last_modified_by)
AttributeError: 'NoneType' object has no attribute 'user_editable'
```

This is because old existing images that we are updating do not necessarily have data source set, as they come from old times when `data_source` was missing from the image model.

This should be fixed by adding `bool(self.data_source)` to the Image model `is_user_editable` and `is_user_edited` methods.

Issue #309, opened 2019-04-11
Linkedevents should be configurable using environment


vikoivun:
I'm working on making linkedevents configurable using environment variables.

Rikuoja:
I believe #311 fixed this.

PR #311, opened 2019-04-18
Environment-based configuration

vikoivun:
Pulls all configuration from environment variables / KEY=VALUE style configuration file.

Wrt. issue #309

### DIFF
--- a/config_dev.toml.example
+++ b/config_dev.toml.example
@@ -1,0 +1,139 @@
+# Linkedevents environment configuration
+# This file defines a set of (environment) variables that configure most
+# of the functionality of linkedevents. In order for linkedevents to read
+# this file, rename it to `config_dev.toml`. As the name implies, this
+# file is supposed to be used only in development. For production use
+# we recommend setting the environment variables using the facilities
+# of your runtime environment.
+# Linkedevents reads this file by itself. However, it can also be
+# used in conjunction with the included docker-compose.yml. Then
+# you don't need to inject the file into the container. Instead
+# Docker defines environment variables that linkedevents will read.
+# Note that although this file has the extension .toml, it is actually
+# read by a parser only capable of reading key=value pairs + comments.
+# We use the .toml extension to indicate that it is readable by TOML
+# parser.
+# Following are the settings and their explanations, with useful values
+# for development:
+
+# Whether to run Django in debug mode
+# Django setting: DEBUG https://docs.djangoproject.com/en/2.2/ref/settings/#debug
+DEBUG=True
+
+# Namespace ID used for events added through the API. All events added
+# through api will have their "data_source" field set to this value and
+# their identifier will be prefixed by this value.
+SYSTEM_DATA_SOURCE_ID=ledev
+
+# Languages that this Linkedevents instance will support. This affects
+# several things and cannot be changed nilly-willy. Most importantly,
+# you will need to run `manage.py sync_translation_fields` after
+# changing this. You will also need to rebuild your elasticsearch
+# indexes: `manage.py rebuild_indexes`, but only if you are actually
+# running elasticsearch, ie. specifying ELASTICSEARCH_URL
+# Django setting: LANGUAGES https://docs.djangoproject.com/en/2.2/ref/settings/#languages
+LANGUAGES=fi,en
+
+# Configures database for Linkedevents using URL style. Format is:
+# postgis://USER:PASSWORD@HOST:PORT/NAME
+# Unused components may be left out, only Postgis is supported
+# The example below configures Linkedevents to use local PostgreSQL database
+# called "linkedevents", connecting same as username as Django is running as.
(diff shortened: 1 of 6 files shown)


Rikuoja:
What thinks @jussiarpalahti , is this the way you would want dockerified django applications to be configured?

config_dev.toml.example
```diff
@@ -0,0 +1,139 @@
...
+SYSTEM_DATA_SOURCE_ID=ledev
+
+# Languages that this Linkedevents instance will support. This affects
+# several things and cannot be changed nilly-willy. Most importantly,
+# you will need to run `manage.py sync_translation_fields` after
+# changing this. You will also need to rebuild your elasticsearch
+# indexes: `manage.py rebuild_indexes`, but only if you are actually
+# running elasticsearch, ie. specifying ELASTICSEARCH_URL
+# Django setting: LANGUAGES https://docs.djangoproject.com/en/2.2/ref/settings/#languages
+LANGUAGES=fi,en
```
Rikuoja: The default languages should be the same as those in `settings.py`, as existing migrations are done assuming that. Then, `sync_translation_fields` only updates munigeo tables and everything else works with no changes.

Therefore, the languages should be

```

    ('fi', 'Finnish'),
    ('sv', 'Swedish'),
    ('en', 'English'),
    ('zh-hans', 'Simplified Chinese'),
    ('ru', 'Russian'),
    ('ar', 'Arabic'),
)
```

with the names included. How are those names generated here?

config_dev.toml.example
```diff
@@ -0,0 +1,139 @@
...
+# for development:
+
+# Whether to run Django in debug mode
+# Django setting: DEBUG https://docs.djangoproject.com/en/2.2/ref/settings/#debug
+DEBUG=True
+
+# Namespace ID used for events added through the API. All events added
+# through api will have their "data_source" field set to this value and
+# their identifier will be prefixed by this value.
+SYSTEM_DATA_SOURCE_ID=ledev
```
Rikuoja: Why is this different from that specified in `settings.py`?

vikoivun: They are supposed to be examples for developer. Hmmh. I suppose I'll comment out the settings that would cause harm.

config_dev.toml.example
vikoivun: Again, I'll comment this out too. It is a bit of shame, that the migrations require a certain set of languages though.
See here about the language names: https://github.com/City-of-Helsinki/linkedevents/blob/28521539e285d15ca354537cf0c79a1ee2b1a594/linkedevents/settings.py#L159-L167

vikoivun: Oh. Nice you can just read it here. 😁

vikoivun: There really was no point in repeating the translations done by Django developers.

vikoivun:
> I really have no idea whether Django developers would want to add an extra layer of abstraction on top of `settings.py` and `local_settings.py`. Also, I don't know how to do here certain things (i.e. adding to the settings lists on the fly in `local_settings.py`) needed for e.g. django-debug-toolbar. So the jury is very much out on this one.

Do note that this is entirely optional. You can just not rename the file (config_dev.toml.example -> config_dev.toml) and everything will be just as it was. local_settings.py is still executed and is ready to serve for adding ye olde debugging aids.

config_dev.toml.example
Rikuoja: Yeah; though an example like `helsinki` (as opposed to `system` which is in `settings.py`) might be a sensible namespace example?

config_dev.toml.example
Rikuoja: Well, the migrations probably strictly don't require a certain set, `sync_translation_fields` might be able to remove as well as add languages. Haven't tested that. However, perhaps best use the default set?

Rikuoja:
If we can really disregard the whole env/toml thing, it is better to document that too in `settings.py` and `local_settings.py`, so django developers don't get confused by all the stuff that looks like it reads the environment instead.

config_dev.toml.example
vikoivun: Probably. Maybe some other city though. Let's go for "jyvaskyla"? Maybe the namespaces should be explained in the main README.

vikoivun:
> If we can really disregard the whole env/toml thing, it is better to document that too in `settings.py` and `local_settings.py`, so django developers don't get confused by all the stuff that looks like it reads the environment instead.

Well you can disregard it during development, but if the setting needs to be used for running linkedevents, it needs to be read from environment. settings.py is more for putting the application together (like INSTALLED_APPS, which is normally not edited when you are just running linkedevents).

local_settings.py would only be used during development and it could be really minimal. I think we could have the debug-tools-du-jour configurable from config_dev.toml too. In fact they could be automatically activated if that file exists (with an option to disable them).

Rikuoja:
Looks good, do you want to re-rebase this on master?

Rikuoja:
Damn, no need if we squash the branch anyway.

Issue #305, opened 2019-02-15
Prevent changing event publisher


Rikuoja:
Currently, the user may specify the publisher of the event at PUT to be different than the original publisher, if e.g. the user default organization has changed.

After creating, the publisher (and data source) fields should be non-editable by any user, even if the user default organization changes.

PR #313, opened 2019-04-24
Feature/allow specified event publisher

halarotu:


events/api.py
```diff
@@ -558,12 +558,14 @@ def validate_data_source(self, value):
     def validate_publisher(self, value):
         if value:
             # TODO: allow optionally declaring publisher in any user admin org (regular org for drafts)
-            if value not in (self.publisher, self.publisher.replaced_by):
+            # TODO: self.user.admin_organizations.all() or user.get_admin_organizations_and_descendants()?
+            if value not in (self.publisher, self.publisher.replaced_by) and\
+                    value not in self.user.admin_organizations.all():
```
Rikuoja: Actually, here `self.publisher` is one of the orgs in `self.user.admin_organizations.all()`.

However, we must also check `replaced_by` just in case, so perhaps better just use `if value not in (self.user.admin_organizations.all() | map(lambda x: getattr(x, 'replaced_by'))` or something similar, to check all admin_organizations and their replacements?

events/api.py
```diff
@@ -558,12 +558,14 @@ def validate_data_source(self, value):
     def validate_publisher(self, value):
         if value:
             # TODO: allow optionally declaring publisher in any user admin org (regular org for drafts)
-            if value not in (self.publisher, self.publisher.replaced_by):
+            # TODO: self.user.admin_organizations.all() or user.get_admin_organizations_and_descendants()?
```
Rikuoja: `self.user.admin_organizations.all()` should be enough, because we do not need users actually publishing new events on behalf of their descendants.

Descendants must be allowed, however, in existing events users from the top level organizations do edit. Therefore, extra check will have to be done by me so that validation at PUT will be done differently, as per issue #305 . So this is fine for now.

Rikuoja: Damn! We will need `user.get_admin_organizations_and_descendants()` after all, this will make implementing #305 a breeze :D

PR #315, opened 2019-05-06
Update readme

halarotu:


README.md
```diff
@@ -39,18 +39,28 @@ Set up project with Docker
     docker exec -it linkedevents-backend python manage.py event_import helmet --events
     # Import events from Espoo
     docker exec -it linkedevents-backend python manage.py event_import espoo --events
+    # Import organizations
+    docker exec -it linkedevents-backend python manage.py import_organizations https://api.hel.fi/paatos/v1/organization/ -s helsinki:ahjo
     ```
-6. Update search indexes:
+6. Add datasources helsinki and helfi:
```
Rikuoja: We will have to somehow automate this with @vikoivun ; while this is correct, it is not desired that the user will have to do stuff in the shell by hand.

Also, perhaps we want to make the default setup non-helsinki specific. However, the problem remains that the current UI is helsinki-specific and requires the data sources below, so we cannot solve this right away.

README.md
```diff
@@ -39,18 +39,28 @@ Set up project with Docker
     docker exec -it linkedevents-backend python manage.py event_import helmet --events
     # Import events from Espoo
     docker exec -it linkedevents-backend python manage.py event_import espoo --events
+    # Import organizations
```
Rikuoja: Perhaps we should add `(optionally)` to all these phases 5-8, and explain that the empty database works already without running any Helsinki-specific stuff!

README.md
```diff
@@ -39,18 +39,28 @@ Set up project with Docker
...
+8. (Optionally) install API frontend templates:
     * `docker exec -it linkedevents-backend python manage.py install_templates helevents`
 
-8. Start your Django server:
+9. Start your Django server:
     * `docker exec -it linkedevents-backend python manage.py runserver 0:8000`
 
 Now your project is live at [localhost:8000](http://localhost:8000)
 
+For UI, set authentication keys to local_settings.
```
Rikuoja: Perhaps we should list here separately the backend steps required to use the Helsinki UI, and make it clear that the backend works also without Helsinki UI.

README.md
```diff
@@ -223,44 +233,32 @@ CUSTOM_MAPPINGS = {
...
-                    }
-                },
-                "filter": {
-                    "voikko_filter": {
-                        "type": "voikko",
-                    }
-                }
-            }
-        }
+        'BASE_ENGINE': 'haystack.backends.simple_backend.SimpleEngine'
```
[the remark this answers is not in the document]
vikoivun: Yes. The elasticsearch is now automatically configured based on the set of languages in use. If elasticsearch URL is not specified it uses SimpleEngine.

PR #320, opened 2019-09-05
Add browsable API templates for Helsinki Linked Courses

tuomas777:
Screenshot 2019-09-05 at 22 18 41

### DIFF
--- a/linkedevents/settings.py
+++ b/linkedevents/settings.py
@@ -147,7 +147,6 @@ INSTALLED_APPS = [
     'helusers.providers.helsinki',
 
     'helevents',
-    'helcourses',
     'munigeo',
     'leaflet',
     'django_orghierarchy',
--- a/linkedevents/settings.py
+++ b/linkedevents/settings.py
@@ -55,6 +55,7 @@ env = environ.Env(
     SENTRY_ENVIRONMENT=(str, 'development'),
     COOKIE_PREFIX=(str, 'linkedevents'),
     INTERNAL_IPS=(list, []),
+    INSTANCE_NAME=(str, 'Linked Events'),
 )
 
 # Build paths inside the project like this: os.path.join(BASE_DIR, ...)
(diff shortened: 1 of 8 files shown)


linkedevents/settings.py
```diff
@@ -144,10 +147,11 @@ def get_git_revision_hash() -> str:
     'helusers.providers.helsinki',
 
     'helevents',
+    'helcourses',
     'munigeo',
     'leaflet',
     'django_orghierarchy',
-]
+] + env('EXTRA_INSTALLED_APPS')
```
Rikuoja: Where do we need extra installed apps, I wonder?

vikoivun: Däts for extensions?

Rikuoja: Nope, there is `AUTO_ENABLED_EXTENSIONS = env('AUTO_ENABLED_EXTENSIONS')` for that?

vikoivun: That enables extensions for every request, yes? Instead of activating by url parameter, yes?

Rikuoja: Or, if this setting is meant to be changed whenever an extension is auto enabled, then surely `helcourses` should not be in the regular setup, but in the `env('EXTRA_INSTALLED_APPS')`?

vikoivun: Ah. Indeed.

Rikuoja: Yes it does. So, the question is, why do we need extra installed apps, if extensions are already installed?

tuomas777: Linked Courses needs both `"extension_course"` in `INSTALLED_APPS` and `"course"` in `AUTO_ENABLED_EXTENSIONS`, the latter does nothing without the former. `AUTO_ENABLED_EXTENSIONS` is about auto enabling an installed extension on every request, just like Ville said. 

You are right that there is no need for `helcourses` to be installed by default, I'll change that.

Also, I'm not sure if this `EXTRA_INSTALLED_APPS` method is at all the best for allowing `INSTALLED_APPS` to be modified from env, feel free to propose a better solution :)

Rikuoja: Okay, let's include instructions on how to install the courses extension as an example under `https://github.com/City-of-Helsinki/linkedevents#event-extensions`.

As courses is the only extension at the moment, we should just have instructions for that in the readme. The generic extension instructions do not really help anyone deploying courses.

tuomas777:
Replaced by https://github.com/City-of-Helsinki/linkedevents/pull/321

Issue #325, opened 2019-10-15
Expose event `created_by` and `last_modified_by` fields to admins of the same organization tree


Rikuoja:
Within an organization tree, admins often have the need to figure out who is responsible for creating or editing a specific event.

Therefore, *if* the user is *admin* (not a regular user), the API should display the names of the user who created or edited an event *within the same organization hierarchy*, i.e. child, parent or child of any parent, however distant.

Rikuoja:
Required by https://github.com/City-of-Helsinki/linkedevents-ui/issues/433

Rikuoja:
Do note that affiliated organizations do not count; only display the info for admins of *regular* organizations.

Rikuoja:
As detailed in https://github.com/City-of-Helsinki/linkedevents/pull/326 , this issue is performance-wise easily implementable for listings of `n` events as long as we do *not* consider possible replacement organizations outside the admin organization hierarchy. Such replacements across different root nodes should not be present in the hierarchy as it is.

Rikuoja:
Implemented in #326 , replacements and all.

PR #326, opened 2019-10-16
Implement issue #325

Rikuoja:


Rikuoja:
Refactoring and additional testing in progress.

Rikuoja:
Looks like implementing #325 for a list of `n` events can be done easily with a constant number of database queries in the MPTT tree if we do *not* consider that any organization can be replaced by any other.

Therefore, I suggest that we do not consider cases where an organization under one MPTT root node has been replaced with an organization under another MPTT root node. This is not an intended use case anyway, it is just an artefact of the additional foreign key field in the model that replacements could go across the MPTT hierarchy between any two objects.

It creates crosslinks in the MPTT tree that no algorithm can traverse easily, and would require evaluating both user admin organization queryset *and* event publisher queryset to look for replacement organizations, increasing the number of database queries required to evaluate the permission from constant to `O(n)` with the number of events and `O(n)` with the number of user admin organizations.

tituomin:
LGTM

PR #328, opened 2019-10-21
Make development docker Work Again

vikoivun:
Cleanups and upgrades for dev Dockerization

Postgres was upgraded to version 10, with Postgis 2.5.
Added a nifty endpoint with migrate and runserver commands.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,10 +6,8 @@ services:
             POSTGRES_USER: linkedevents
             POSTGRES_PASSWORD: linkedevents
             POSTGRES_DB: linkedevents
-        ports:
-            - "5432:5432"
         volumes:
-            - postgres-data-volume:/var/lib/postgres/data
+            - linkedevents-postgres:/var/lib/postgres/data
         container_name: linkedevents-db
 
     django:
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -24,17 +16,16 @@ services:
...
             - "8000:8000"
         depends_on:
             - postgres
-            - elastic
         container_name: linkedevents-backend
 
 volumes:
-        postgres-data-volume:
-        django-media-volume:
+        linkedevents-postgres:
```
igordavydsson: docker prefixes all volumes with the folder's name, so the volume will be called `linkedevents_linkedevents-postgres`

docker/django/Dockerfile
```diff
@@ -1,6 +1,7 @@
 FROM python:3.6
 
-ENV PYTHONUNBUFFERED 0
+# Force flush all messages immediately, not really Docker specific
+ENV PYTHONUNBUFFERED 1
```
igordavydsson: why does no one like unbuffered output? 😓 userful for development

docker-compose.yml
vikoivun: Argh. Forgot about that. I'll kick out the prefix. Is the `volume` postfix useful in some situations? Ie. are they listed in context where it is not immediately obvious they are volumes?

igordavydsson: I think `volume` postfix can be dropped.

sam-hosseini: And Ville, if you could please pull latest changes before working again on commits, it'd be great as my Git global config was making my commits anonymous (which was intentional 😅) but I fixed it and had to do a force push (only and only this time) 👍

vikoivun: All right. It seems to have confused git a wee bit, but nothing major.

sam-hosseini:
https://github.com/City-of-Helsinki/linkedevents/pull/356 has been inspired by this PR and therefore includes all the changes already, hence the conflicting files. If you see fit, this can be closed :)

PR #329, opened 2019-10-22
Remove elastic search from Docker

sam-hosseini:
I recommend reviewing commit-by-commit :)


sam-hosseini:
> Okay, if Elasticsearch is removed from the Docker configuration, then I believe step 11 of the docker instructions (optionally setting up Elasticsearch) should be removed too?

Correct. Done in aa8d963

PR #330, opened 2019-10-24
Update readme about linked events purpose

sam-hosseini:
Based on @vikoivun 's work in https://github.com/City-of-Helsinki/linkedevents/tree/feature/docs-config-management, I updated only the introduction part of README.md as I wanted my changes to remain small and manageable in scope so I can go to other parts of README.md step by step.


Rikuoja:
Within current usage statistics, > 50 % of the event data is fetched from other sources through importers, and < 50 % (but growing) of the events are actually published in Linkedevents through the API by other systems and the UI.

Rikuoja:
You can use this picture (courtesy of @juhasuv) as a basis for a general English diagram, though here there are no arrow directions either. 
image

sam-hosseini:
Thank you @Rikuoja for reviewing 👍

>  I would add the mention that in addition to aggregating data from other sources, Linkedevents also provides a two way event API, i.e. it supports fetching event data but also publishing new event data by PUT/POST in the API /event endpoint, with granular user permissions and a hierarchical organization structure supporting different publishing rights for different organizations.

Done in 7f3aa7c9fe930f6482825622a62da2bc4d9dd9ff

> Therefore, also the diagram should probably have one-way arrows from the importers to the database, but two-way arrows across the internet, API and database.

I updated the diagram diagram

sam-hosseini:
> You can also use this picture (courtesy of @juhasuv) as a basis for a general English diagram if you wish, though here there are no arrow directions either, and this is specific to the existing importers and API users. So event data here flows from sources at the left to users on the right, and you can see the Django importers as well as the users who POST data to the REST API directly.

That's a very good picture. If I understood correctly, the first 4 elements in tapahtumattiedot (Syöttölomake UI until Taidemuseo) directly write new events to the API using their systems, and the last 4 elements (Kulttuuritalot until Espoo) event information will be fetched by our Python importers, right?

In addition in explains the authentication (using the Tunnistamo) and also where we get places information (using Toimipisterekisteri).

If so, I think it'd be useful to maybe have a Helsinki-related-data diagram in another part of README.md or in another file (in another PR), and I can add it to our board and prioritize it using the PO.

Rikuoja:
> That's a very good picture. If I understood correctly, the first 4 elements in tapahtumattiedot (Syöttölomake UI until Taidemuseo) directly write new events to the API using their systems, and the last 4 elements (Kulttuuritalot until Espoo) event information will be fetched by our Python importers, right?

You are correct, the first 4 are active API clients. The UI is the only client that authenticates with Tunnistamo though.

> In addition it explains the authentication (using the Tunnistamo) and also where we get places information (using Toimipisterekisteri).
> 
> If so, I think it'd be useful to maybe have a Helsinki-related-data diagram in another part of README.md or in another file (in another PR), and I can add it to our board and prioritize it with the PO's input as well.

Yeah, would be useful to have the data diagram that would actually contain *all* possible importers that may be used currently (including Harrastushaku). With Harrastushaku, though, the courses extension needs to be installed, and the user may be slightly confused, since in actuality there are two instances:

1) api.hel.fi/linkedevents/v1 which run all the event importers *except* harrastushaku, and
2) api.hel.fi/linkedcourses/v1 which *only* runs harrastushaku and kulttuuritalot (with the --courses parameter) importers

The background data, i.e. places and YSO keywords and helsinki organization importer (which is missing from the picture!) are run in both instances though.

PR #332, opened 2019-10-28
Update linked events local dev env doc

sam-hosseini:
You may find reviewing commit-by-commit easier :)

I didn't update the Helsinki-specific installation section because of two reasons:

* size of PR was already getting big
* `event_import tprek --places` fails because of `HAYSTACK_CONNECTIONS` in `local_settings.py`.

I'll fix the failing command and update the instructiosn of README.md in another PR :)


README.md
```diff
@@ -71,27 +82,18 @@ Steps 5-10 are needed if you wish to use location, address or event data from th
...
-Finish the install:
-
-11. Start your Django server:
-    * `docker exec -it linkedevents-backend python manage.py runserver 0:8000`
-
-Now your project is live at [localhost:8000](http://localhost:8000)
-
-For UI, set authentication keys to local_settings.
-
 Installation for development
```
Rikuoja: Okay, since above we have the docker instructions, probably better change this headline to "Development installation on physical or virtual machine" or something similar, to clarify the difference between the two instructions? Also, since the commands are almost identical, I believe in the future these non-docker steps should be merged with the docker instructions above.

sam-hosseini: > As regards the haystack problem, I believe the settings should be removed from local_settings.py.template, or does @vikoivun agree?

I checked https://github.com/City-of-Helsinki/linkedevents/blob/master/linkedevents/settings.py#L309 as well. I removed it in 525299a but I can revert the commit if it's needed.

> Okay, since above we have the docker instructions, probably better change this headline to "Development installation on physical or virtual machine" or something similar, to clarify the difference between the two instructions?

Done in 6ca4fb1

 > Also, since the commands are almost identical, I believe in the future these non-docker steps should be merged with the docker instructions above.

I think so too. I can do that in the future after fixing the HH importer (and the pre-reqruisites before it) 👍

PR #334, opened 2019-10-29
Update Helsinki specific installation docs

sam-hosseini:
I made a new django management command to entirely setup all the data needed for a local Helsinki installation of a Linked Events application.

### DIFF
--- a/events/management/commands/setup_helsinki.py
+++ b/events/management/commands/setup_helsinki.py
@@ -1,0 +1,46 @@
+from django.core.management.base import BaseCommand, CommandError
+from django.core.management import call_command
+from django.conf import settings
+
+
+class Command(BaseCommand):
+    help = 'Setup all data needed for boostrapping a Helsinki specific installation'
+
+    def handle(self, *args, **options):
+        if settings.DEBUG is not True:
+            raise CommandError('This command can be run only when DEBUG is True')
+
+        # Import general Finnish ontology (used by Helsinki UI and Helsinki events)
+        call_command('event_import', 'yso', '--all')
+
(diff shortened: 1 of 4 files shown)


events/management/commands/setup_helsinki.py
```diff
@@ -0,0 +1,46 @@
...
+        call_command('add_helfi_topics')
+
+        # Import places from Helsinki metropolitan region service registry (used by events from following sources)
+        call_command('event_import', 'tprek', '--places')
+
+        # Import places from Helsinki metropolitan region address registry (used as fallback locations)
+        call_command('event_import', 'osoite', '--places')
+
+        # Import events from Helsinki metropolitan region libraries
+        call_command('event_import', 'helmet', '--events')
```
Rikuoja: Here, the problem is that not all instances (e.g. linkedcourses) might want the actual events. Which events are published depends on the instance.

Therefore, I would separate this in two parts:

1) one command for installing Helsinki support data (places, addresses, keywords, audiences, topics, organization, template, municipalities, districts). Of these, the municipality importer and the YSO keywords could be useful outside Helsinki as well, so I'm not sure we want to hide those functionalities inside the Helsinki command either.

2) separate commands for importing actual events in each dataset. Which importers to run here depends on what is the purpose of your installation: publishing Helsinki events, publishing Espoo events, publishing all capital region events, or perhaps publishing courses instead?

sam-hosseini:
My understanding of importers and linked events and linked courses have evolved a lot since this PR and therefore I'm closing this in favor of https://github.com/City-of-Helsinki/linkedevents/pull/347 in which we'd leave it up to the reader to decide what importers to use and not the management command.

PR #340, opened 2019-10-31
Documenting importers

sam-hosseini:
I wanted to get the ball rolling with publishing the first importer documentation, to collect feedback if any arises.

### DIFF
--- a/importers.md
+++ b/importers.md
@@ -27,7 +27,7 @@ YSO, short for *Yleinen suomalainen ontologia*, is basically a +30,000 collectio
 >
 > From https://finto.fi/yso/en/. You can browse the concepts there if you wish
 
-Basic example would be, let's say a guitar player coming to Helsinki to perform in a concert and use suana afterwards.
+Basic example would be, let's say a guitar player coming to Helsinki to perform in a concert and use sauna afterwards.
 
 We might categorize it with "concert", "popular music", and "sauna" *concepts*.
 
--- a/importers.md
+++ b/importers.md
@@ -5,7 +5,7 @@
 * [YSO - General Finnish ontology **(Required)**](#yso---general-finnish-ontology-required)
   * [What is it?](#what-is-it)
   * [What depends on it?](#what-depends-on-it)
-  * [What does the importer do?](#what-does-the-importer-do)
+  * [How to use it?](#how-to-use-it)
 
 <!-- vim-markdown-toc -->
 
@@ -37,7 +37,7 @@ These *concepts* come from YSO.
 Linked Events application has *events*. Events must be described by *concepts*.
 Therefore, all events depend on YSO.
 
-### What does the importer do?
+### How to use it?
   ```bash
   python manage.py event_import yso --keywords --all
   ```
--- a/importers.md
+++ b/importers.md
@@ -14,10 +14,11 @@
 ### What is it?
 First, some terminology is in order. Starting with *ontology*:
 
-> Ontology is the philosophical study of being. More broadly, it studies concepts that directly relate to being, in particular
-> becoming, existence, reality, as well as the basic categories of being and their relations.
+> In computer science and information science, an ontology encompasses a representation,
+> formal naming and definition of the categories, properties and relations between the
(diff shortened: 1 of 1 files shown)


importers.md
```diff
@@ -0,0 +1,44 @@
...
+
+## YSO - General Finnish ontology **(Required)**
+
+### What is it?
+First, some terminology is in order. Starting with *ontology*:
+
+> Ontology is the philosophical study of being. More broadly, it studies concepts that directly relate to being, in particular
+> becoming, existence, reality, as well as the basic categories of being and their relations.
+>
+> From https://en.wikipedia.org/wiki/Ontology
```
vikoivun: While this is decidedly cool, you stilll might want to use the computer sciencey definition instead: https://en.wikipedia.org/wiki/Ontology_(information_science)

importers.md
```diff
@@ -0,0 +1,44 @@
...
+
+YSO, short for *Yleinen suomalainen ontologia*, is basically a +30,000 collection of **concepts** in Finnish culture, translated in 3 languages.
+
+> General Finnish Ontology YSO is a trilingual ontology consisting mainly of general concepts.
+> YSO has been founded on the basis of concepts in Finnish cultural sphere.
+>
+> From https://finto.fi/yso/en/. You can browse the concepts there if you wish
+
+Basic example would be, let's say a guitar player coming to Helsinki to perform in a concert and use suana afterwards.
```
vikoivun: Typo here.

importers.md
```diff
@@ -0,0 +1,44 @@
...
+
+These *concepts* come from YSO.
+
+### What depends on it?
+Linked Events application has *events*. Events must be described by *concepts*.
+Therefore, all events depend on YSO.
+
+### What does the importer do?
+  ```bash
+  python manage.py event_import yso --all
```
vikoivun: This is, of course, an artefact of the LE importer structure, but perhaps still document the usage as:
`python manage.py event_import yso --keywords`
`--all` is more future proof though. Better might be to support default import categories for importers (ie. yso would import keywords by default)

importers.md
```diff
@@ -0,0 +1,44 @@
...
+
+We might categorize it with "concert", "popular music", and "sauna" *concepts*.
+
+These *concepts* come from YSO.
+
+### What depends on it?
+Linked Events application has *events*. Events must be described by *concepts*.
+Therefore, all events depend on YSO.
+
+### What does the importer do?
```
vikoivun: Looks like this heading should be "How to use the importer". Of course this might evolve as you progress through the importers.

importers.md
sam-hosseini: Done in ab9735d3005db6493e2ff0fa2b3962594c392afa

importers.md
sam-hosseini: Fixed in 0f12c84abaf8f4db802a9130ab32f12f15d13225 , thanks!

importers.md
sam-hosseini: Done in 7f6e243410c71841cc8fc066db7ec75cb957783e

importers.md
sam-hosseini: Done in 7056f392f1a09e38d569795945f5dffb1a27afe9

Rikuoja:
Also, in the contribution guidelines down the path, we might mention that we perhaps want such description for any new importers people offer to the mix. To keep this file up to date.

Rikuoja:
Damn. I am too late, but perhaps we should add some mention of data synching and scheduled runs of importers? E.g. examples which importers should be run daily etc.?

For example, the ontology importer (in our setup) is run daily. When the YSO ontology changes, the importer automatically deprecates old concepts and moves events related to old concepts to up-to-date concepts instead. Keyword replacement is not (yet) saved into database, so we cannot refer any old keyword queries to new keywords automatically, but I made an issue of it: https://github.com/City-of-Helsinki/linkedevents/issues/341

sam-hosseini:
I added the scheduled runs to both importers documentations in https://github.com/City-of-Helsinki/linkedevents/pull/347 👍

PR #348, opened 2019-11-12
TH-40 HH importer location duplication fix

sam-hosseini:
Please review commit-by-commit if you think the changes are big :)

### DIFF
--- a/events/importer/harrastushaku.py
+++ b/events/importer/harrastushaku.py
@@ -91,6 +91,56 @@ class HarrastushakuImporter(Importer):
                 message = e if isinstance(e, HarrastushakuException) else traceback.format_exc()
                 logger.error('Error handling location {}: {}'.format(location.get('id'), message))
 
+    def map_harrastushaku_location_ids_to_tprek_ids(self, harrastushaku_locations):
+        '''
+        Example mapped dictionary result:
+        {
+            '95': 'harrastushaku:95',
+            '953': 'harrastushaku:953',
+            '968': 'tprek:20479',
+            '97': 'tprek:8062',
+            '972': 'tprek:9079',
+            '987': 'harrastushaku:987',
+            '99': 'tprek:8064',
+        }
+        '''
+        result = dict()
+
+        for harrastushaku_location in harrastushaku_locations:
+            harrastushaku_location_id = harrastushaku_location['id']
+
+            strict_filters = {
+                'id__startswith': self.tprek_data_source,
+                'name': harrastushaku_location['name'],
+                'address_locality': harrastushaku_location['city'],
+                'postal_code': harrastushaku_location['zip'],
+                'street_address': harrastushaku_location['address'],
+            }
+            flexible_filters = {
+                'id__startswith': self.tprek_data_source,
+                'address_locality': harrastushaku_location['city'],
+                'postal_code': harrastushaku_location['zip'],
+                'street_address': harrastushaku_location['address'],
+            }
+
+            strict_matched_places = Place.objects.filter(**strict_filters)
+
+            if len(strict_matched_places) >= 1:
+                harrastushaku_location_mapped_id = strict_matched_places.first().id
+
+            elif len(strict_matched_places) == 0:
+                flexible_matched_places = Place.objects.filter(**flexible_filters)
+
+                if len(flexible_matched_places) >= 1:
+                    harrastushaku_location_mapped_id = flexible_matched_places.first().id
+
+                elif len(flexible_matched_places) == 0:
+                    harrastushaku_location_mapped_id = f'{self.data_source.id}:{harrastushaku_location["id"]}'
+
+            result[harrastushaku_location_id] = harrastushaku_location_mapped_id
+
+        return result
+
     def import_courses(self):
         """Import Harrastushaku activities as Courses
 
--- a/events/importer/harrastushaku.py
+++ b/events/importer/harrastushaku.py
@@ -248,7 +248,7 @@ class HarrastushakuImporter(Importer):
     @transaction.atomic
     def handle_location(self, location_data):
         harrastushaku_location_id = location_data.get('id')
-        harrastushaku_location_mapped_id = self.mapped_harrastushaku_locations.get(harrastushaku_location_id)
+        harrastushaku_location_mapped_id = self.location_id_to_place_id.get(harrastushaku_location_id)
 
         if harrastushaku_location_mapped_id.startswith(self.tprek_data_source.id):
             return
--- a/events/importer/harrastushaku.py
+++ b/events/importer/harrastushaku.py
@@ -195,12 +195,7 @@ class HarrastushakuImporter(Importer):
             logger.warning('No location data fetched, aborting course import.')
             return
 
-        self.location_id_to_place_id = {
-            l['id']: '{}:{}'.format(self.tprek_data_source.id, l['tpr_id'])
-            if l.get('tpr_id')
-            else '{}:{}'.format(self.data_source.id, l['id'])
-            for l in locations
-        }
+        self.location_id_to_place_id = self.map_harrastushaku_location_ids_to_tprek_ids(locations)
 
         activities = self.fetch_courses()
         if not activities:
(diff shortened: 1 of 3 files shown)


events/importer/harrastushaku.py
```diff
@@ -91,6 +92,56 @@ def import_places(self):
...
+            elif len(strict_matched_places) == 0:
+                flexible_matched_places = Place.objects.filter(**flexible_filters)
+
+                if len(flexible_matched_places) >= 1:
+                    harrastushaku_location_mapped_id = flexible_matched_places.first().id
+
+                elif len(flexible_matched_places) == 0:
+                    harrastushaku_location_mapped_id = f'{self.data_source.id}:{harrastushaku_location["id"]}'
+
+            result[harrastushaku_location_id] = harrastushaku_location_mapped_id
```
tuomas777: The above `elif`s are unnecessary, if `len(something)` is not `>=1`, then it must be `0` :)

Actually the whole structure seems a bit overly complicated to me, how about just

```
tprek_place = Place.objects.filter(**strict_filters).first() or Place.objects.filter(**flexible_filters).first()

if tprek_place:
    result[harrastushaku_location_id] = tprek_place.id
else:
    result[harrastushaku_location_id] = f'{self.data_source.id}:{harrastushaku_location["id"]}'
```

tuomas777: One other thing, using f-strings limits Python version to >= 3.6. Whether that is ok or not I don't know, required Python version for LE doesn't seem to be specified anywhere. Python 3.5 isn't *that* old, it is the default in Ubuntu 16.04, which still seems to be used quite a bit in some circles.

AFAIK other LE code runs at least on 3.5, so we probably shouldn't break that by using just a couple f-strings (I've been in a situation where exactly one f-string caused a lot of headache 🙂). Unless of course it will be decided that from now on LE requires at least Python 3.6. But this is something for @Rikuoja to decide.

events/importer/harrastushaku.py
```diff
@@ -73,16 +73,17 @@ def setup(self):
...
 
-        - If a place in Harrastushaku does not have "tprek_id", it will be
-          imported having data source "harrastushaku".
+        - If, however, we cannot find a match, location object will be imported with data source "harrastushaku".
         """
         logger.info('Importing places...')
 
         locations = self.fetch_locations()
         logger.debug('Handling {} locations...'.format(len(locations)))
+        self.mapped_harrastushaku_locations = self.map_harrastushaku_location_ids_to_tprek_ids(locations)
```
tuomas777: Hmm, is there a reason why this  cannot be named `self.location_id_to_place_id`, as it seems to be exactly the same thing?

events/importer/harrastushaku.py
sam-hosseini: > The above elifs are unnecessary, if len(something) is not >=1, then it must be 0 :)

You are right, my bad. I was trying to be more explicit than implicit, but the way I was trying to do that was making everything else complex. Your suggested approach is much cleaner and easier to understand, and I replaced it with mine in 0fe669c6a508865c15af10b3c563109f9af203ba

> One other thing, using f-strings limits Python version to >= 3.6. Whether that is ok or not I don't know, required Python version for LE doesn't seem to be specified anywhere. Python 3.5 isn't that old, it is the default in Ubuntu 16.04, which still seems to be used quite a bit in some circles.

I thought Linked Events supports Python 3.6 because the `Dockerfile` image is using `python:3.6`, but again, my bad, as I knew that we don't do Docker deployments at the moment and therefore `Dockerfile` is a not reference to anything in production. I updated the code in cab321dbe67dbade40599d3d673c3caa4dc78210

events/importer/harrastushaku.py
sam-hosseini: Correct. Done in 935714dac1615e46a1f88e62f58385df92124ded

events/importer/harrastushaku.py
Rikuoja: Yeah, 3.6 is our minimum. Mostly because we soon want to upgrade to newer django versions, which will drop 3.5 support anyway, and we don't run tests on 3.5 any more.

tuomas777: Ok. That would be a good addition to the README :)

tuomas777:
@Rikuoja you need to run HH place import first.

sam-hosseini:
@Rikuoja I was able to reproduce the error you mentioned by running this first:

```bash
python manage.py event_import harrastushaku --courses
```

Before having imported the places using :

```bash
python manage.py event_import harrastushaku --places
```

However, once I imported the places *first*, and then tried to importe the courses, the results were as following:

- There's around ~850 `activities` (which I think is HH terminology for courses) in Harrastushaku.
- Some of these activities are recurring (which means that there will be many events created in the `Event` table for a singular `activity`) and some are non-recurring.
- At the end of the import, ~10,700 events were created which are from HH.
Screenshot 2019-11-19 at 17 37 26

- ~7700 events of which have `tprek` locations.
Screenshot 2019-11-19 at 17 38 05

- ~3000 events of which are `harrastushaku` locations.
Screenshot 2019-11-19 at 17 38 46

Based on this, I confirm that courses from both `tprek` and `harrastushaku` locations were imported 👍 

Is that okay for you?

sam-hosseini:
> Okay, then I believe we should add instructions for the user of the importer to first import harrastushaku places and only then harrastushaku courses.
 

Sure. Done in d37b0c9

> I guess the only other alternative would be that we only import those harrastushaku places that have no tprek duplicate, and that would mean more importer work.

@Rikuoja I'm not sure if I'm following. This PR, and specifically 4bb4b1f17fcd0d238d93c19d0c96690ded0bf4c7, is ensuring that if we find a close enough match of any HH location to tprek, that location would *not* be imported to database. So:

> we only import those harrastushaku places that have no tprek duplicate

is already done.

> Also (and this is more of a design and data quality question), importing all the harrastushaku places will then result also all duplicate places present in the backend, even though they have no courses. So, when designing the Linkedcourses UI, care should be taken that the user is first and foremost given the tprek options, so people don't start using harrastushaku locations when there is a corresponding tprek location available.
 
Similar to above, if this PR would do what we intend it to do, there will not be any location that would be imported from HH which would have a corresponding tprek location.

However, about importing locations which do not have courses in them, I see your point 👍 I can implement the necessary changes to prevent importing of the locations which do not have courses after consulting the product owner 👍

Rikuoja:
> > Okay, then I believe we should add instructions for the user of the importer to first import harrastushaku places and only then harrastushaku courses.
> 
> Sure. Done in [d37b0c9](https://github.com/City-of-Helsinki/linkedevents/commit/d37b0c9a5a3cf682d1421201845add48d8a441b5)
> > we only import those harrastushaku places that have no tprek duplicate
> 
> is already done.

I'm not sure I follow. Here 1) you say the user should import *all* HH places first, and then 2) you say that *only* those places that have no tprek duplicate are imported. Which one is true, 1) or 2)?

Rikuoja:
I believe it is still possible the user might need some HH locations without courses that are not present in tprek. Therefore, it is not a problem in my opinion if there are HH locations without courses present. It will offer a backup location option if a location for some reason does not exist in tprek.

However, the preferred way of introducing new course locations should be to add them in tprek. This is the same way as Linkedevents does it: many city departments actually have permissions to edit and add places to Tprek, so in the future, if a course location is missing, it might be more useful  to add it to tprek, not just use HH location. But this is up to the product owner to consider in the future.

sam-hosseini:
@Rikuoja 

> I believe it is still possible the user might need some HH locations without courses that are not present in tprek. Therefore, it is not a problem in my opinion if there are HH locations without courses present. It will offer a backup location option if a location for some reason does not exist in tprek.

Good point 👍 I will not raise the issue with the PO then.

> However, the preferred way of introducing new course locations should be to add them in tprek. This is the same way as Linkedevents does it: many city departments actually have permissions to edit and add places to Tprek, so in the future, if a course location is missing, it might be more useful to add it to tprek, not just use HH location. But this is up to the product owner to consider in the future.

Yes, I understand that it is important that we would have a single source of truth for locations (tprek) and only use HH locations if it's not possible to add locations to tprek for some reason. I'll make sure this understanding gets taken into account in all of future decisions if I'm present (and listening carefully :D ) 👍

PR #353, opened 2019-11-25
Devops improvements

sam-hosseini:
This PR, in a commit-by-commit fashion, does the following:

- Separates different environments in docker-compose files:
    * `docker-compose.yml`: this is the `base` configuration file which would contain configurations we would like to share among all environments.
    * `docker-compose.override.yml`: this is the local development environment file. `override` is docker-compose convention. Quote from [understanding multiple Compose files](https://docs.docker.com/compose/extends/):

      > By default, Compose reads two files, a docker-compose.yml and an optional docker-compose.override.yml file. By convention, the docker-compose.yml contains your base configuration. The override file, as its name implies, can contain configuration overrides for existing services or entirely new services. 

    * For local development environment, it is still enough to just do `docker-compose up`.
    * For deployment with future other docker-compose files, one can do:
    `docker-compose -f DOCKER_COMPOSE_FILE_NUMBER_1 -f DOCKER_COMPOSE_FILE_NUMBER_2`

- Add support for `waiting` for postgres `container` to be `initialized`. In docker-compose, `depends_on` waits only until the target container is `ready`, and readiness does not necessarily mean fully initialized. In our case, it takes ~5 seconds for Postgres server to initialize with postgis extension, and we must wait for this many seconds before Django is able to connect to the databse, otherwise Django would crash.

- Implements `local_development_environment` django management command which would be run automatically when someone does `docker-compose up` (which would read the `base` docker-compose file and `override`). This eliminates the need to apply migrations and translations syncronizations manually everytime when we drop the database in local development environment (which I do many many times).


events/management/commands/local_development_environment.py
```diff
@@ -0,0 +1,18 @@
+from django.core.management.base import BaseCommand, CommandError
+from django.core.management import call_command
+from django.conf import settings
+
+
+class Command(BaseCommand):
+    def handle(self, *args, **options):
```
igordavydsson: I think, these are best be handled by `docker-entrypoint.sh` too

igordavydsson:
There is a good rule of 50/72 (or 72/72 in its less strict form) for titles and descriptions of commits, which makes commits look cleaner IMO. No need to change your commits now, unless you feel like it yourself 😅 
It comes from the Linux kernel dev community and here is the relevant part of the [style guide](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/process/submitting-patches.rst?id=bc7938deaca7f474918c41a0372a410049bd4e13#n664), while here is [Linus' own explanation](https://github.com/torvalds/linux/pull/17#issuecomment-5661185) for why it is like this

sam-hosseini:
@igordavydsson thanks for the wonderful review, I'm enlightened :) 👍  

About the git commit messages, I was not aware of `summary` and `description` character limit conventions. Sounds reasonable to me. I updated my workflow in https://github.com/sam-hosseini/dotfiles/commit/1d70621cc68d1d98d750d8266c8cee7ad65d211b and from now on, my editor highlights for me if I'm going above the line limit, which should ensure that character limits won't be exceeded anymore.

About the `wait-for-it.sh` script, I think according to our internal Dockerfile guidelines, `wait-for-it.sh` is the convention. You can also see my conversation about it in the `infra` channel.

About the `docker-entrypoint.sh`, what a neat idea. I was unaware of our base images and the Dockerfile guidelines before this PR.

Since I need to make lots of changes according to the guidelines and the fact that commit messages are not looking good, I'll just create another neat PR with proper messages, and send that one for review 👍

PR #356, opened 2019-12-13
 TH-152 Dockerization

sam-hosseini:
This PR is already getting too big, so I'll do the changes necessary so that Linked Events can be deployed to KuVa infra in another PR.

@Rikuoja @vikoivun all files changed are Docker related, except splitting of `requirements.in` into two files, one for application packages in `requirements.in` and one for dev packages in `requirements-dev.in`. ~~So this PR should not cause any changes to the application or deployment of it in Kanslia infra, except the script/mechanism in which you install the packages, which would need to install dev packages now as well~~. Update: actually the server doesn't need to install the dev packages anyway, my bad 😄 . All is Docker related now :)

@igordavydsson I've read the best practices in our internal Jira and used open-city-profile as inspiration, but I implemented my own changes on top. Would be good to get your input again on this Dockerization 👍


docker/django/Dockerfile
```diff
@@ -1,18 +1,45 @@
...
 
-WORKDIR /code
+# ==============================
+FROM base_stage AS development
+# ==============================
 
-COPY . /code
+RUN apt-install.sh gdal-bin build-essential
+
+COPY --chown=appuser:appuser requirements.txt /app/
```
igordavydsson: you could move this command to the `base_stage` to avoid repeating it.

igordavydsson: as well as avoid repeating `pip install --no-cache-dir -r /app/requirements.txt`

docker/django/Dockerfile
```diff
@@ -1,18 +1,45 @@
...
+RUN pip --no-cache-dir install -r /app/requirements.txt
+
+COPY --chown=appuser:appuser requirements-dev.txt /app/
+RUN pip --no-cache-dir install -r /app/requirements-dev.txt
+
+COPY --chown=appuser:appuser . /app/
```
igordavydsson: now that you changed the owner of the files, it would be a good idea to change the user within the container:
```
USER appuser
```

docker-compose.yml
```diff
@@ -1,32 +1,36 @@
...
         volumes:
-            - .:/code
+            - .:/app:cached
             - django-media-volume:/var/media/
         ports:
-            - "8000:8000"
-        depends_on:
-            - postgres
-        container_name: linkedevents-backend
+            - "127.0.0.1:8000:8000"
```
igordavydsson: this is a more secure option, yes, but it might cause issues, if/when you'll try to connect to this container via a docker network. e.g. if you will have several containerized City-of-Helsinki projects trying to talk to each other on `helsinki` network. then this container will not be accessible for the others.

docker/postgres/Dockerfile
```diff
@@ -1,11 +1,8 @@
...
+ARG POSTGRES_VERSION=9.6
+ARG POSTGIS_VERSION=2.5
+ARG IMAGE_VARIANT=alpine
 
-RUN apt-get update \
-    && apt-get install --no-install-recommends -y \
-        postgis postgresql-9.5-postgis-2.4 postgresql-9.5-postgis-2.4-scripts
+FROM helsinkitest/postgis:${POSTGRES_VERSION}-${POSTGIS_VERSION}-${IMAGE_VARIANT}
 
-RUN localedef -i fi_FI -c -f UTF-8 -A /usr/share/locale/locale.alias fi_FI.UTF-8
```
igordavydsson: there were some issues with locales, when I created the inital Dockerfiles. you sure it works now fine for Finnish?

docker/django/Dockerfile
sam-hosseini: I've intentionally written the `development` stage like above. I could write this in the `base_stage`:

```Dockerfile
COPY --chown=appuser:appuser requirements.txt /app/

RUN apt-install.sh build-essential gdal-bin \
    && pip install --no-cache-dir -r /app/requirements.txt \
    && apt-cleanup.sh build-essential gdal-bin
``` 

But then, if `requirements.txt` changes, `COPY` directive and anything after that invalidates. Meaning that installation of `build-essential gdal-bin` has to be done again.

At the moment, with the way `development` stage is written, if `requirements.txt` changes:

```Dockerfile
RUN apt-install.sh gdal-bin build-essential
```

remains valid. It's IMO important to keep the `development` stage build as fast as possible, as the need to rebuild images in normal dev work is frequent enough.

docker/django/Dockerfile
sam-hosseini: Surely. I donno why I was under the wrong assumption that `USER` was already set in the parent Dockerfile, but when I double-checked, it wasn't. Thanks for catching this, done in 
8cfd723 👍

docker-compose.yml
sam-hosseini: hmm.. I've been under the assumption that when using the [short syntax](https://docs.docker.com/compose/compose-file/#ports) for ports in the form of `HOST:CONTAINER`, what's on the left only matters to the host machine and containers can talk with each other in their own "Docker" network (I don't know the correct term to refer to this "Docker" network).. and so in this case, a frontend container running on port 3000 can talk with this linkedevents backend on port 8000 because of the right side of the short syntax. 

Is that incorrect?

docker/postgres/Dockerfile
sam-hosseini: No, I'm not sure it works for Finnish now. I wanted to face the issue first, however, and then add those locale definitions. So far, the builds succeed, the `yso` and `tprek` importers add a lot of Finnish content to database and it works just fine, so I didn't have a reason to add those yet 🤷‍♂

docker-compose.yml
igordavydsson: ah, yes, you are right. checked it just now and docker containers use the port on the right side, no matter if it is an explicit network (like the `helsinki` one we have in the newer dockerized projects) or an implicit one (like the one docker creates for the containers from a `docker-compose.yml`). 👍

docker/postgres/Dockerfile
igordavydsson: ok, it might be okay then. let's go with your version 🙂

docker/django/Dockerfile
igordavydsson: if it would have been set in the parent Dockerfile, then some of the `RUN` commands would not work in your Dockerfile, as they would require `root` access 🙂

docker/django/Dockerfile
igordavydsson: well, I guess it boils down to preference. on the one hand you have faster builds, if you install some libraries in a layer before `COPY requirements* ...` (but with a relatively new computer the time you save this way is very minimal). on the other hand you have cleaner Dockerfile, which has less repeating lines (but I guess it is not THAT messy if you have a couple of duplicate lines). 🤔 at least I don't know of clear technical advantages of either one of the approaches (although it might exist and someone with more experience would tell us about it 😅 ) 

I guess, we can leave the final decision for you. you can keep whichever you prefer. 🙂

docker/django/Dockerfile
```diff
@@ -1,18 +1,45 @@
...
-RUN mkdir /code
+EXPOSE 8000
 
-WORKDIR /code
+# ==============================
+FROM base_stage AS development
+# ==============================
 
-COPY . /code
+RUN apt-install.sh gdal-bin build-essential
```
igordavydsson: don't forget to run `apt-cleanup.sh build-essential` during the same `RUN` layer to decrease the total image size 🙂

docker/django/Dockerfile
sam-hosseini: I changed my mind, I think you're right on the limited time saving opportunity versus the benefit of a DRY-er version of Dockerfile 👍 Done in 4489697

docker/django/Dockerfile
sam-hosseini: Done in 4489697

docker/django/Dockerfile
sam-hosseini: True. I even tried setting the `USER` at the top before `apt-install.sh` and it was failing with permission error :)

docker-compose.yml
sam-hosseini: Cool. I also updated the default network name to `helsinki` in d8146d71a7bee7cdcd1440e53456cdbfb25610a6 for uniformness among our newly Dockerized projects.

igordavydsson: that is a welcome addition 👍

PR #359, opened 2019-12-20
TH-112 - Kuva infra deployment

sam-hosseini:


sam-hosseini:
instead of gunicorn, uwsgi should have been used due to its widely used attribute in City of Helsinki, and done in https://github.com/City-of-Helsinki/linkedevents/pull/412

whitenoise is being integrated in https://github.com/City-of-Helsinki/linkedevents/pull/413

Closing this, as it is now obsolete.

Issue #366, opened 2020-01-20
Use autocomplete components to enable keyword set and event admin


Rikuoja:
Currently, dropdowns do not render, they are too big.

PR #369, opened 2020-01-21
Admin autocomplete fields

Doatti:
Resolves https://github.com/City-of-Helsinki/linkedevents/issues/366


Doatti:
The autocomplete list uses the model's `__str__()` method, so we could add the id there inside parentheses for example.

In DataSource and Language admins, `id` is editable. Shall I add `id` as a read-only field to all admins other than these two?

Rikuoja:
> In DataSource and Language admins, `id` is editable. Shall I add `id` as a read-only field to all admins other than these two?

Hm, I believe we should definitely prevent editing the ids for existing objects there, too. Otherwise we will get massive problems in the database. Is it possible to make it a read-only field *after* an object is created?

Rikuoja:
Also, we should not allow deleting any data sources. That would cause even worse problems than renaming a data source.

Doatti:
> > In DataSource and Language admins, `id` is editable. Shall I add `id` as a read-only field to all admins other than these two?
> 
> Hm, I believe we should definitely prevent editing the ids for existing objects there, too. Otherwise we will get massive problems in the database. Is it possible to make it a read-only field _after_ an object is created?

Yes, it's possible by overriding `get_readonly_fields()` on the admin class.

>Also, we should not allow deleting any data sources. That would cause even worse problems than renaming a data source.

Okay, I'll remove the "Delete" button from data source admin.

Doatti:
@Rikuoja I found that keyword sets can't be created from admin site, because the form doesn't have the `id` field and it isn't automatically generated. Should I add it as an editable field during creation or use the `AutoIdBaseAdmin` mixin?

Rikuoja:
> @Rikuoja I found that keyword sets can't be created from admin site, because the form doesn't have the `id` field and it isn't automatically generated. Should I add it as an editable field during creation or use the `AutoIdBaseAdmin` mixin?

Ah! I believe the ids are of the form data_source:id. I believe we should force the keyword sets to have the system data source (defined in settings.py), but the latter part should be editable on creation.

In `BaseModel`, the id is constructed using the separate origin_id field, so how about we allow specifying the origin_id here and then system data source is prepended to get an id of the right format?

Rikuoja:
> > @Rikuoja I found that keyword sets can't be created from admin site, because the form doesn't have the `id` field and it isn't automatically generated. Should I add it as an editable field during creation or use the `AutoIdBaseAdmin` mixin?
> 
> Ah! I believe the ids are of the form data_source:id. I believe we should force the keyword sets to have the system data source (defined in settings.py), but the latter part should be editable on creation.
> 
> In `BaseModel`, the id is constructed using the separate origin_id field, so how about we allow specifying the origin_id here and then system data source is prepended to get an id of the right format?

Ah! Looks like AutoIdBaseAdmin does this automatically :D

Doatti:
> Ah! Looks like AutoIdBaseAdmin does this automatically :D

Yes, the latter part is randomly generated, though.

events/admin.py
```diff
@@ -61,9 +61,10 @@ class KeywordAdmin(AutoIdBaseAdmin, TranslationAdmin, VersionAdmin):
 admin.site.register(Keyword, KeywordAdmin)
 
 
-class KeywordSetAdmin(BaseAdmin):
-    fields = ('data_source', 'origin_id', 'name', 'keywords', 'usage')
+class KeywordSetAdmin(AutoIdBaseAdmin):
+    fields = ('id', 'data_source', 'origin_id', 'name', 'keywords', 'usage')
     autocomplete_fields = ('keywords', )
+    readonly_fields = ('id',)
```
Rikuoja: I believe we need to do the same as with data sources and languages here: allow specifying the id on create, but not otherwise. In my understanding, if the user sets an id here, `AutoIdBaseAdmin` will use that to generate a `data_source:id` format id?

Doatti: We discussed this on Slack and decided to modify `AutoIdBaseAdmin` so it uses origin_id to create id of form `data_source:origin_id` if origin_id has been set.

events/admin.py
```diff
@@ -24,6 +24,8 @@ class AutoIdBaseAdmin(BaseAdmin):
     def save_model(self, request, obj, form, change):
         system_id = settings.SYSTEM_DATA_SOURCE_ID
         obj.data_source_id = system_id
+        if obj.origin_id:
+            obj.id = ':'.join([system_id, obj.origin_id])
```
Rikuoja: One more thing :D Here, what might happen (an edge case, I admit) is that we are saving an existing object. In that case, if id exists, we definitely should not change the id. Therefore, construct a new id *only* when id is missing, i.e. move these two lines inside the `if not obj.id` clause.

Rikuoja: Hm, actually best move *all* the lines inside the clause. We don't want to touch origin_id either if the object has been created earlier.

Doatti: Done.

Doatti:
Admin user can currently edit the "Origin id" of an existing keyword set, but it gets overwritten by `AutoIdBaseAdmin` to match the latter part of `id`. Maybe we should add `origin_id` to read-only fields when the object exists (i.e. not creating a new one)?

Rikuoja:
> Admin user can currently edit the "Origin id" of an existing keyword set, but it gets overwritten by `AutoIdBaseAdmin` to match the latter part of `id`. Maybe we should add `origin_id` to read-only fields when the object exists (i.e. not creating a new one)?

Yeah, `origin_id` should be a read-only field in any existing object. `data_source` too, but it should *never* be given by the user, even when creating, since system data source should be used.

events/admin.py
```diff
@@ -25,43 +25,54 @@ def save_model(self, request, obj, form, change):
...
-    fields = ('data_source', 'origin_id', 'name', 'keywords', 'usage')
+class KeywordSetAdmin(AutoIdBaseAdmin):
+    fields = ('id', 'data_source', 'origin_id', 'name', 'keywords', 'usage')
+    autocomplete_fields = ('keywords', )
+
+    def get_readonly_fields(self, request, obj=None):
+        if obj:
+            return ['id', 'origin_id', 'data_source']
+        else:
+            return ['id']
```
Rikuoja: Always add `data_source` here, since we don't want to allow specifying it. All new objects generated in the admin should come from `SYSTEM_DATA_SOURCE_ID`.

Doatti: Done.

PR #374, opened 2020-01-30
quick fix for non-yso keywords in the event description

dankiki:


dankiki:
I agree with @Rikuoja, misunderstood the code when drafting this PR.

Issue #385, opened 2020-02-12
Allow listing also deleted events (places, keywords) with the parameter `?show_deleted=True`


Rikuoja:
This way, a client could see if a resource (e.g. event) has been deleted since they last checked the API. This helps huge clients with local copies of data, such as IBM and MyHelsinki.

Combined with the `?last_modified_since=` filter, this would help such clients update *all* data that has been modified after their last visit, including recently deleted items, instead of having to go through the whole dataset.

Rikuoja:
To prevent actually using the data, we could even display a completely empty event, with only the `deleted=True` flag and the event id displayed in the json.

Doatti:
`show_deleted` and `show_deprecated` have already been implemented for places and keywords: https://github.com/City-of-Helsinki/linkedevents/blob/master/events/api.py#L951
https://github.com/City-of-Helsinki/linkedevents/blob/master/events/api.py#L766

There are no tests for these parameters, though.

Rikuoja:
Ah! Let's add tests then, so we remember the parameters exist :D

Doatti:
Should we include translations of `DELETED` for Finnish and Swedish, or is it okay to simply replace names in each language with the same string?

Rikuoja:
Yeah, let's add `POISTETTU` and `RADERAD`.

aleksisalonen:
If it's going to be showing up on user applications, I think we should provide the info in three languages.

This info is meant for human eyes after all.

FI: Tapahtuma poistettu
SV: Evengemang raderad (?)
EN: Event deleted

Rikuoja:
Well, it most likely is not shown in user applications, if the application coder knows what they are doing. It's mostly meant for the coder etc. to see that these items (with deleted=True) are no longer available, if they have specifically asked for those items to be listed.

aleksisalonen:
True. But if we're going natural language anyway, why stop at DELETED instead of going Event deleted?

I'm fine with either one.

Rikuoja:
Yeah, i'm just hoping the coder will visually notice this instantly. Tapahtuma poistettu looks like it doesn't stand out, even if the other fields are empty.

aleksisalonen:
I suppose DELETED has the most shock value :)

Doatti:
> Note that by default, the URI to the deleted event does not return this data. It returns HTTP 410 Gone as before.

[...]

PR #389, opened 2020-02-18
Show deleted parameter

Doatti:
Resolves https://github.com/City-of-Helsinki/linkedevents/issues/385


Doatti:
> Realized it would be best to check that all the other deleted event data is _not_ shown, since it may be erroneous or non-public. So we should check in the test that all the other fields are empty or non-existent, _only_ the expected fields are there.

The test does test for this. For every key in the response data, it asserts that it is included in `expected_keys`

Rikuoja:
> > Realized it would be best to check that all the other deleted event data is _not_ shown, since it may be erroneous or non-public. So we should check in the test that all the other fields are empty or non-existent, _only_ the expected fields are there.
> 
> The test does test for this. For every key in the response data, it asserts that it is included in `expected_keys`

Ah! My bad! Excellent :D

PR #394, opened 2020-02-26
Build productionish container using uwsgi and whitenoise

vikoivun:
Functionality-wise this still needs an opinionated way to run scheduled tasks. Instead of making everyone figure it out by themselves.

Also documentation.


vikoivun:
Maybe some updates still to README?

vikoivun:
Also still incoming more change for kulke importer

PR #411, opened 2020-05-18
Update EMAIL_BACKEND in local dev env

sam-hosseini:
When using Docker in local dev env, an admin user is created so that we can immediately login to django admin views after the server is run.

At the moment, `docker-compose up` crashes on the first run because it tries to send an email when the admin user is created. This PR fixes this, by checking if DEBUG is False to send the email after a user is created.

### DIFF
--- a/events/signals.py
+++ b/events/signals.py
@@ -1,5 +1,6 @@
 import logging
 
+from django.conf import settings
 from django.core.mail import send_mail
 from django.contrib.auth import get_user_model
 from django.contrib.sites.models import Site
@@ -22,7 +23,7 @@ def organization_post_save(sender, instance, created, **kwargs):
 
 
 def user_post_save(sender, instance, created, **kwargs):
-    if created:
+    if created and settings.DEBUG is False:
         User = get_user_model()
         recipient_list = [item[0] for item in User.objects.filter(is_superuser=True).values_list('email')]
         notification_type = NotificationType.USER_CREATED
--- a/events/signals.py
+++ b/events/signals.py
@@ -1,6 +1,5 @@
 import logging
 
-from django.conf import settings
 from django.core.mail import send_mail
 from django.contrib.auth import get_user_model
 from django.contrib.sites.models import Site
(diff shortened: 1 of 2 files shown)


sam-hosseini:
> Wouldn't the way to handle this be to catch the exception (if an email cannot be sent) and log the error?
Right, that's better. Reverted in ae137cd

This is the error stack:
```
django_1    | Traceback (most recent call last):
django_1    |   File "/app/events/signals.py", line 44, in user_post_save
django_1    |     html_message=rendered_notification['html_body']
django_1    |   File "/usr/local/lib/python3.6/site-packages/django/core/mail/__init__.py", line 60, in send_mail
django_1    |     return mail.send()
django_1    |   File "/usr/local/lib/python3.6/site-packages/django/core/mail/message.py", line 291, in send
django_1    |     return self.get_connection(fail_silently).send_messages([self])
django_1    |   File "/usr/local/lib/python3.6/site-packages/django/core/mail/backends/smtp.py", line 103, in send_messages
django_1    |     new_conn_created = self.open()
django_1    |   File "/usr/local/lib/python3.6/site-packages/django/core/mail/backends/smtp.py", line 63, in open
django_1    |     self.connection = self.connection_class(self.host, self.port, **connection_params)
django_1    |   File "/usr/local/lib/python3.6/smtplib.py", line 251, in __init__
django_1    |     (code, msg) = self.connect(host, port)
django_1    |   File "/usr/local/lib/python3.6/smtplib.py", line 336, in connect
django_1    |     self.sock = self._get_socket(host, port, self.timeout)
django_1    |   File "/usr/local/lib/python3.6/smtplib.py", line 307, in _get_socket
django_1    |     self.source_address)
django_1    |   File "/usr/local/lib/python3.6/socket.py", line 724, in create_connection
django_1    |     raise err
django_1    |   File "/usr/local/lib/python3.6/socket.py", line 713, in create_connection
django_1    |     sock.connect(sa)
django_1    | OSError: [Errno 99] Cannot assign requested address
```

I can fix the crash by catching the `OSError` exception or by setting the `EMAIL_BACKEND` during development to some value (for example [Console Backend](https://docs.djangoproject.com/en/2.2/topics/email/#console-backend)). Which one do you think is better?

Rikuoja:
Wow, didn't know there's a Console Backend, sounds like that is the right tool for the job!

PR #413, opened 2020-05-22
Whitenoise integration

sam-hosseini:
I tried to explain the configuration reasons in the commit messages.


vikoivun:
Have you tested if this still works when running in a "traditional" mode. Ie. with nginx serving out the /static folder?

sam-hosseini:
> Have you tested if this still works when running in a "traditional" mode. Ie. with nginx serving out the /static folder?

If I have understood whitenoise [usecase](http://whitenoise.evans.io/en/stable/#what-s-the-point-in-whitenoise-when-i-can-do-the-same-thing-in-a-few-lines-of-apache-nginx-config) correctly, there's no need for nginx to serve out the `/static` folder anymore with this PR.

* Without whitenoise

Depending on the incoming request, nginx decides whether to get the files from the filesystem or pass it on to uwsgi, like below:
image

* With whitenoise

Django can serve even static files and there's no need to configure nginx to serve `/static` folder. Both "static" and "dynamic" requests will be handled by uwsgi.

image

* To answer your question:

No I haven't tested because I didn't have access to a test environment. Could you please `checkout` this branch in an appropriate VM and see if it works @vikoivun ?

vikoivun:
I can, but it will take some time. Anyway, /static directory is an interface towards the runtime environment and compatibility must be maintained for some time.

sam-hosseini:
> I can, but it will take some time. Anyway, /static directory is an interface towards the runtime environment and compatibility must be maintained for some time.

Alrighty. I think compatibility shouldn't change at all because nginx can still remain configured to serve the `/static` folder and only pass the dynamic requests to uwsgi 👍🏼

And no problem, I'll work on https://helsinkisolutionoffice.atlassian.net/browse/HAP-67 in the meantime 👍🏼

sam-hosseini:
> I did a cursory test and this seems to work with the traditional setup with nginx serving the files. So no problems there.

Great, thanks for checking 👍🏼

PR #417, opened 2020-06-04
Add pytest watch

sam-hosseini:
At the moment, running the tests locally even for a single module creates a database and applies the migrations before proceeding to the test itself. This, on my machine, takes 20 seconds each time, which makes local testing rather slow. 5231b43 fixes that.

On top of that, while writing the tests, it saves time for the tests to keep (re-)running after making changes to code, so one wouldn't have to run the test command again and again. `pytest-watch` fixes that too.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,6 +1,6 @@
 [pytest]
 DJANGO_SETTINGS_MODULE=linkedevents.test_settings
-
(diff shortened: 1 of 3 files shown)


dankiki:
I would also think that commits could be squashed to keep the commit history cleaner.

PR #419, opened 2020-06-08
HAP-94 add audience min_age and max_age filters

dankiki:
The PR adds `audience_min_age_lt`, `audience_min_age_gt`, `audience_max_age_lt`, `audience_max_age_gt` filtering together with the required tests and documentation. It keeps the existing `audience_max_age` and `audience_min_age` filters for compatibility purposes, but marks them obsolete in the documentation.

### DIFF
--- a/helevents/templates/rest_framework/event_list.html
+++ b/helevents/templates/rest_framework/event_list.html
@@ -134,6 +134,19 @@ in <code>fi</code>, <code>sv</code>, <code>en</code>, <code>ru</code>,
 <pre><code>event/?language=ru
 </code></pre>
 <p><a href="?language=ru" title="json">See the result</a></p>
+<h3 id="audience-age">Event audience age boundaries.</h3>
+<p>To find events that are designed for specific age audiences use the query paramteres <code>audience_min_age_lt</code>,
+<code>audience_min_age_gt</code>, <code>audience_max_age_lt</code>, <code>audience_max_age_gt</code>.</p>
+
+<p> <code>audience_min_age_lt</code> returns the events whose minimal age is lower than or equals the specified value, <code>audience_min_age_gt</code> returns the events whose minimal age is greater than or equals the specified value. <code>max_age</code> parameteres, naturally, work the same way only for the maximum age of the event audience. Note, that the events that are not designed for the specific audiences will be omitted.</p>
+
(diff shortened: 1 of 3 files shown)


dankiki:
Guys, not that I want everyone to review, but getting someone's review would be great )

helevents/templates/rest_framework/event_list.html
```diff
@@ -134,6 +134,19 @@ <h3 id="event-language">Event language</h3>
 <pre><code>event/?language=ru
 </code></pre>
 <p><a href="?language=ru" title="json">See the result</a></p>
+<h3 id="audience-age">Event audience age boundaries.</h3>
```
Rikuoja, line 137: Here, the problem (and the reason why this has not been documented thus far) is that we actually have zero events which have the audience age data available in the production API at the moment. Therefore, this filter doesn't do anything in production.

Therefore, only when we have event data that contains audience age for some sensible subset of events, we can add documentation for this filter in production, *with* the caveat that most events *don't* contain this data at all. All the other documented filters actually work for all events.

Rikuoja: Okay, perhaps wording "to find events that are designed for specific age audiences" is enough 🤔 That explains to the API user why the filter doesn't do anything?

dankiki:
> If we are reordering all the imports in api.py (what is the new order based on?), then at least add comments that tells what the order is based on?

it was done by isort https://pypi.org/project/isort/ according to the following principles 

> isort parses specified files for global level import lines (imports outside of try / except blocks, >functions, etc..) and puts them all at the top of the file grouped together by the type of import:

>Future
>Python Standard Library
>Third Party
>Current Python Project
>Explicitly Local (. before import, as in: from . import x)
>Custom Separate Sections (Defined by forced_separate list in configuration file)
>Custom Sections (Defined by sections list in configuration file) 

if you feel this is totally out of place, I can reverse this. Just had isort hooked together with flake8 on precommit.

helevents/templates/rest_framework/event_list.html
dankiki: Not exactly 0... https://api.hel.fi/linkedevents/v1/event/?audience_min_age=16 kulke and harrastushaku support audience_min/max_age, this is needed for harrastuspassi. I will update the docs.

PR #425, opened 2020-08-20
add keyword similarity search

dankiki:
This PR adds trigram similarity based search for the keywords.


Rikuoja (approved):
Looks nice! Wonder if similar searches are needed or could be added to place and event endpoints, or what is the specific user story behind this feature?

PR #426, opened 2020-08-24
combine text and keyword searches

dankiki:
This PR adds the combined search for event content and keywords. The same keyword is searched for in the event and is used to retrieve the keywords which are then added to the event search. As this is just the first iteration, no documentation is added so far.


Rikuoja (approved):
Looks sensible!

PR #431, opened 2020-09-29
Added an importer for Mikkeli Nyt  -events

anttileppa:
Added support for importing events from Mikkeli Nyt

### DIFF
--- a/events/importer/mikkelinyt.py
+++ b/events/importer/mikkelinyt.py
@@ -118,12 +118,9 @@ class MikkeliNytImporter(Importer):
 
         _id = 'mikkelinyt:{}'.format(origin_id)
 
-        external_links = []
+        external_links = {}
         if registration:
-            external_links.append({
-                'name': 'rekisteröityminen',
-                'link': registration
-            })
+            external_links['rekisteröityminen'] = registration
 
         event = {
             'id': _id,
--- a/events/importer/mikkelinyt.py
+++ b/events/importer/mikkelinyt.py
@@ -1,0 +1,216 @@
+import os
+import re
+import requests
+import requests_cache
+import pytz
+import logging
+import hashlib
+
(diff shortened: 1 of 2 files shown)


dankiki:
Hi, Antti! I think that in order to add local importers for the non-Helsinki events and, in general, to use the linkedevents for other municipalities you'd need to fork the project.

Rikuoja:
It's not my decision anymore, but up to now we have accepted external code to the codebase if they are separate modules, so that anybody can use the code :) We don't want to fork the codebase if it is not necessary?

dankiki:
That might be possible as well =) I guess we'd need to discuss this a bit more internally =)

Rikuoja (approved):
This looks sensible to me, but will not merge as I am not the one who will be responsible for the codebase in the future :) In general, I would recommend accepting separate modules that do no changes to the base codebase, to prevent the codebases of different cities diverging if it can be prevented.

dankiki:
I don't mind adding these modules as well. Only please have some tests showing the structure of the imported data, that would make the code more editable in the future.

dankiki (changes requested):
very sensible, but would be great to see the structure of the imported data. in a test.

anttileppa:
@dankiki added basic tests for importing and removing events, is it ok now or should i add more tests?

dankiki (approved):
looks good!

dankiki:
Hi! @anttileppa I was thinking you'd merge them yourself, just in case you realize the last minute that you want to add something, but I could of course do that for you as well.
would you like to join a common slack channel for possible discussions?

anttileppa:
@dankiki I would really much like to join the slack channel but I don't think i have permission to merge the PR (it says "Only those with write access to this repository can merge pull requests.")

PR #438, opened 2020-12-16
Add Language translations to Django admin

jpaakko:
**Added**

- Made it possible to view and edit `Language` object translations in the Django admin

**Fixed**

- The `Language` rows not being clickable in the Django admin. Currently, when no languages have been created and an importer is run, a set of languages are dynamically created based on `settings.LANGUAGES` and the languages supported by the importer. However, the current implementation sets only the `id` field for the new `Language` and not the `name` field which is shown in the Django admin list view. Thus, the created languages can't be clicked in the Django admin list view and the only way to get to the edit view is by a direct URL. This change fixes the issue by setting the `name` field for the created `Language` so that the language can be clicked in the Django admin.


charn:
Cherry-picked the relevant fix commit and pushed it to `develop`. :shipit:

PR #443, opened 2021-02-10
Feature/improve search performance

dankiki:
This PR adds 
- regex & memcached powered query parameters for the quick search of the local and internet based events. The query parameters in question share the _ongoing_ word in their naming.  
Example search:` https://api.hel.fi/linkedevents/v1/event/?local_ongoing_OR=teatteri,lapsi`
- grouping for _ongoing_ and keyword searches to support (x OR y) AND (a OR b) search logic
Example search: `https://api.hel.fi/linkedevents/v1/event/?local_ongoing_OR=musiikki&keyword_OR_set1=yso:p11185,yso:p1808&keyword_OR_set2=yso:p18434,yso:p26626`

This code is already running for a while but has never passed through peer review, so wasn't merged to master yet.

Second search that is added in this PR is based on the Postgres full-text search functionality and hence is language specific. Separate tsvector columns are created in the db for Finnish, Swedish, and English. They are populated and kept up-to-date on the db side, see migration 0080.
urls.py has a small correction of the imports, not related to the main topic of the commit.


events/management/commands/populate_local_event_cache.py
```diff
@@ -30,6 +32,26 @@ def handle(self, *args, **options):
...
+                                                         'location__street_address_fi', 'location__street_address_sv',
+                                                         'location__name_fi', 'location__name_sv', 'location__name_en',  # noqa E501
+                                                         'location__description_fi', 'location__description_sv',
+                                                         'location__description_en')
+        event_dict = {i[0]: set() for i in inet_events}
+        for i in inet_events:
+            event_dict[i[0]].update(i[1:])
+            event_dict[i[0]].discard(None)
+
+        event_strings = {k: " ".join(v).replace('\n', ' ').replace('\r', ' ') for k, v in event_dict.items()}
```
nikomakela, line 55: This event_strings creation could be in a reusable function, because the pattern is always the same.

events/migrations/0080_populate_tsvectors_and_indices.py
```diff
@@ -0,0 +1,86 @@
...
+                ),
+                migrations.RunSQL(sql="CREATE INDEX CONCURRENTLY events_swedish_content_index ON events_event USING GIN (search_vector_sv);",
+                                  reverse_sql="DROP INDEX events_swedish_content_index;"
+                ),
+                migrations.RunSQL(sql="CREATE INDEX CONCURRENTLY events_english_content_index ON events_event USING GIN (search_vector_en);",
+                                  reverse_sql="DROP INDEX events_english_content_index;"
+                )
+            ]
+        )
+    ]
```
nikomakela, line 86: Is there some tests for this? At least it should be really carefully tested manually, but also unit test could be written. For me as an outsider this seems overwhelming and super hard to understand what happens here. So hard actually, that I won't even try :)

events/models.py
```diff
@@ -688,6 +705,9 @@ def save(self, *args, **kwargs):
                                                  str(self.audience.filter(deprecated=True).values('id')) +
                                                  ". Please use up-to-date keywords.")})
 
+        # if self.location__divisions__ocd_id__endswith == MUNIGEO_MUNI:
+        #     self.local = True
+
```
nikomakela, line 710: Remove the commented code that is not needed.

nikomakela (commented):
Sorry @petterill and @dankiki, I noticed only today that you had requested review for this PR.

Repetitive code should be replaced with functions, some documentation about used caches should be added, application commands should be tested and some query managers could be used. The PR is huge and hard to understand for someone who is not working with LE everyday. Thats the reason why unit tests are really important. Otherwise I cannot be sure whether all this does what is requested.

Also it seems like parts of this code is removed in later pull requests.

events/api.py
```diff
@@ -1748,32 +1737,154 @@ def parse_duration_string(duration):
...
+        all_sets = rc.findall(''.join(params))
+        for i in all_sets:
+            val = params.get(i, None)
+            if val:
+                val = val.split(',')
+                queryset = queryset.filter(keywords__pk__in=val)
+
+    val = params.get('internet_based', None)
+    if val and validate_bool(val, 'internet_based'):
+        queryset = queryset.filter(location__id__contains='internet')
```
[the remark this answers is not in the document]
nikomakela: > A similar mapping and reusable function that I requested earlier could be used here as well.
https://github.com/City-of-Helsinki/linkedevents/pull/443#discussion_r628078216

events/api.py
```diff
@@ -1748,32 +1737,154 @@ def parse_duration_string(duration):
...
+        cached_ids = {k: v for i in cache.get_many(['internet_ids', 'local_ids']).values() for k, v in i.items()}
+        ids = {k for k, v in cached_ids.items() if rc.search(v, concurrent=True)}
+        queryset = queryset.filter(id__in=ids)
+
+    val = params.get('all_ongoing_OR', None)
+    if val:
+        rc = _terms_to_regex(val, 'OR')
+        cached_ids = {k: v for i in cache.get_many(['internet_ids', 'local_ids']).values() for k, v in i.items()}
+        ids = {k for k, v in cached_ids.items() if rc.search(v, concurrent=True)}
+        queryset = queryset.filter(id__in=ids)
```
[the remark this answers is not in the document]
nikomakela: I'm not quite sure if the function can be selected and called like that, but the idea could be get from here :D

events/tests/test_event_get.py
```diff
@@ -979,3 +983,36 @@ def test_keyword_and_text(api_client, event, event2, keyword):
     event.save()
     response = get_list(api_client, query_string='combined_text=lapset,aikuiset')
     assert_events_in_response([event], response)
+
```
quyenlq, line 986: Some tests are still missing and I feel it's quite important to have. For example: `local_ongoing_text`, `local_ongoing_OR`, `local_onging_AND` ... Would be great if the all available filter params are tested properly

events/api.py
```diff
@@ -369,18 +370,6 @@ def __init__(self, *args, **kwargs):
                     del self.fields[key]
             del self.fields[field_name]
 
-    # def get_field(self, model_field):
```
quyenlq, line 372: Unused code should be removed :)

PR #454, opened 2021-04-21
LINK-264 Change the logic of the KeywordMatcher

dankiki:
KeywordMatcher takes as input the strings of words with or without language specification and transforms them into a list of Keywords. Postgres full-text search is used to find a label matched by lexeme. The results are ranked with TrigramSimilarity as full-text SearchRank is not suitable for ranking individual words. If no language is passed we cycle through all options as specified in FULLTEXT_SEARCH_LANGUAGES and select the best match according to similarity. If no match is found the string is checked for the possibility that it could be split and search for matches is repeated.
Syntactic dust - punctuation marks etc. - is ignored.
If the language is specified the search is conducted only within the language specific search vectors.


nikomakela (approved):
Looks good to me.

PR #455, opened 2021-04-28
Feature/event type

dankiki:
LINK-181 This PR adds the type_id field with three possible values: General, Courses, and Volunteering,


events/keywords.py
```diff
@@ -1,112 +1,78 @@
...
-        print('Initialized', len(self.labels), 'keyword keys')
+        pass
 
-    def match(self, text):
-        if self.skip:
-            return None
-        wordsplit = re.compile(r'\s+')
-        # labels = KeywordLabel.objects
-        # match = labels.filter(name__iexact=text)
+    def full_text_matching(self, text, language=None):
```
nikomakela, line 12: This method has good comments, but it definitely needs some unit tests. It would not only to make it even clearer after comments, but would also be a proof that it really works.

events/keywords.py
```diff
@@ -1,112 +1,78 @@
...
+        else:
             return None
 
-        keyword_ids = set()
-        if match_type not in ['exact', 'subword']:
-            cmatch = get_close_matches(
-                text, [m.lower() for m in matches], n=1)
-            if len(cmatch) == 1:
-                keyword_ids = self.name_to_keyword_ids.get(cmatch[0])
+    def label_match(self, text, language=None):
```
nikomakela, line 42: Good comments, like in full_text_matching -method, but needs unit tests also.

events/keywords.py
```diff
@@ -1,112 +1,78 @@
...
-                aggregate_name = re.sub(ENDING_PARENTHESIS_PATTERN, '', aggregate_keyword.name_fi)
-                result = [aggregate_keyword]
-                for o in objects.exclude(name_fi__istartswith=aggregate_name):
-                    result.append(o)
-                return result
-            except Keyword.DoesNotExist:
-                pass
-            return objects
-        return objects
+    def match(self, text, language=None):
```
nikomakela, line 70: If any, this needs unit tests, since label_match is a self created unknown and untested method.

nikomakela (commented):
I would like to request some unit tests to clarify what the code does and to proof it works. Some questions asked about values syncing.

dankiki:
Closing and creating a new one comparing to another branch.

PR #458, opened 2021-05-10
LINK-383 Add endpoints for emailing feedback form

dankiki:
This PR adds two endpoints that allow creating a feedback for signed in and guest users. In both cases the content of the form is recorded in the Feedback table and an email is sent to the address specified in settings as a support email.


nikomakela (approved):
Looks good to me. I have nothing to add or improve.

PR #459, opened 2021-05-19
Feature/event type

dankiki:
LINK-181 This PR adds the type_id field with three possible values: General, Courses, and Volunteering

Queyr parameter added: `event_type`.


events/models.py
```diff
@@ -647,7 +661,20 @@ class SuperEventType:
 
     replaced_by = models.ForeignKey('Event', on_delete=models.SET_NULL, related_name='aliases', null=True, blank=True)
 
+    maximum_attendee_capacity = models.PositiveIntegerField(
+        verbose_name=_('maximum attendee capacity'), null=True, blank=True
+    )
+    remaining_attendee_capacity = models.PositiveIntegerField(
```
dankiki, line 667: The remaining attendee capacity was left unchanged for two reasons:
- first the event organizer should be able to change it if the registration is maintained elsewhere.
- registration will most probably be the topic of upcoming sprint and then this will be discussed and possibly changed.

events/api.py
```diff
@@ -1938,6 +1938,18 @@ def _filter_event_queryset(queryset, params, srs=None):
...
+    if val:
+        vals = val.lower().split(',')
+        event_types = {k[1].lower(): k[0] for k in Event.TYPE_IDS}
+        search_vals = []
+        for v in vals:
+            if v not in event_types:
+                raise ParseError(_(f'Event type can be of the following values:{" ".join(event_types.keys())}'))
+            search_vals.append(event_types[v])
+
+        queryset = queryset.filter(type_id__in=search_vals)
```
nikomakela, line 1951: I would create a function of this - 1 function per parameter-feature -handling. If the handlers are repetitive, then make it generic enough to serve all the kind.

events/models.py
nikomakela: Yes, I remember that I reviewed this in some pull request. Should the feature be left out if it's already known that it will be discussed in near future? I was just thinking that if any data is really collected to this field, a painful data migration is ahead. Is the event organizer going to keep this updated manually or how is the field really maintained? How the overbooking is taken care of? This `remaining_attendee_capacity` really seems something that should be automatically calculated with a signal or on post save process. Are the actual enrolments stored in LinkedEvents or are they stored in the client system? If they are in LE, it would be easy to auto calculate them - If they are in client system, it's as easy to them to update the taken seats than amount of seats left.

events/models.py
```diff
@@ -647,7 +661,20 @@ class SuperEventType:
...
+        verbose_name=_('remaining attendee capacity'), null=True, blank=True
+    )
+
+    # TODO: make into agreement with schema.org
     # Custom fields not from schema.org
+    minimum_attendee_capacity = models.PositiveIntegerField(
```
nikomakela, line 673: PositiveSmallIntegerField could be used instead: https://docs.djangoproject.com/en/3.2/ref/models/fields/#positivesmallintegerfield

events/models.py
nikomakela, line 667: PositiveSmallIntegerField could be used instead: https://docs.djangoproject.com/en/3.2/ref/models/fields/#positivesmallintegerfield

events/utils.py
```diff
@@ -32,8 +32,9 @@ def get_value_from_tuple_list(list_of_tuples, search_key, value_index):
                         returned and which is used as a key
     :return: Value from either side of tuple
     """
+
     for i, v in enumerate(list_of_tuples):
-        if v[value_index ^ 1] == search_key:
+        if str(v[value_index ^ 1]) == str(search_key):
```
nikomakela, line 37: I don't get the idea of this `^1`. What it does? 🤔

nikomakela (commented):
Some minor notes added. There was 1 test written, but some unit tests for different parameter handlings should still be added. New model fields especially, didn't have any.

events/models.py
dankiki: agreed

events/api.py
dankiki: that's the style of the whole file, then I would need to refactor pretty much all of it.

events/models.py
dankiki: True, positive small integer suit better. agreed.

events/utils.py
dankiki: this is a bitwise comparison. This function takes in a list of tuples and then treats it as a dictionary, with either the first or the second entry of the tuple being the key, and the other, respectively, the value.
Let's call this whole function g(list_of_tuples, search_key, index) and let's say we have a list of tuples  l = [('a',1),('b',2),('c',3)], then 
g(l, 'a', 0) is None
g(l, 'a', 1) == 1
g(l, 1, 0) == 'a'
g(l, 1, 1) is None

PR #460, opened 2021-05-26
LINK-490 Set "General" as default type for event list

dankiki:
When event_type query parameter is not specified only events of type "General" are returned.

### DIFF
--- a/events/api.py
+++ b/events/api.py
@@ -1951,7 +1951,7 @@ def _filter_event_queryset(queryset, params, srs=None):
 
         queryset = queryset.filter(type_id__in=search_vals)
     else:
-        queryset = queryset.filter(type_id=1)
+        queryset = queryset.filter(type_id=Event.Type_Id.GENERAL)
 
     val = params.get('last_modified_since', None)
     # This should be in format which dateutil.parser recognizes, e.g.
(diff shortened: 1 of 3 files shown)


events/tests/test_event_get.py
```diff
@@ -94,14 +95,22 @@ def assert_event_fields_exist(data, version='v1'):
     assert_fields_exist(data, fields)
 
 
-def assert_events_in_response(events, response):
+def assert_events_in_response(events, response, query=''):
     response_event_ids = {event['id'] for event in response.data['data']}
     expected_event_ids = {event.id for event in events}
-    assert response_event_ids == expected_event_ids
+    if query:
+        assert response_event_ids == expected_event_ids, f'\nquery: {query}'
```
dankiki, line 102: the point of this change is to see the query in the pytest output, should assertion fail

events/api.py
```diff
@@ -1950,6 +1950,8 @@ def _filter_event_queryset(queryset, params, srs=None):
             search_vals.append(event_types[v])
 
         queryset = queryset.filter(type_id__in=search_vals)
+    else:
+        queryset = queryset.filter(type_id=1)
```
nikomakela, line 1954: Instead of hardcoded type_id value, a reference to `Event.Type_Id.GENERAL` should be used: `Event.objects.filter(type_id=Event.Type_Id.GENERAL)`, which makes this 

```
    else:
        queryset = queryset.filter(type_id=Event.Type_Id.GENERAL)

nikomakela (approved):
A minor change requested for future code maintenance sake.

events/api.py
dankiki: absolutely!

PR #469, opened 2021-08-19
LINK-542 Remove feedback form body limit

dankiki:
Feedback form should not limit the size of the feedback a user could leave.


nikomakela (approved):
ok!

PR #472, opened 2021-09-22
LINK-485. Fix failing PUT requests

dankiki:
Querying *.objects.all() for related fields results in huge requests whenever an Event is updated. Internal DRF mechanics for related fields as specified in relations.py uses get() for every updated related field and presumes an existing queryset. The goal of this bug fix is to provide querysets for the related fields that would contain only the related objects mentioned in the request

### DIFF
--- a/events/api.py
+++ b/events/api.py
@@ -1,6 +1,5 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
-
 import base64
 import re
 import struct
@@ -300,6 +299,18 @@ class JSONLDRelatedField(relations.HyperlinkedRelatedField):
 
     def is_expanded(self):
         return getattr(self, 'expanded', False)
+
+    def get_queryset(self):
+        #  For certain related fields we preload the queryset to avoid *.objects.all() query which can easily overload
+        #  the memory as database grows.
+        if isinstance(self._kwargs['serializer'], str):
+            return super(JSONLDRelatedField, self).get_queryset()
+        current_model = self._kwargs['serializer'].Meta.model
+        preloaded_fields = {Place: 'location', Keyword: 'keywords'}
+        if current_model in preloaded_fields.keys():
+            return self.context.get(preloaded_fields[current_model])
+        else:
+            return super(JSONLDRelatedField, self).get_queryset()
 
 
 class EnumChoiceField(serializers.Field):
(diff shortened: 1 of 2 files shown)


events/api.py
```diff
@@ -2452,8 +2465,33 @@ def filter_queryset(self, queryset):
...
+        if isinstance(serializer.validated_data, list):
+            event_data_list = serializer.validated_data
+        else:
+            event_data_list = [serializer.validated_data]
+        for event_data in event_data_list:
+            org = self.organization
+            if hasattr(event_data, 'publisher'):
+                org = event_data['publisher']
+            if not self.request.user.can_edit_event(org, event_data['publication_status']):
+                raise DRFPermissionDenied()
```
nikomakela, line 2491: This could have been in different commit so 1 commit contains changes related to 1 subject only.

events/api.py
```diff
@@ -2484,15 +2517,76 @@ def perform_update(self, serializer):
             if not self.request.user.can_edit_event(org, event_data['publication_status']):
                 raise DRFPermissionDenied()
 
+        if isinstance(serializer, EventSerializer) and not self.request.user.can_edit_event(
+                serializer.instance.publisher,
+                serializer.instance.publication_status,
+        ):
+            raise DRFPermissionDenied()
```
nikomakela, line 2524: permission changes could have been in another commit

events/api.py
```diff
@@ -2484,15 +2517,76 @@ def perform_update(self, serializer):
...
+        serializer = EventSerializer(self.filter_queryset(self.get_queryset()),
+                                     data=request.data,
+                                     context=context,
+                                     many=bulk,
+                                     partial=partial,)
+        if not serializer.instance:
+            raise serializers.ValidationError('Missing matching events to update.')
+        serializer.is_valid(raise_exception=True)
+        self.perform_update(serializer)
+        return Response(serializer.data, status=status.HTTP_200_OK)
```
nikomakela, line 2541: Fixing bulk_update is a 3rd subject

events/api.py
```diff
@@ -2484,15 +2517,76 @@ def perform_update(self, serializer):
...
+        location_ids = []
+        for event in events:
+            keyword_ids.extend(retrieve_ids('keywords', event))
+            keyword_ids.extend(retrieve_ids('audience', event))
+            location_ids.extend(retrieve_ids('location', event))
+        if location_ids:
+            locations = Place.objects.filter(id__in=location_ids)
+        if keyword_ids:
+            keywords = Keyword.objects.filter(id__in=keyword_ids)
+        return keywords, locations, bulk
```
nikomakela, line 2589: Could the retrieve ids be a more common util? It seems like it could be handy in some other parts of the code also. Changes in cache related fields seems to 4th subject in same commit.

events/models.py
```diff
@@ -928,7 +928,7 @@ class Feedback(models.Model):
     name = models.CharField(verbose_name=_('Name'), max_length=255, blank=True)
     email = models.EmailField(verbose_name=_('E-mail'))
     subject = models.CharField(verbose_name=_('Subject'), max_length=255, blank=True)
-    body = models.TextField(verbose_name=_('Body'), blank=True)
+    body = models.TextField(verbose_name=_('Body'), max_length=10000, blank=True)
```
nikomakela, line 931: This is 5th subject. Should this have a migration file?

nikomakela (commented):
Multiple different subjects in 1 single commit makes the PR a bit hard to review. Doesn't the changes need any updates to unit tests? Should the textfield max length change have a migration file?

events/models.py
dankiki: hei! thanks for spotting this one out. it got here by mistake indeed. this change has already been merged a while ago.

dankiki:
> Multiple different subjects in 1 single commit makes the PR a bit hard to review. Doesn't the changes need any updates to unit tests? Should the textfield max length change have a migration file?

textfield indeed got here by mistake, all the others. I'm afraid, belong together to a single commit. one without the others would be a breaking change.

PR #473, opened 2021-09-30
LINK-533 Fix bulk event update for non general event type

dankiki:
During update the events are passed through `EventViewset.filter_queryset()` for authorization check up, this method invokes  `_filter_queryset`. The latter is used to filter events by params in GET request and should not be used for other request methods.

### DIFF
--- a/events/tests/test_event_put.py
+++ b/events/tests/test_event_put.py
@@ -197,7 +197,7 @@ def test__update_minimal_event_with_autopopulated_fields_with_put(api_client, mi
 
 
 @pytest.mark.django_db
-def test__update_an_event_with_put(api_client, complex_event_dict, user):
+def test__update_an_event_complex_dict(api_client, complex_event_dict, user):
 
     # create an event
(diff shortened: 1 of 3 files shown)


events/tests/test_event_put.py
```diff
@@ -234,6 +235,47 @@ def test__update_an_event_with_put(api_client, complex_event_dict, user):
...
+    response2 = update_with_put(api_client, event_id, data2)
+    assert str(response2.data['detail']) == 'Invalid value "Non existing event type"'
+
+
+@pytest.mark.django_db
+def test__bulk_update_single_event(api_client, complex_event_dict, user):
+    api_client.force_authenticate(user=user)
+    response = create_with_post(api_client, complex_event_dict)
+
+    data2 = response.data
```
nikomakela, line 267: trivial, but data2 and data3 variable names are bit confusing

nikomakela (approved):
Good!

PR #505, opened 2023-01-24
Fix YSO importer

voneiden:
Remove deprecated/unused bulk_mode and enforce importing only supported languages to avoid importing Northern Sami (language code "se") which was introduced to YSO sometime around 2022.


charn (approved):
AFAIK LGTM! :+1: :shipit:

PR #508, opened 2023-01-25
LINK-1123 | Reformat code

charn:
This PR reformats the code with `black`, `isort` and `pre-commit` hooks. Tooling is based on the #507. `flake8` based errors are ignored in this PR since they have to be fixed by hand. So logic is not changed, only formatting.

Also adds `.git-blame-ignore-revs` and documents it in `README` for ignoring big formatting related commits.


danipran (approved):
Should be fine, tests pass (at least the ones that work), LGTM 👍

PR #509, opened 2023-01-26
Add Github Actions CI/CD workflow

voneiden:
The workflow setups the project and runs checks for migrations, black, flake8, isort, coverage and runs the tests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,7 +76,6 @@ jobs:
           black --check .
 
       - name: Check flake8
-        continue-on-error: true # TODO remove once flake8 issues are sorted out
         uses: liskin/gh-problem-matcher-wrap@v2
         with:
           linters: flake8
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,101 @@
+name: linkedevents-ci
+
+on:
+  push:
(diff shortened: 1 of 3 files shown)


voneiden:
Oletan että Azuren putki hajoaa koska PR tulee mun forkista. Ei varmaan isompaa merkitystä.

voneiden:
Also note some steps have continue on error set (migration check, flake8, pytest) as they still need to be addressed in other issues to pass.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,105 @@
...
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: psycopg2 prerequisites
+        run: sudo apt-get install libpq-dev
+
+      - name: System requirements for Django
+        run: sudo apt-get install gdal-bin gettext postgresql-client
+
+      - name: Setup postgres extensions
```
charn, line 47: I wonder if we really this step. :thinking: There are Django migrations in place which at least in theory should load `hstore` and `pg_trgm` since we're not skipping migrations (running `pytest` with `--no-migrations`).

voneiden: Oh, looks like it works fine without.  Whoops.

charn (approved):
LGTM! :+1:

PR #510, opened 2023-01-26
Fix elasticsearch dep

voneiden:
django-haystack==3.2.1 supports ES 1, 2, 5 ja 7 versions. Currently the unconstrainted elasticsearch dependency picks v8 package.

Without restricting to v7, `python manage.py event_import tprek --places` breaks with `TypeError: bulk() got an unexpected keyword argument 'doc_type'`

### DIFF
--- a/README.md
+++ b/README.md
@@ -235,7 +235,9 @@ Linkedevents uses Elasticsearch for generating results on the /search-endpoint. 
 
 1. Install elasticsearch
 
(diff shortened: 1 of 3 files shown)


danipran (approved):
LGTM 👍 Would've appreciated a reference to the version support in PR description (since you most likely got it from somewhere instead of just doing very good guesswork). Found it from the documentation: https://django-haystack.readthedocs.io/en/v3.2.1/installing_search_engines.html#elasticsearch

PR #511, opened 2023-01-27
LINK-1123 | Handle flake8 errors

charn:
- Fix simple errors
- Ignore too complex functions from QA
- Raise allowed level of complexity
- Remove duplicated test code
- Replace custom `active_language` function with `override` from Django

Refactoring `super` calls to have no arguments wasn't part of `flake8` errors. It however highlights the `super` calls where the immediate superclass is not called.

Refs LINK-1123

### DIFF
--- a/events/api.py
+++ b/events/api.py
@@ -12,6 +12,7 @@ from datetime import time as datetime_time
 from datetime import timedelta
 from functools import partial, reduce
 from operator import or_
+from typing import Iterable, Optional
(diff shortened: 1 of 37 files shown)


danipran (approved):
The functions with `# noqa: C901` are gonna come back and haunt us, aren't they? 😓 Anyway, within the scope of the PR,  LGTM, good calls on handling the errors 👍

PR #513, opened 2023-01-30
Fix deprecated use of imp

voneiden:
Remove deprecated use of `imp` module as described @ https://stackoverflow.com/questions/32175693/python-importlibs-analogue-for-imp-new-module

`imp` was deprecated in 3.4 and will be removed in 3.12 ( https://docs.python.org/3/library/imp.html )

### DIFF
--- a/linkedevents/settings.py
+++ b/linkedevents/settings.py
@@ -440,11 +440,11 @@ INSTANCE_NAME = env("INSTANCE_NAME")
 # like database and email that differ between development and production.
 f = os.path.join(BASE_DIR, "local_settings.py")
 if os.path.exists(f):
-    import imp
     import sys
+    import types
 
     module_name = "%s.local_settings" % ROOT_URLCONF.split(".")[0]
-    module = imp.new_module(module_name)
+    module = types.ModuleType(module_name)
     module.__file__ = f
     sys.modules[module_name] = module
     exec(open(f, "rb").read())


linkedevents/settings.py
```diff
@@ -440,11 +440,11 @@ def dummy_haystack_connection_for_lang(language_code):
 # like database and email that differ between development and production.
 f = os.path.join(BASE_DIR, "local_settings.py")
 if os.path.exists(f):
-    import imp
     import sys
+    import types
 
     module_name = "%s.local_settings" % ROOT_URLCONF.split(".")[0]
-    module = imp.new_module(module_name)
+    module = types.ModuleType(module_name)
```
charn, line 447: The documentation suggests using `importlib.util.module_from_spec` :thinking:

danipran (approved):
Works, and replacing deprecated stuff is good, but would consider/prefer alternative solutions, e.g. just try importing the file instead of creating your very own module.

linkedevents/settings.py
voneiden: It appears to be preferred. ModuleType documentation suggests 

> Use [importlib.util.module_from_spec()](https://docs.python.org/3/library/importlib.html#importlib.util.module_from_spec) to create a new module if you wish to set the various import-controlled attributes.

https://github.com/python/cpython/blob/main/Lib/imp.py#L49

` types.ModuleType(module_name)` appears to be identical to what the `imp.new_module` does

PR #515, opened 2023-01-31
LINK-1138 | Dockerized collectstatic and template installation

dankiki:
This PR adds collectstatic and templates installation to the docker-entrypoint 

- Update `.env.example` to have good defaults for development
- Add new `staticbuilder` and `production` build stages to the `Dockerfile`. Production container will now have static files collected out of the box.
- Add a new `INSTALL_TEMPLATE` env variable for installing a template/theme e.g. `helevents`
- Development requirements will no longer be available in the production container
- Fix broken django-debug-toolbar and make it work with docker
- Bump django, ipython, django-debug-toolbar, werkzeug


docker/django/docker-entrypoint.sh
```diff
@@ -20,12 +20,20 @@ if [[ "$CREATE_SUPERUSER" = "true" ]]; then
     ./manage.py create_admin_superuser
 fi
 
+if [[ "$COLLECT_STATIC" = "true" ]]; then
```
charn, line 23: Another common solution for doing this would be to have a production docker image (additional build stage in the `Dockerfile`) where templates would be installed as part of the build process.

charn: Exmples:
- https://github.com/City-of-Helsinki/maritime-maas/blob/main/Dockerfile
- https://github.com/City-of-Helsinki/kukkuu/blob/master/Dockerfile

charn: Running `collectstatic` in the `Dockerfile` will make the static files part of the image. Doing it in the entrypoint will of course allow the image be more dynamic, but it will add more to the startup time for the container. Not much, but some still.

Is there a situation where we would not like to run `collectstatic` in a production environment? In your local dev environment (development image), at least usually, you don't want/need `collectstatic` to happen.

`install_templates` looks like something we might not want to have in every situation, so maybe having at least that in the entrypoint script is good.

charn: That said, the current solution also works and we can refactor it later if needed, your call. :+1: :shipit:

charn (approved):
:shipit:

docker/django/docker-entrypoint.sh
tuomas777: I would remove the COLLECT_STATIC variable and bind static collecting to the DEBUG variable. Needing to run static collecting when debug is true should be very rare, you can still do it pretty easily manually. Having fewer settings and more consistency between the projects in Platta will be beneficial in the future at least for our team 🙂

docker/django/docker-entrypoint.sh
```diff
@@ -20,12 +20,20 @@ if [[ "$CREATE_SUPERUSER" = "true" ]]; then
...
+    ./manage.py collectstatic
+fi
+
+if [[ "$INSTALL_TEMPLATES" = "true" ]]; then
+    ./manage.py install_templates helevents
```
tuomas777, line 28: As other templates are supported, I think it would be nicer if this allowed those to be installed also, not just Helsinki ones. Maybe the installed templates' name could come from the variable?

tuomas777 (commented):
[Openshift runs containers with an arbitrary user](https://helsinkisolutionoffice.atlassian.net/wiki/spaces/HELFI/pages/7725613137/Running+containers+in+Openshift#OpenShift-specialities), which I **think** will cause permission problems with both of these commands when they are run in Platta. Can you test this also in OpenShift?

docker/django/Dockerfile
```diff
@@ -2,34 +2,60 @@ ARG PYTHON_VERSION=3.9
...
-ENTRYPOINT ["/app/docker/django/docker-entrypoint.sh"]
+USER appuser
+
+EXPOSE 8000/tcp
+
+# ==============================
+FROM appbase AS staticbuilder
+# ==============================
+
+ENV VAR_ROOT /app
```
danipran, line 48: Same as above, use `<key>=<value>` syntax.

danipran (changes requested):

docker/django/Dockerfile
```diff
@@ -2,34 +2,60 @@ ARG PYTHON_VERSION=3.9
...
 # ==============================
-FROM base_stage AS development
+FROM appbase AS development
 # ==============================
 
 COPY --chown=appuser:appuser requirements-dev.txt /app/
+RUN pip install --no-cache-dir -r /app/requirements-dev.txt
 
-RUN pip --no-cache-dir install -r /app/requirements-dev.txt
+ENV DEV_SERVER=1
```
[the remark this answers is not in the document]
danipran: Alternatively, switch the `true`s to 1s. I guess that might make even more sense. Maybe. Either way, consistency!

docker/django/Dockerfile
```diff
@@ -2,34 +2,60 @@ ARG PYTHON_VERSION=3.9
...
-FROM helsinkitest/python:${PYTHON_VERSION}-${IMAGE_VARIANT} AS base_stage
+FROM helsinkitest/python:${PYTHON_VERSION}-${IMAGE_VARIANT} AS appbase
 # ==============================
+RUN mkdir /entrypoint
 
-
-ENV PYTHONDONTWRITEBYTECODE true
-ENV PYTHONUNBUFFERED true
+ENV PYTHONDONTWRITEBYTECODE 1
+ENV PYTHONUNBUFFERED 1
```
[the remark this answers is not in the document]
charn: Good point and catch! :+1:

docker/django/Dockerfile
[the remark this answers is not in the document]
charn: I just personally favor numbers over strings. But maybe I'll give in and just put `true` here. :sob:

danipran (dismissed):
LGTM :+1:

Edit: nvm, found something

docker/django/Dockerfile
```diff
@@ -2,34 +2,60 @@ ARG PYTHON_VERSION=3.9
...
-FROM helsinkitest/python:${PYTHON_VERSION}-${IMAGE_VARIANT} AS base_stage
+FROM helsinkitest/python:${PYTHON_VERSION}-${IMAGE_VARIANT} AS appbase
 # ==============================
+RUN mkdir /entrypoint
 
-
-ENV PYTHONDONTWRITEBYTECODE true
-ENV PYTHONUNBUFFERED true
+ENV PYTHONDONTWRITEBYTECODE=true
+ENV PYTHONUNBUFFERED=true
```
danipran, line 10: Does this work? These are usually 1s

charn: Afaik they should work, documentation says "if this is set to a non-empty string" for both.

danipran: Could've googled this myself before asking. Yes, it does. https://docs.python.org/3/using/cmdline.html#envvar-PYTHONDONTWRITEBYTECODE

danipran (approved):
LGTM :+1:

PR #516, opened 2023-02-01
Make django-haystack manage elasticsearch package version

charn:
`django-haystack` has an optional extra requirement, which currently will limit `elasticsearch` package to `elasticsearch>=5,<8`


voneiden (approved):
Ooh, nice :clap:

PR #517, opened 2023-02-01
Migration related fixes

voneiden:
This PR fixes various migration related issues and enables CI migration check. 

@tuomas777 @dankiki : The increase of field length in first_name and last_name is up for debate. The original was 30. Any thoughts whether to keep the old 30 or migrate to the Django defaults of 150?


linkedevents/management/commands/check_migrations.py
```diff
@@ -0,0 +1,35 @@
+from django.db.migrations.loader import MigrationLoader
+from django.core.management import call_command
+from django.core.management.base import BaseCommand
+
+
+class Command(BaseCommand):
+    """
+    Note, to use you probably need to set EXTRA_INSTALLED_APPS=linkedevents
```
danipran, line 8: So do I need to or not? "Probably" isn't a good enough answer; please elaborate. For example, why? When?

danipran (changes requested):
Better help text/docstring would be nice.

helevents/migrations/0005_auto_20230201_1241.py
```diff
@@ -0,0 +1,27 @@
+# Generated by Django 3.2 on 2023-02-01 10:41
```
charn, line 1: Doing this migration is totally fine since it is generated due to changes in Django. If you want, you can rename the migration to something like `0005_increase_user_model_name_field_length.py`, but up to you! :+1:

voneiden:
The `check_migrations` command will likely be dropped from this PR in favour of direct fixes to the sources:

* https://github.com/City-of-Helsinki/django-orghierarchy/pull/29
* https://github.com/City-of-Helsinki/django-munigeo/pull/50

I'll rework this PR once the above have been resolved.

charn (approved):
LGTM! :+1: :shipit:

PR #518, opened 2023-02-02
Python 2 removals

charn:
Python 2 is not supported. Remove/refactor encoding related comment/code which has no effect in Python 3. Also removes a redundant `tests` module.


danipran (approved):
LGTM :+1:

PR #519, opened 2023-02-02
Prevent stopped docker-compose containers restarting on boot

charn:
`unless-stopped` is similar to `always` flag, but it doesn't restart any stopped containers when the Docker daemon restarts (e.g. on boot).

https://docs.docker.com/config/containers/start-containers-automatically/#use-a-restart-policy


danipran (approved):
LGTM 👍

PR #520, opened 2023-02-06
LINK-952: Add distance filter (aka "circle filter")

danipran:
Adds two new event filters, `dwithin_origin` and `dwithin_metres`, which filters all events within the given distance (in metres) from the origin point (x, y).

E.g.
`/v1/event/?dwithin_origin=24.941430575828296,60.17186028180228&dwithin_metres=1000`
Get all events within 1 km radius of Helsinki Central Station.

Refs LINK-952

### DIFF
--- a/events/api.py
+++ b/events/api.py
@@ -3645,12 +3645,18 @@ class EventFilter(django_filters.rest_framework.FilterSet):
             # Need both values for filtering
             return queryset
 
-        origin_x, origin_y = origin
-        metres = float(metres)
+        try:
+            origin_x, origin_y = [float(i) for i in origin]
+        except ValueError:
+            raise ParseError("Origin must be a tuple of two numbers")
+        try:
+            metres = float(metres)
+        except ValueError:
+            raise ParseError("Metres must be a number")
 
-        places = Place.geo_objects.filter(
-            position__distance_lte=(
-                Point(float(origin_x), float(origin_y)),
+        places = Place.objects.filter(
+            position__dwithin=(
+                Point(origin_x, origin_y),
                 D(m=metres),
             )
         )
--- a/events/api.py
+++ b/events/api.py
@@ -13,6 +13,7 @@ from typing import Iterable, Optional
 from uuid import UUID
 
 import bleach
+import django.forms
 import django_filters
 import environ
 import pytz
@@ -21,6 +22,7 @@ from django.conf import settings
 from django.contrib.auth.models import AnonymousUser
 from django.contrib.gis.db import models
 from django.contrib.gis.geos import Point
+from django.contrib.gis.measure import D
 from django.contrib.postgres.search import SearchQuery, TrigramSimilarity
 from django.core.cache import caches
 from django.core.exceptions import PermissionDenied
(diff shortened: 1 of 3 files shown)


events/tests/test_event_get.py
```diff
@@ -192,6 +192,30 @@ def test_get_event_list_verify_bbox_filter(api_client, event, event2):
...
+    # Sanity check for distances
+    assert origin.distance(event.location.position) > 36
+    assert origin.distance(event2.location.position) < 34
+
+    # Don't filter if missing either origin or metres.
+    get_list_and_assert_events(f"dwithin_origin={origin.x},{origin.y}", [event, event2])
+    get_list_and_assert_events("dwithin_metres=35", [event, event2])
+
+    get_list_and_assert_events(
+        f"dwithin_origin={origin.x},{origin.y}&dwithin_metres=35", [event2]
```
voneiden, line 207: Throwing in garbage non-numeric values produces 500 internal server error instead of  400 bad request..  although I don't know what is the general state of validation in this project. The above test on bbox does return 400 "bbox values must be floating points or integers", although that's thanks to munigeo doing the validation.

danipran: Yeah, I spotted that as well, wasn't quite sure whether to add error handling or not. I suppose I could add ParseErrors or something 🤔

events/api.py
```diff
@@ -3618,11 +3627,39 @@ class EventFilter(django_filters.rest_framework.FilterSet):
...
+        try:
+            origin_x, origin_y = [float(i) for i in origin]
+        except ValueError:
+            raise ParseError("Origin must be a tuple of two numbers")
+        try:
+            metres = float(metres)
+        except ValueError:
+            raise ParseError("Metres must be a number")
+
+        places = Place.geo_objects.filter(
```
tuomas777, line 3654: I think `geo_objects` is just some relic from the past, and you could and should use just `objects` instead.

events/api.py
```diff
@@ -3618,11 +3627,39 @@ class EventFilter(django_filters.rest_framework.FilterSet):
...
+            position__distance_lte=(
```
tuomas777, line 3655: IMO the filter should not be named `dwithin` when it actually does not use the [DWithin SQL function](https://postgis.net/docs/ST_DWithin.html) 🙂 `distance_lt` would be consistent with at least some of the other filters (`audience_max_age_lt`).

events/tests/test_event_get.py
```diff
@@ -192,6 +192,47 @@ def test_get_event_list_verify_bbox_filter(api_client, event, event2):
...
+        "dwithin_origin=0,0,0&dwithin_metres=35",
+        "dwithin_origin=0,foo&dwithin_metres=35",
+    ],
+)
+@pytest.mark.django_db
+def test_get_event_list_verify_dwithin_filter_error(
+    query_string, api_client, event, event2
+):
+    response = get_list_no_code_assert(api_client, query_string=query_string)
+    assert response.status_code == 400
```
tuomas777, line 233: Because it is pretty easy to get a 400 response from DRF when something else is wrong, I would like that this also checked that we are actually hitting the error message we are excepting.

events/tests/test_event_get.py
tuomas777: 500s will end up in Sentry, and we certainly don't want that 🙂

tuomas777 (commented):
Description of the new filter could be also added to https://github.com/City-of-Helsinki/linkedevents/blob/develop/helevents/templates/rest_framework/event_list.html

events/api.py
danipran: Right, not sure why I used `distance_lte` in the first place since I was going to use `dwithin` 😅 well spotted

danipran: Kept the name, but changed the query to use `dwithin` instead of `distance_lte`

voneiden (approved):

charn (approved):
LGTM! :+1: :shipit:

tuomas777 (approved):
Good stuff! 👍 🥇

PR #524, opened 2023-02-08
LINK-1137 | Unknown users to group

charn:
This PR adds `LinkedEventsJWTAuthentication` which will add user to a default organization upon authentication if user doesn't already belong to an organization.


events/tests/test_auth.py
```diff
@@ -42,3 +89,66 @@ def test_is_regular_user(self):
...
+    elif not default_organisation:
+        assert logged_user.organization_memberships.count() == 0
+    else:
+        assert logged_user.organization_memberships.count() == 1
+        set_org = logged_user.organization_memberships.first()
+
+        if default_organisation == "default":
+            assert set_org.id == HELSINKI_DEFAULT_ORGANIZATION_ID
+        elif default_organisation == "existing":
+            assert set_org.id == organization.id
```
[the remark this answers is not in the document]
charn: I'll do some refactoring here. :+1:

events/utils.py
```diff
@@ -6,11 +6,14 @@
...
 from django.conf import settings
 from django.db import transaction
+from django_orghierarchy.models import Organization
 from rest_framework.exceptions import ParseError
 
-from events.models import Keyword, Place
+from events.models import DataSource, Keyword, Place
 from events.sql import count_events_for_keywords, count_events_for_places
 
+DEFAULT_ORGANIZATION_ID = "others"
```
danipran, line 15: I think it might make more sense to have this in `settings` rather than `utils`.

linkedevents/settings.py
```diff
@@ -259,6 +261,12 @@ def get_git_revision_hash() -> str:
 ACCOUNT_LOGOUT_ON_GET = True
 SOCIALACCOUNT_ADAPTER = "helusers.adapter.SocialAccountAdapter"
 
+# Default organization to which a user will be assigned if user doesn't
+# have any organization.
+USER_DEFAULT_ORGANIZATION_ID = env("USER_DEFAULT_ORGANIZATION_ID")
```
danipran, line 266: What's the difference between `DEFAULT_ORGANIZATION_ID` and `USER_DEFAULT_ORGANIZATION_ID` atm? 🤔 Do we have a case where these need to be separate values?

charn: Ideally they should be the same value yes. `USER_DEFAULT_ORGANIZATION_ID` sort of controls whether the "put user without an organization into the default organization" feature is enabled. That's probably not optimal. Maybe I'll still refactor this so that we have a setting like `ENABLE_USER_DEFAULT_ORGANIZATION` and `USER_DEFAULT_ORGANIZATION_ID` (and the default org would automatically get created if it doesn't exist. :thinking:

linkedevents/settings.py
```diff
@@ -292,6 +300,7 @@ def get_git_revision_hash() -> str:
     "JWT_PAYLOAD_GET_USER_ID_HANDLER": "helusers.jwt.get_user_id_from_payload_handler",
     "JWT_AUDIENCE": env("TOKEN_AUTH_ACCEPTED_AUDIENCE"),
     "JWT_SECRET_KEY": env("TOKEN_AUTH_SHARED_SECRET"),
+    "JWT_AUTH_HEADER_PREFIX": "Bearer",
```
charn, line 303: I set up tunnistamo locally and tested that without this change the authentication didn't work. :shrug:

voneiden: I did a deep dive into the code and for all I can tell this seems to be the way. But, there's one other Helsinki project with a similar configuration, but without JWT_AUTH_HEADER_PREFIX:

https://github.com/City-of-Helsinki/respa/blob/develop/respa/settings.py#L302 

Hmm hmm..

voneiden: In dev env the backend has the default "JWT" value for  JWT_AUTH_HEADER_PREFIX. However frontend requests depart with "Bearer" (or also seen lowercase "bearer")

Either I'm completely misunderstanding this or there's something that flips Bearer -> JWT in between frontend and backend?

charn: @dankiki Any thoughts on this? :sweat_smile:

danipran (approved):
LGTM :+1: I guess the JWT discussion needs to be resolved before merging, though 🤔

linkedevents/settings.py
voneiden: I'm suspicious about using `helusers.jwt.JWTAuthentication` in the first place. Looks like Yrjölä added it into helusers in 2015 ( https://github.com/City-of-Helsinki/django-helusers/commit/469c2932575daaf42d8cec5578c087f4e5c340af ) and around the same time into  linkedevents. helusers readme doesn't mention this at at all - instead helusers tells to use  helusers.oidc.ApiTokenAuthentication for django JWT stuff in the [readme](https://github.com/City-of-Helsinki/django-helusers#django-rest-framework-api-authentication-using-jwt). ApiTokenAuthentication (and RequestJWTAuthentication) are both newer features.

I get a feeling that `helusers.jwt.JWTAuthentication` is unmaintained and exists as mainly because nobody dares to remove it. 

Lets still try to get some confirmation whether something unusual happens to the header on the way though..

charn: Since `JWTAuthentication` is the one that is in use currently, I'd say that we could make updating to OIDC a separate task, since that most likely also requires refactoring the settings/pipeline configuration. The logic which I added in this PR should afaik be quite easily transferable over to `helusers.oidc.ApiTokenAuthentication`?

voneiden: My current assumption is that JWTAuthentication is actually overriden by local_settings.py in ninja/prod environments as ApiTokenAuthentication.  The error message I get out with a invalid token from the ninja env matches the error message ApiTokenAuthentication would produce. JWTAuthentication would not produce that message.

But I agree a separate issue might be a good idea to convert all references of JWTAuthentication to ApiTokenAuthentication once we have clarity on this matter.

voneiden (approved):
Good to go if we split the JWT/Authentication method thing into a separate task/issue.

PR #525, opened 2023-02-08
Fix light admin migration

voneiden:
The original implementation failed to link permissions to the group because normally Django updates permissions after all migrations have been applied. With this fix we force the update of permissions to ensure they can be added.

EDIT note permissions need to be fixed still manually in existing environments. Is this OK or should I add a new migration that simply  checks the permissions are not missing and adds them if they are?


danipran:
> note permissions need to be fixed still manually in existing environments. Is this OK or should I add a new migration that simply checks the permissions are not missing and adds them if they are?

If it's something that needs to be fixed in existing environments then yes, I think there should be a new migration for that. Less manual work == better, IMO.

danipran (approved):
LGTM :+1:

PR #526, opened 2023-02-08
Fix SonarCloud warnings

tuomas777:
SonarCloud was giving warnings after scanning, so two changes were needed:
* changed the GHA job to not use a shallow clone ([more info](https://community.sonarsource.com/t/sonarcloud-with-github-actions/22713/2))
* added a workaround that overrides source path in the generated coverage report file ([more info](https://community.sonarsource.com/t/sonar-on-github-actions-with-python-coverage-source-issue/36057))

Also excluded tests from Sonar scans.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -102,6 +102,10 @@ jobs:
       - name: Codecov
         uses: codecov/codecov-action@v3
 
+      # Without this workaround Sonar reports a warning about an incorrect source path
+      - name: Override coverage report source path for Sonar
+        run: sed -i 's@'$GITHUB_WORKSPACE'@/github/workspace/@g' coverage.xml
+
       - name: SonarCloud Scan
         uses: SonarSource/sonarcloud-github-action@master
         env:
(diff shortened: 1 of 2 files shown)


danipran (approved):
LGTM :+1:

voneiden (commented):
Confirm whether `sonar.tests.inclusions` is a valid property?  See: https://docs.sonarqube.org/9.6/project-administration/narrowing-the-focus/#exclusion-and-inclusion

> Naming of parameters

> Note that the initial scoping parameter for test code is sonar.tests (that's tests with an s!) while the exclusion and inclusion parameters for test code are sonar.test.exclusions and sonar.test.inclusions (that's test, without an s!).

tuomas777:
> Confirm whether `sonar.tests.inclusions` is a valid property? See: https://docs.sonarqube.org/9.6/project-administration/narrowing-the-focus/#exclusion-and-inclusion
> 
> > Naming of parameters
> 
> > Note that the initial scoping parameter for test code is sonar.tests (that's tests with an s!) while the exclusion and inclusion parameters for test code are sonar.test.exclusions and sonar.test.inclusions (that's test, without an s!).

Wow, that is some 🦅 👁️ right there 😲 Fixing.

voneiden (approved):

PR #528, opened 2023-02-13
LINK-952: Add SRID in distance filter, add SRID query parameter documentation

danipran:
The dwithin filter (or the "circle filter") didn't work; it was missing the correct SRID. Oops.

Added the SRID parameter, similar to how it's implemented for the bbox filter.

Turns out there's also an undocumented SRID query parameter, so added documentation for that.


events/api.py
```diff
@@ -3654,9 +3657,10 @@ def filter_dwithin(queryset, name, value: tuple[tuple[str, str], str]):
         except ValueError:
             raise ParseError("Metres must be a number")
 
+        print(self.srs.srid)
```
voneiden, line 3660: 👀

danipran: 🤡

helevents/templates/rest_framework/event_list.html
```diff
@@ -72,7 +72,8 @@ <h4 id="bounding-box">Bounding box</h4>
 </code></pre>
 <p>Where <code>west</code> is the longitude of the rectangle's western boundary,
 <code>south</code> is the latitude of the rectangle's southern boundary,
-and so on. The coordinate system is the trusty old EPSG:4326 known from all online maps.</p>
+and so on. The default coordinate system is the trusty old EPSG:4326 known from all online
+maps, and can be overriden with <code>srid</code> parameter.</p>
```
tuomas777, line 76:
```suggestion
maps, and can be overridden with <code>srid</code> parameter.</p>
```

voneiden (approved):
Doesn't concern the code in this PR, but the existing srid parsing validates number but doesn't validate it's a valid option, ie. srid=0 produces 500 internal server error

```
GDALException at /v1/event/
OGR failure.
```
At least on my computer. I see there is already handling for some exception in `def srid_to_srs(srid):`

```
    try:
        srs = SpatialReference(srid)
    except SRSException:
        raise ParseError("SRID %d not found (try 4326 for GPS coordinate system)" % srid)
```

Issue #267, opened 2018-05-29
Allow POSTing full-day events


Rikuoja:
This means events which only have a date (not time) in the `start_time` field, and optionally a date (not time) in the `end_time` field.

Rikuoja:
Required by https://github.com/City-of-Helsinki/linkedevents-ui/pull/274

PR #529, opened 2023-02-16
Fix overflow and DST handling in parse_time and add tests

voneiden:
Warning: Don't hurt yourself trying to read this.

Backstory: So.. in the Events API, the fields `start_time`, `end_time` and `date_published` are DateTimeFields, but utilizing a surprise `to_internal_value` method  some preprocessing is done to allow them also be date strings or special nouns like `now` or `today`. Additionally for dates (and `today`)  there's a concept of `is_start`, which if `False`, means the date is actually not the given date but the next day at 00:00:00. Because there are two different virtual types living the same field, there's a concept of `is_exact`. If the object is not exact, it means it's a date (disguised as a datetime). This information is stored in a helper field `has_start_time` which is `False` if the object is not `is_exact`.  This allows us to use a condition like follows to check if the object is a date disguised as a datetime:

`if obj.start_time and not obj.has_start_time:`

Food for thought: I'm not sure if the real world actually ever POSTs anything other than valid DateTimeField values. However there are 60 test cases that fail if the support for posting dates is dropped. 

The OG feature was made in https://github.com/City-of-Helsinki/linkedevents/issues/267 to support old frontend https://github.com/City-of-Helsinki/linkedevents-ui/issues/54 but I don't think linkedcomponents has this kind of "whole day" concept.

### DIFF
--- a/events/api.py
+++ b/events/api.py
@@ -2509,9 +2509,13 @@ class EventSerializer(
             val = data.get(field, None)
             if val:
                 if isinstance(val, str):
-                    data[field], data["has_" + field] = utils.parse_time(
-                        val, not field == "end_time"
-                    )
+                    if field == "end_time":
+                        dt, is_date = utils.parse_end_time(val)
+                    else:
+                        dt, is_date = utils.parse_time(val)
+
+                    data[field] = dt
+                    data[f"has_{field}"] = not is_date
         return data
 
     def to_internal_value(self, data):
@@ -2596,12 +2600,7 @@ class EventSerializer(
                 data["end_time"] = None
             else:
                 data["has_end_time"] = False
-                data["end_time"] = (
-                    timezone.localtime(data["start_time"])
-                    .replace(hour=0, minute=0, second=0, microsecond=0)
-                    .astimezone(pytz.utc)
-                )
-                data["end_time"] += timedelta(days=1)
+                data["end_time"] = utils.start_of_next_day(data["start_time"])
 
         past_allowed = self.data_source.create_past_events
         if self.instance:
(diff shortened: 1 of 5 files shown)


voneiden:
The answer to the last question is: probably maybe and therefore API backwards compatibility will be maintained.

danipran:
wtf

Any way of making this function/its usage more readable/maintainable?

voneiden:
Renaming `is_exact` and `has_xyz` could already go a long way. I can give it a shot

events/tests/test_utils.py
```diff
@@ -0,0 +1,79 @@
...
+        local_tz.localize(datetime(2022, 7, 17, 17, 17, 17)),
+        True,
+    )
+
+
+@pytest.mark.django_db
+def test_parse_time_iso8601_invalid_offset(local_tz):
+    with pytest.raises(ParseError) as ex:
+        parse_time("2022-07-17T17:17:17+25:00", True)
+        assert "out of bounds" in str(ex.value)
```
charn, line 30: Afaik, `parse_time` raises so the assertation clause is never reached unless you make the following change. And making that change seems to make the test fail since the error message is not what is expected here. :eyes: 
```suggestion
    with pytest.raises(ParseError) as ex:
        parse_time("2022-07-17T17:17:17+25:00", True)
    assert "out of bounds" in str(ex.value)
```

events/tests/test_utils.py
```diff
@@ -0,0 +1,79 @@
...
+        parse_time("2022-07-17T17:17:17+23:00", True)
+        assert "out of bounds" in str(ex.value)
+
+
+@pytest.mark.django_db
+def test_parse_time_invalid_format(local_tz):
+    with pytest.raises(ParseError) as ex:
+        # Different error deep under the hood than above
+        parse_time("asdfdsaasdf", True)
+        assert "time in invalid format" in str(ex.value)
```
charn, line 43: assert needs to be moved :point_left:

events/tests/test_utils.py
```diff
@@ -0,0 +1,79 @@
...
+@pytest.mark.django_db
+def test_parse_time_iso8601_invalid_offset(local_tz):
+    with pytest.raises(ParseError) as ex:
+        parse_time("2022-07-17T17:17:17+25:00", True)
+        assert "out of bounds" in str(ex.value)
+
+    with pytest.raises(ParseError) as ex:
+        # Different error deep under the hood than above
+        parse_time("2022-07-17T17:17:17+23:00", True)
+        assert "out of bounds" in str(ex.value)
```
charn, line 35: assert needs to be moved :point_left:

events/tests/test_utils.py
voneiden: Excellent, thanks. Not sure how that happened. Probably copilot :trollface:

danipran (approved):
LGTM :+1: thanks for making it more readable! Still a bit wtf, but better not touch the underlying logic for now...

events/api.py
```diff
@@ -3271,9 +3274,8 @@ def _filter_event_queryset(queryset, params, srs=None):  # noqa: C901
         )
 
     if end:
-        dt = utils.parse_time(end, is_start=False)[0]
-        if not dt.tzinfo:
-            dt = dt.astimezone(pytz.timezone("UTC"))
+        # Inconsistent behaviour, see test_inconsistent_tz_default
+        dt = utils.parse_end_time(end, default_tz=pytz.timezone(settings.TIME_ZONE))[0]
```
voneiden, line 3278: @tuomas777 any thoughts on this? TL;DR create and list event use different tz for naive date(times). List localizes to settings.TIME_ZONE and create sets them to UTC.

PR #530, opened 2023-02-17
Update azure-pipelines-dev.yml for Azure Pipelines

d812k:


requirements.txt
```diff
@@ -93,7 +93,7 @@ django-mptt==0.13.4
     #   -r requirements.in
     #   django-munigeo
     #   django-orghierarchy
-django-munigeo @ git+https://github.com/dankiki/django-munigeo.git@0.3.10
+django-munigeo==0.3.8
```
charn, line 96: This will need a change in `requirements.in` as well.

charn:
Not required any more

PR #531, opened 2023-02-17
Handle environment variables through settings

charn:
This is the way. There's no added benefit to doing it any other way. This should also fix some test failures due to missing `SEAT_RESERVATION_DURATION` env variable (since I used a default value of 15 minutes).


danipran (approved):
This is the way.

voneiden (approved):
He will show us the way

PR #532, opened 2023-02-24
Refactor events api permission checking and fix LINK-1118

voneiden:
Background: In LINK-1118 attempting to attach an image that the user did not have permission to edit led to a permission error because of the complex nature of permissions checking in the events API, especially because of doing permission checks in the LinkedEventsObjectSerializer superclass. This PR fixes that particular issue by doing some refactoring. Mostly reusing old code. Key goals:

1) Move generic permission checking in to a permission class
2) Try to keep serializer context in .. serializer.context instead of  using custom self attributes. 
3) Prefer DRF compatibile behaviour (with super() calls) in viewset methods instead completely custom solutions
4) Avoid touching existing tests except for a few cases where slight refactoring was necessary

It's not perfect though, there are still quite a few permission checks littered in the views, but at least the serializers are now almost free of permission logic.


events/api.py
```diff
@@ -1115,7 +1116,7 @@ def get_queryset(self):
         return queryset
 
 
-register_view(KeywordRetrieveViewSet, "keyword")
+register_view(KeywordViewSet, "keyword")
```
charn, line 1119: What was this change?

voneiden: KeywordRetrieveViewSet renamed to more generic KeywordViewSet since the viewset has UpdateModelMixin and DestroyModelMixin.

charn: Ah, indeed. :+1:

events/api.py
```diff
@@ -3778,7 +3727,7 @@ def filter_queryset(self, queryset):
         TODO: convert to use proper filter framework
         """
         original_queryset = super().filter_queryset(queryset)
-        if self.request.method in SAFE_METHODS:
+        if self.action in "list":
```
charn, line 3730: This works, but should it be?
```suggestion
        if self.action == "list":
```

events/api.py
```diff
@@ -3806,17 +3755,19 @@ def filter_queryset(self, queryset):
...
+                    original_queryset = Event.objects.filter(
+                        id__in=[i.get("id", "") for i in self.request.data]
+                    )
+                except:  # noqa E722
+                    raise DRFPermissionDenied("Invalid JSON in request.")
 
-        if self.request.method == "GET":
+                queryset = self.request.user.get_editable_events(original_queryset)
+
+        if self.action == "list":
```
charn, line 3770: This if clause could be combined with the `if self.action == "list"` above?

charn (approved):
As far as I can tell this will be an improvement since we're adding proper permission classes and getting rid of some of the permission checks baked deep into the viewset/serializer code. :+1: LGTM :+1: :shipit:

danipran:
Originally, we decided to postpone merging this to after LinkedEvents had been migrated to Platta. However, it's still not finished, and this PR is becoming a bit of a blocker, so we're gonna go along and merge it anyway.

Unfortunately, there's some big conflicts with https://github.com/City-of-Helsinki/linkedevents/pull/563, so this required a bit of resolving. Instead of rebasing this, created a new merge commit (cd4923da8b853b7de797466f82545f4cbc18d58e), as I'm not 100% sure if I got it right.

Still need to clean up the commit log and apply Juha's earlier suggestion, but should otherwise be ready for review.

events/api.py
```diff
@@ -1799,19 +1758,6 @@ def get_queryset(self):
                 queryset = queryset.none()
         return queryset
 
-    def destroy(self, request, *args, **kwargs):
-        qset = Organization.objects.filter(id=kwargs["pk"])
-        if qset.count == 0:
```
charn, line 1804: Good to get rid of this (it should've been `count()`).
```
In [3]: Organization.objects.none().count == 0
Out[3]: False
```

PR #534, opened 2023-03-08
Increase ImageField max_length

voneiden:
Increase Event.image file name max_length from default 100 to 255.

Goes hand in hand with City-of-Helsinki/linkedcomponents-ui/pull/158  where the max file length is limited to 200 to leave some room for "images/" prefix and a unique suffix that can be added in the backend processing.


danipran (approved):
LGTM :+1:

PR #541, opened 2023-03-13
LINK-1122 | Fix RegistrationViewSet checks permission on delete

charn:
There wasn't any permission checks done for DELETE since `DestroyModelMixin` doesn't use the serializer at all. Currently permission check is done only in the serializer.

Also refactor the `test_basic_registration_functionality` into multiple unit test, which only test one thing.

Refs LINK-1122


voneiden (approved):
Nice catch, I guess this would've slipped past #532

PR #545, opened 2023-03-17
LINK-1158 | Get SECRET_KEY from the environment variable by default

charn:
Secret key from a generated file was used by default, which means it changes every time a new container is started. The `SECRET_KEY` environment variable was not actually read into the `SECRET_KEY` setting.


danipran (approved):
LGTM :+1:

PR #547, opened 2023-03-20
Unify mailgun environment variable naming

jorilindell:
## Description
config_dev.toml.example still has some old mailgun environment variable. Use same variable names as in linkedevents/settings.py


charn (approved):
LGTM! :+1: :shipit:

PR #554, opened 2023-03-23
Update CustomEsSearchBackend to use elasticsearch7_backend

voneiden:
Now that platta is running ES7 switch to using elasticsearch7_backend in django-haystack as pointed out by @charn


voneiden:
Tested to have a positive improvement, eg. https://linkedevents-api-dev.agw.arodevtest.hel.fi/v1/search/?type=event&input=testitapahtuma now returns two events.

charn (approved):
Awesome! :+1: :shipit:

PR #556, opened 2023-03-24
LINK-1027 | Fix "some registrations not found" bug

jorilindell:
## Description
Don't filter registrations by event_type by default. It causes 404 errors when requesting a single registration which has event_type course or volunteering

## Closes
[LINK-1027](https://helsinkisolutionoffice.atlassian.net/browse/LINK-1027)

## Testing
**Producing the bug before the fix:**
1. Create an a new event
2. Create a registrations and use created event in it
3. Change event type of the event to course or volunteering
http://localhost:8080/v1/registration/[id]/ returns 404 error

e.g. https://linkedevents-api-dev.agw.arodevtest.hel.fi/v1/registration/28/?include=signups&nocache=true

[LINK-1027]: https://helsinkisolutionoffice.atlassian.net/browse/LINK-1027?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/registrations/api.py
+++ b/registrations/api.py
@@ -157,7 +157,18 @@ class RegistrationViewSet(
 
     def filter_queryset(self, queryset):
         events = Event.objects.exclude(registration=None)
-        events = _filter_event_queryset(events, self.request.query_params)
+
+        # Copy query_params to get mutable version of it
+        query_params = self.request.query_params.copy()
+        # By default _filter_event_queryset only returns events with GENERAL type.
+        # This causes problem when getting a registration details, so filter registrations
+        # by event_type only when explicitly set
+        if not query_params.get("event_type"):
+            event_types = {k[1].lower(): k[0] for k in Event.TYPE_IDS}
+            query_params["event_type"] = ",".join(event_types.keys())
+
+        events = _filter_event_queryset(events, query_params)
+
         val = self.request.query_params.get("admin_user", None)
         if val and str(val).lower() == "true":
             if isinstance(self.request.user, AnonymousUser):
--- a/registrations/tests/test_registration_admin_side.py
+++ b/registrations/tests/test_registration_admin_side.py
@@ -54,6 +54,22 @@ def test_create_registration(api_client, event, has_right_to_edit, expected_stat
     assert response.status_code == expected_status
 
 
+@pytest.mark.parametrize(
+    "event_type",
+    [Event.TypeId.GENERAL, Event.TypeId.COURSE, Event.TypeId.VOLUNTEERING],
+)
+@pytest.mark.django_db
+def test_get_registration(api_client, event, event_type, registration):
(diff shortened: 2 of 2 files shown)


registrations/api.py
```diff
@@ -157,7 +157,15 @@ def destroy(self, request, *args, **kwargs):
 
     def filter_queryset(self, queryset):
         events = Event.objects.exclude(registration=None)
-        events = _filter_event_queryset(events, self.request.query_params)
+
+        query_params = self.request.query_params.copy()
+        # By default don't filter events by event_type
```
charn, line 162: Indeed it seems that without this, you would only get events with `GENERAL` type from `_filter_event_queryset`. Maybe something about that could be added to the comment just to document this part for later generations.

registrations/tests/test_registration_admin_side.py
```diff
@@ -54,6 +54,21 @@ def test_create_registration(api_client, event, has_right_to_edit, expected_stat
...
+)
+@pytest.mark.django_db
+def test_get_registration(api_client, event, event_type, registration):
+    event.type_id = event_type
+    event.save()
+
+    registration_url = reverse("registration-detail", kwargs={"pk": registration.id})
+    response = api_client.get(registration_url)
+
+    assert response.status_code == status.HTTP_200_OK
```
danipran, line 69: Check that the content is somewhat sane as well. Or at least in my experience, 200 OK doesn't always mean it's actually OK.
```suggestion
    assert response.status_code == status.HTTP_200_OK
    assert response.data["id"] == registration.id
```

registrations/api.py
jorilindell: > Also, does the copy serve some purpose?

The purpose is to get mutable version of query_params. `query_params["event_type"] = ",".join(event_types.keys())` throws **AttributeError: This QueryDict instance is immutable** error without the copy.
Added a comment and hopefully it makes more sense now.

Other option would be set query_params mutable explicitly: `query_params._mutable=True`

jorilindell: > Indeed it seems that without this, you would only get events with `GENERAL` type from `_filter_event_queryset`. Maybe something about that could be added to the comment just to document this part for later generations.

Added a comment to clarify the reason to add event_type attribute if not set.

registrations/tests/test_registration_admin_side.py
jorilindell: Totally agree. Added an assertion for registration id

registrations/api.py
voneiden: Oh I see, didn't realize it's immutable. I think copy is better than changing _mutable. :+1:

voneiden (approved):
LGTM

PR #557, opened 2023-03-27
LINK-1250 | Events without registration

jorilindell:
## Description
Modify registration parameter to return events with a registration when true, and without a registration when false

## Closes
[LINK-1250](https://helsinkisolutionoffice.atlassian.net/browse/LINK-1250)

[LINK-1250]: https://helsinkisolutionoffice.atlassian.net/browse/LINK-1250?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/events/api.py
+++ b/events/api.py
@@ -2522,9 +2522,13 @@ def _filter_event_queryset(queryset, params, srs=None):  # noqa: C901
     # Filter by string (case insensitive). This searches from all fields
     # which are marked translatable in translation.py
 
+    # filter to get events with or without a registration.
     val = params.get("registration", None)
-    if val:
+    if val and parse_bool(val, "registration"):
         queryset = queryset.exclude(registration=None)
+    elif val:
+        queryset = queryset.filter(registration=None)
+
     val = params.get("enrolment_open", None)
     if val:
         queryset = (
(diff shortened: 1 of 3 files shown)


events/api.py
```diff
@@ -2522,9 +2522,16 @@ def _filter_event_queryset(queryset, params, srs=None):  # noqa: C901
...
 
+    # filter to get events with a registration.
     val = params.get("registration", None)
     if val:
         queryset = queryset.exclude(registration=None)
+
+    # filter to get events without a registration.
+    val = params.get("registration!", None)
+    if val:
+        queryset = queryset.filter(registration=None)
```
charn, line 2533: It seems that using an exclamation mark is an existing pattern here, so I'm not going to say anything about that. Though I find it a bit misleading that you can get the same results/filtering with `registration=true` and `registration=false`,

Maybe the whole thing could be replaced with something like this?


```suggestion
    # filter to get events with or without a registration.
    val = params.get("registration", None)
    if val and parse_bool(val):
        queryset = queryset.exclude(registration=None)
    elif val: 
        queryset = queryset.filter(registration=None)
```

jorilindell: Very good idea. Changed as suggested, fixed the tests and updated the documentation

voneiden: Just a general note, this "do something if value is defined regardless of what the value is" elsewhere also an existing pattern linkedevents. But  :+1: for avoiding that.

voneiden (approved):
LGTM

PR #560, opened 2023-03-29
LINK-1224 | Replace memcached cache backend with Redis

charn:
- Add django-redis as a requirement and remove python-memcached
- Replace memcached cache with Redis
- Add REDIS_URL to .env.example
- Add redis container and remove memcached in docker-compose.yml
- Remove separate ongoing_events cache, just use a separate timeout
- Ignore, but log errors related to Redis cache. This is similar
  to memcached in a way that connection error to memcached doesn't
  raise an exception.
- Only enable Redis cache if it's configured, use LocMemCache as default.

Note: Used https://github.com/City-of-Helsinki/kaavapino/blob/development/kaavapino/settings.py#L124 and https://github.com/City-of-Helsinki/kaavoitus-api/blob/development/api_project/settings.py#L174 as examples.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -20,7 +20,7 @@ repos:
       - id: flake8
         additional_dependencies: [pep8-naming, flake8-bugbear]
         exclude: "migrations"
-  - repo: https://github.com/timothycrosley/isort
+  - repo: https://github.com/pycqa/isort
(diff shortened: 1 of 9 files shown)


d812k (approved):
Looks good to me.

charn:
Made small change, that `LocMemCache` will be used if Redis is not configured.

PR #567, opened 2023-04-12
LINK-488 | Exclude search vector fields from the event output

jorilindell:
## Description
Currently search_vector_fi sv and en are included in the event, but they are needed only internally. Exclude them from the output

## Closes
[LINK-488](https://helsinkisolutionoffice.atlassian.net/browse/LINK-488)


charn (approved):
LGTM!  :+1: :shipit:

PR #569, opened 2023-04-14
Remove deprecated codecov dependency from CI

danipran:
> On April 12, 2023, Codecov removed the codecov package from PyPI. Our intent was to remove a deprecated, rarely-used package from active support.

https://about.codecov.io/blog/message-regarding-the-pypi-package/

This PR fixes the broken CI caused by this.


tuomas777 (approved):
👍

PR #571, opened 2023-04-18
LINK-664 | Support include parameter in registration endpoints

jorilindell:
## Description
At the moment UI has to first load registration data and after that event information in separate request. Motivation for this PR is to allow to request event data via registration endpoints

## Closes
[LINK-664](https://helsinkisolutionoffice.atlassian.net/browse/LINK-664)

## Testing
e.g. http://localhost:8080/v1/registration/?include=event,location,keywords,in_language,audience

[LINK-664]: https://helsinkisolutionoffice.atlassian.net/browse/LINK-664?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Link to UI PRs: https://github.com/City-of-Helsinki/linkedcomponents-ui/pull/175 and https://github.com/City-of-Helsinki/linkedregistrations-ui/pull/48

### DIFF
--- a/events/api.py
+++ b/events/api.py
@@ -106,7 +106,7 @@ from events.translation import (
 from helevents.models import User
 from helevents.serializers import UserSerializer
 from linkedevents.registry import register_view, viewset_classes_by_model
-from registrations.serializers import RegistrationSerializer
+from registrations.serializers import RegistrationBaseSerializer
 
 LOCAL_TZ = pytz.timezone(settings.TIME_ZONE)
 
@@ -351,7 +351,7 @@ class JSONLDRelatedField(relations.HyperlinkedRelatedField):
                 context["include"] = [
                     x
                     for x in context["include"]
-                    if x != "sub_events" and x != "super_event"
+                    if x != "sub_events" and x != "super_event" and x != "registration"
                 ]
             return self.related_serializer(
                 obj, hide_ld_context=self.hide_ld_context, context=context
@@ -559,18 +559,14 @@ class TranslatedModelSerializer(serializers.ModelSerializer):
 
 class LinkedEventsSerializer(TranslatedModelSerializer, MPTTModelSerializer):
     """Serializer with the support for JSON-LD/Schema.org.
-
     JSON-LD/Schema.org syntax::
-
       {
          "@context": "http://schema.org",
          "@type": "Event",
          "name": "Event name",
          ...
       }
-
     See full example at: http://schema.org/Event
-
     Args:
       hide_ld_context (bool):
         Hides `@context` from JSON, can be used in nested
@@ -1948,6 +1944,64 @@ class VideoSerializer(serializers.ModelSerializer):
     class Meta:
         model = Video
         exclude = ["id", "event"]
+
+
+# RegistrationSerializer is in this file to avoid circular imports
+class RegistrationSerializer(LinkedEventsSerializer, RegistrationBaseSerializer):
+    event = JSONLDRelatedField(
+        serializer="EventSerializer",
+        many=False,
+        view_name="event-detail",
+        queryset=Event.objects.all(),
+    )
+
+    def __init__(self, instance=None, *args, **kwargs):
+        super().__init__(instance=instance, *args, **kwargs)
+
+        event_dict = kwargs["context"]["request"].data.get("event", None)
+
+        # Check user permissions if creating a new registration.
+        # LinkedEventsSerializer __init__ checks permissions in case of new registration
+        if not instance and isinstance(event_dict, dict) and "@id" in event_dict:
+            event_url = event_dict["@id"]
+            event_id = urllib.parse.unquote(event_url.rstrip("/").split("/")[-1])
+            user = kwargs["context"]["user"]
+
+            try:
+                event = Event.objects.get(pk=event_id)
+                error_message = _("User {user} cannot modify event {event}").format(
+                    user=user, event=event
+                )
+
+                if not user.is_admin(event.publisher):
+                    raise DRFPermissionDenied(error_message)
+
+                if isinstance(user, ApiKeyUser):
+                    # allow updating only if the api key matches instance data source
+                    if event.data_source != user.data_source:
+                        raise DRFPermissionDenied(_(error_message))
+                else:
+                    # allow updating only if event data_source has user_editable_resources set to True
+                    if not event.is_user_editable_resources():
+                        raise DRFPermissionDenied(_(error_message))
+            except Event.DoesNotExist:
+                pass
+
+    def create(self, request, *args, **kwargs):
+        try:
+            instance = super().create(request, *args, **kwargs)
+        except IntegrityError as error:
+            if "duplicate key value violates unique constraint" in str(error):
+                raise serializers.ValidationError(
+                    {"event": _("Event already has a registration.")}
+                )
+            else:
+                raise error
+        return instance
+
+    # LinkedEventsSerializer validates name which doesn't exist in Registration model
+    def validate(self, data):
+        return data
 
 
 class EventSerializer(BulkSerializerMixin, EditableLinkedEventsObjectSerializer):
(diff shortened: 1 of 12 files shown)


events/api.py
```diff
@@ -599,9 +599,16 @@ def __init__(
             # check permissions *before* validation
             if not instance.can_be_edited_by(self.user):
                 raise PermissionDenied()
+
             if isinstance(self.user, ApiKeyUser):
+                # Registration uses event data source for permission check
```
charn, line 604: The other way for this would be to add a property to `Registration` model:

```python
@property
def data_source(self):
    return instance.event.data_source
```

But, your call. :+1:

events/api.py
```diff
@@ -2001,6 +2008,46 @@ class Meta:
...
+
+    def get_current_attendee_count(self, obj):
+        return SignUp.objects.filter(
+            registration__id=obj.id, attendee_status=SignUp.AttendeeStatus.ATTENDING
+        ).count()
+
+    def get_current_waiting_list_count(self, obj):
+        return SignUp.objects.filter(
+            registration__id=obj.id, attendee_status=SignUp.AttendeeStatus.WAITING_LIST
+        ).count()
```
charn, line 2041:
```suggestion
        return obj.signups.filter(attendee_status=SignUp.AttendeeStatus.WAITING_LIST).count()
```

events/api.py
danipran: Yeah, I definitely agree with @charn here, you should do this instead. This class is already making assumptions about what attributes/methods the instances have (e.g. `can_be_edited_by`), which you are implementing in the `Registration` model, so it's a bit of a weird difference.

An even better solution would probably not make assumptions and use interfaces/mixins or something, but that might not be exactly straight-forward.

registrations/serializers.py
```diff
@@ -66,27 +71,82 @@ class Meta:
...
+            except Event.DoesNotExist:
                 pass
+
+    def create(self, request, *args, **kwargs):
+        try:
+            instance = super().create(request, *args, **kwargs)
+        except IntegrityError as error:
+            if "duplicate key value violates unique constraint" in str(error):
+                raise serializers.ValidationError(
+                    {"event": _("Event already has a registration.")}
```
danipran, line 145: Is this message correct? I assume an event can have multiple registrations. Or am I missing a context here? 🤔

registrations/serializers.py
```diff
@@ -66,27 +71,82 @@ class Meta:
...
+                    if not event.is_user_editable_resources():
+                        raise DRFPermissionDenied(_(error_message))
+            except Event.DoesNotExist:
                 pass
+
+    def create(self, request, *args, **kwargs):
+        try:
+            instance = super().create(request, *args, **kwargs)
+        except IntegrityError as error:
+            if "duplicate key value violates unique constraint" in str(error):
```
danipran, line 143: This is an interesting way of checking whether a registration exists already or not. Any other way of checking this (rather than relying on an error message)? I mean, I suppose it's valid if there's a very specific reason for doing this this way, but otherwise, seems a bit unconventional.

events/api.py
```diff
@@ -2001,6 +2008,46 @@ class Meta:
...
+    last_modified_time = DateTimeField(
+        default_timezone=pytz.UTC, required=False, allow_null=True
+    )
+
+    created_by = serializers.StringRelatedField(required=False, allow_null=True)
+
+    last_modified_by = serializers.StringRelatedField(required=False, allow_null=True)
+
+    def get_current_attendee_count(self, obj):
+        return SignUp.objects.filter(
```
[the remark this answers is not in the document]
jorilindell: Changed

events/api.py
jorilindell: I prefer having data_source property in Registration modal. Checking model type in LinkedEventsSerializer felt a little bit hack to me

danipran (commented):
The more I look at it, the more I find `LinkedEventsSerializer` very problematic in general. It tries to be some kind of a do-it-all serializer but doesn't really work for everything. However, keeping the scope of the PR in mind, just probably have to deal with it for now.

Have some comments that *aren't* related to `LinkedEventsSerializer` though.

registrations/serializers.py
tuomas777: https://github.com/City-of-Helsinki/linkedevents/blob/d945861f0d0dceb543bdfc5407b64dda75c91bf6/registrations/models.py#L17
🙂

registrations/serializers.py
tuomas777: FWIW I don't think there really is a better way to do that. Or do you have some suggestion?

registrations/serializers.py
danipran: Well... that settles that 😂

registrations/serializers.py
danipran: Hmm, well, the reason I asked this question was because I'm not actually sure if there's a better way of doing this. 🙂 It might very well be the way to go, and it seemed like a very deliberate choice. I can take a look if there's a better way of doing this, but it might've been considered already, which, in turn, would make my premature search for alternatives a bit redundant.

events/api.py
```diff
@@ -2001,6 +2008,46 @@ class Meta:
         exclude = ["id", "event"]
 
 
+# Simplified RegistrationSerializer to avoid circular imports
```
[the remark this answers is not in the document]
jorilindell: This worries me as well. 

The circular import was caused because `EventSerializer` uses `RegistrationSerializer` for `Registration` and `RegistrationSerializer` uses `EventSerializer` for `Event`. Circular imports happen when they are in different files.

I tried to move LinkedEventsSerializer to separate file but end result was that I had to move basically all the Event related serializer + `RegistrationSerializer`  as well. One of the problematic lines was `self.related_serializer = globals().get(self.related_serializer, None)` in `JSONLDRelatedField` which is used to get serializer if it's string. It expects Serializers to be in same file. 

I tried to improve this little bit be adding `RegistrationBaseSerializer` in `registration/serializers` and inherit it for the `RegistrationSerializer` which is now only serializer for Registration. `RegistrationBaseSerializer` container all the fields except event and `RegistrationSerializer` contains event field, methods to mutate Registration and validation. It's not perfect but at least after that we don't have to maintain 2 registration serializers.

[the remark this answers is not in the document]
danipran: This will do nicely, thank you!

registrations/serializers.py
danipran: On further consideration, probably fine to leave it as it is; no point in letting this block anything. Relying on a specific string in an error message usually seems a bit flaky to me, and this looks like something that Django could offer tools for... but on the other hand, I can't come up with any suggestions from the top of my head, and this should work just fine.

So feel free to resolve this thread, but if you have, by any chance, looked into this already, please do share before resolving.

jorilindell: Unfortunately I don't have better idea at this point. I don't like relaying to string either, but didn't find any better way to do it

events/api.py
```diff
@@ -1950,6 +1946,64 @@ class Meta:
...
+    def create(self, request, *args, **kwargs):
+        try:
+            instance = super().create(request, *args, **kwargs)
+        except IntegrityError as error:
+            if "duplicate key value violates unique constraint" in str(error):
+                raise serializers.ValidationError(
+                    {"event": _("Event already has a registration.")}
+                )
+            else:
+                raise error
```
danipran, line 1999: You can re-raise an exception just with a plain `raise` inside an `except` block.
```suggestion
                raise
```

danipran (approved):
Some tiny things I noticed on second round, but LGTM otherwise 👍

charn (changes requested):

events/api.py
jorilindell: Changed

PR #574, opened 2023-04-21
LINK-1289 | Add registration model to Django admin

jorilindell:
## Description
Add registration model to Django admin

## Closes
[LINK-1289](https://helsinkisolutionoffice.atlassian.net/browse/LINK-1289)

## Additional info
This PR and https://github.com/City-of-Helsinki/linkedevents/pull/575 will replace https://github.com/City-of-Helsinki/linkedevents/pull/565 which contains several separate features

[LINK-1289]: https://helsinkisolutionoffice.atlassian.net/browse/LINK-1289?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #565, opened 2023-04-06
LINK-1187, LINK-1043 | Validate mandatory registration fields

jorilindell:
## Description
- Endpoint to get mandatory registration field options: http://linkedevents-backend:8080/v1/registration-mandatory-field/
- API documentation for the mandatory registration field endpoint
- Add mandatory_fields field to registration
- Add registrations and mandatory fields to Django admin
- Move TranslatedModelSerializer and EnumChoiceField to separate files to avoid circular imports
- Validate that selected mandatory fields are filled when adding a new signup

## Closes
[LINK-1187](https://helsinkisolutionoffice.atlassian.net/browse/LINK-1187)
[LINK-1043](https://helsinkisolutionoffice.atlassian.net/browse/LINK-1043)

[LINK-1187]: https://helsinkisolutionoffice.atlassian.net/browse/LINK-1187?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[LINK-1043]: https://helsinkisolutionoffice.atlassian.net/browse/LINK-1043?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #575, opened 2023-04-21
LINK-1043, LINK-1187 | Registration mandatory fields

jorilindell:
## Description
- Save mandatory fields when mutating registration
- Validate that all the mandatory fields are filled when mutating signup
- Fix verbose names of registration fields
- Allow null values for name, city, extra_info and membership_number fields of SignUp model

## Closes
[LINK-1187](https://helsinkisolutionoffice.atlassian.net/browse/LINK-1187)
[LINK-1043](https://helsinkisolutionoffice.atlassian.net/browse/LINK-1043)

## Additional info
This PR and https://github.com/City-of-Helsinki/linkedevents/pull/574 will replace https://github.com/City-of-Helsinki/linkedevents/pull/565 which contains several separate features

[LINK-1187]: https://helsinkisolutionoffice.atlassian.net/browse/LINK-1187?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LINK-1043]: https://helsinkisolutionoffice.atlassian.net/browse/LINK-1043?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


registrations/migrations/0012_mandatory_fields_model.py
```diff
@@ -0,0 +1,117 @@
+# Generated by Django 3.2.18 on 2023-04-21 05:31
+
+from django.db import migrations, models
+
+from registrations.models import MandatoryField
+
+
+def create_initial_mandatory_fields(apps, schema_editor):
+    MandatoryField.objects.create(
```
charn, line 9: In migrations you shouldn't import the model directly, but use a historical version instead, because later on the MandatoryFields model might be different than what is expected by the state of the DB in this moment.

So:

```python
MandatoryField = apps.get_model("registration", "MandatoryField")
MandatoryField.objects.create(...)
...
```

registrations/models.py
```diff
@@ -52,25 +78,33 @@ class Registration(models.Model):
...
     minimum_attendee_capacity = models.PositiveSmallIntegerField(
         verbose_name=_("Minimum attendee capacity"), null=True, blank=True
     )
+
     waiting_list_capacity = models.PositiveSmallIntegerField(
-        verbose_name=_("Minimum attendee capacity"), null=True, blank=True
+        verbose_name=_("Waiting list capacity"), null=True, blank=True
+    )
+
+    mandatory_fields = models.ManyToManyField(
```
charn, line 106: Something that I was thinking about was that do we need to have `MandatoryField` as a model to enable us to define new fields dynamically? To me it seems that mostly we're trying to list what fields from `SignUp` and mandatory for this registration instance. Or do we need to be able to define a field `foobar` or some other field that doesn't exist on `SignUp` model? `DefaultMandatoryField` seems to resemble actual fields on `SignUp` model.

How would you feel about something like this? (Note, this is a dry-coded example)

```python
from django.contrib.postgres.fields import ArrayField

class MandatoryFields(models.TextChoices):
    """Choices for mandatory fields on SignUp model."""
    CITY = "city", _("City")
    NAME = "name", _("Name")
    PHONE_NUMBER = "phone_number", _("Phone number")
    STREET_ADDRESS = "street_address",  _("Street address")
    ZIPCODE = "zipcode", _("ZIP code")


class Registration(models.Model):

    mandatory_fields = ArrayField(
        models.CharField(
            max_length=16,
            choices=MandatoryFields.choices,
            blank=True,
        ),
        default=list,
    )
```

jorilindell: This is really good idea and makes also front-end implementation much simpler. Updated the PR and added mandatory fields to registration form in Django admin as well.

registrations/models.py
```diff
@@ -100,65 +158,87 @@ class NotificationType:
     registration = models.ForeignKey(
         Registration, on_delete=models.CASCADE, related_name="signups"
     )
-    name = models.CharField(verbose_name=_("Name"), max_length=50)
+
+    name = models.CharField(
+        verbose_name=_("Name"), max_length=50, blank=True, null=True, default=None
```
charn, line 163: I guess that this field might be sent as {"name": null} which is the reason for having `null=True` here?

I wonder if there would be another way for this. [In Django documentation](https://docs.djangoproject.com/en/4.2/ref/models/fields/#null) using `null=True` on `CharField` and `TextField` is discouraged. By allowing this field to be `NULL` there's now two different values for "no data" `""` and `NULL`.

And I see you've refactored other Char/TextFields as well to take in null, but I wonder if `blank=True, default=""` would also work? Can we use that instead of `null=True, default=None`?

registrations/serializers.py
```diff
@@ -98,7 +119,9 @@ def get_signups(self, obj):
                 queryset = SignUp.objects.filter(registration__id=obj.id)
                 return SignUpSerializer(queryset, many=True, read_only=True).data
             else:
-                return f"Only the admins of the organization that published the event {event.id} have access rights."
+                return _(
+                    f"Only the admins of the organization that published the event {event.id} have access rights."
+                )
```
tuomas777, line 124: Unfortunately gettext does not support f-strings, so this needs to be something like
```python
return _(
    "Only the admins of the organization that published the event {event_id} have access rights."
).format(event_id=event.id)
```

registrations/models.py
jorilindell: The reason for `null=True` was to support {"name": null} in payload as you expected.

We can send empty string from the UI as well to get rid of `null=True, default=None` or remove `name` from the payload. Changed as suggested

tuomas777: Discussed this a bit with @charn. After studying the already existing models and API, it unfortunately seems that `null=True` is very dominant among the `CharField`s and `TextField`s in LE. Therefore it is actually probably the less bad option to make these registration fields nullable as well to keep things consistent. 😅

jorilindell: Ok 😅. Changed SignUp model fields nullable.

registrations/serializers.py
```diff
@@ -59,7 +40,47 @@ def create(self, validated_data):
...
+            val = data.get(field)
+            if not val:
+                errors[field] = _("This field must be specified.")
+
+        if registration.audience_min_age or registration.audience_max_age:
+            if "date_of_birth" not in data.keys():
+                errors["date_of_birth"] = _("This field must be specified.")
+            else:
+                date_of_birth = data["date_of_birth"]
+                today = date.today()
```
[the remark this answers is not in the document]
jorilindell: Changed.

registrations/serializers.py
jorilindell: Changed. f-string was used in gettext in other place too. Fixed there as well.

registrations/serializers.py
```diff
@@ -59,7 +40,47 @@ def create(self, validated_data):
...
+    def validate(self, data):
+        errors = {}
+        registration = data["registration"]
+
+        for field in registration.mandatory_fields:
+            val = data.get(field)
+            if not val:
+                errors[field] = _("This field must be specified.")
+
+        if registration.audience_min_age or registration.audience_max_age:
```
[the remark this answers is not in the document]
jorilindell: I made date_of_birth required field also if audience_min_age or audience_max_age is 0. I added also unit tests for both cases.

registrations/serializers.py
```diff
@@ -59,7 +41,50 @@ def create(self, validated_data):
             signup.save()
             return signup
         else:
-            raise DRFPermissionDenied("The waiting list is already full")
+            raise DRFPermissionDenied(_("The waiting list is already full"))
```
charn, line 44: No need to change this. Just thinking if this should be also turned into a validation error in the future, so that a proper error could be rendered by the UI more easily. :thinking:

charn (approved):
LGTM! :+1: :shipit:

registrations/serializers.py
jorilindell: That's definitely something the could be done later.

PR #579, opened 2023-04-25
Fix if-clause in an itegrity exception handler

charn:


danipran (approved):
Aside from a typo in the commit message, LGTM :+1:

PR #582, opened 2023-04-28
LINK-1271 | Endpoint to send message to all or selected registration signups

jorilindell:
## Description

- Endpoint to send an email message to all or selected signups.
- Email uses Linked Registration email template for styling
- Email subject and body are required fields.
- User can send email to all signups of the registration by leaving signups attribute empty or send the email to selected signups

## Closes
[LINK-1271](https://helsinkisolutionoffice.atlassian.net/browse/LINK-1271)

## Example request:
POST http://linkedevents-backend:8080/v1/registration/2/send_message/
```
{
    "body":"<p>Message1</p><p>Message2</p>",
    "subject": "MESSAGE SUBJECT",
    "signups": [2]
}
```

## Additional note
PR for FE: https://github.com/City-of-Helsinki/linkedcomponents-ui/pull/180

[LINK-1271]: https://helsinkisolutionoffice.atlassian.net/browse/LINK-1271?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/registrations/api.py
+++ b/registrations/api.py
@@ -1,11 +1,16 @@
+import logging
 from datetime import datetime, timedelta
+from smtplib import SMTPException
 from uuid import UUID
 
+import bleach
 import pytz
 from django.conf import settings
 from django.contrib.auth.models import AnonymousUser
 from django.contrib.gis.db import models
+from django.core.mail import EmailMultiAlternatives, get_connection
 from django.db.models import ProtectedError, Q, Sum
+from django.template.loader import render_to_string
 from django.utils.translation import gettext as _
 from rest_framework import mixins, serializers, status, viewsets
 from rest_framework.decorators import action
@@ -32,6 +37,8 @@ from registrations.exceptions import ConflictException
 from registrations.models import Registration, SeatReservationCode, SignUp
 from registrations.serializers import SeatReservationCodeSerializer, SignUpSerializer
 from registrations.utils import code_validity_duration
+
+logger = logging.getLogger(__name__)
 
 
 class SignUpViewSet(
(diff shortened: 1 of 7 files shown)


registrations/api.py
```diff
@@ -254,6 +261,117 @@ def none_to_unlim(val):
...
+            )
+            messages = []
+            for subject, message, html_message, from_email, recipient_list in datatuple:
+                mail = EmailMultiAlternatives(
+                    subject, message, from_email, recipient_list, connection=connection
+                )
+                mail.attach_alternative(html_message, "text/html")
+                messages.append(mail)
+
+            return connection.send_messages(messages)
```
charn, line 296: I'd move this into `registrations.utils`

registrations/api.py
```diff
@@ -254,6 +261,117 @@ def none_to_unlim(val):
...
+                rendered_body,
+                settings.SUPPORT_EMAIL,
+                [signup.email],
+            )
+            messages.append(message)
+
+        try:
+            send_mass_html_mail(messages, fail_silently=False)
+        except SMTPException as e:
+            logger.error(e, exc_info=True)
```
charn, line 359: https://docs.python.org/3/library/logging.html#logging.Logger.exception

I would do `logger.exception("Couldn't send mass HTML email.")` here.

registrations/api.py
jorilindell: Moved send_mass_html_mail to `registrations.utils`

registrations/api.py
```diff
@@ -254,6 +260,90 @@ def none_to_unlim(val):
...
+            except ValueError as error:
+                errors["signups"] = str(error)
+
+        else:
+            signups = registration.signups.all()
+        signups.exclude(email=None)
+
+        # Return validations errors if there is any
+        if errors:
+            raise serializers.ValidationError(errors)
```
danipran, line 297: Data validation should be something that a serializer does, not a view.

danipran (changes requested):

registrations/api.py
charn: More of an idea rather than working code, but instead of doing what a serializer does you could do something like:

```
# Add a serializer class for the email
class MassEmailSerializer(serializers.Serializer):
    subject = serializers.CharField(required=True)
    body = serializers.CharField(required=True)
    signups = serializers.ListField(child=serializers.CharField(), allow_empty=False, min_length=1,)


# Add this to the viewset to return a correct serializer for the send_message action 
def get_serializer_class(
    if self.action == 'send_message':
        return MassEmailSerializer
    return super().get_serializer_class()


# in this action method you'd then do something like

serializer = self.get_serializer(data=request.data)
serializer.is_valid(raise_exception=True)

# Get values from serializer.validated_data
body = serializer.validated_data["body"]
```

jorilindell: Thanks guys. I added MassEmailSerializer serializer to validate send_message action payload as suggested

registrations/serializers.py
```diff
@@ -164,3 +164,28 @@ class Meta:
...
     def get_expiration(self, obj):
         return obj.timestamp + timedelta(minutes=code_validity_duration(obj.seats))
+
+
+class MassEmailSerializer(serializers.Serializer):
+    subject = serializers.CharField()
+    body = serializers.CharField()
+    signups = serializers.ListField(
+        child=serializers.CharField(), allow_empty=False, required=False
+    )
```
danipran, line 174: This is a list of IDs, no? Could use `PrimaryKeyRelatedField` instead: https://www.django-rest-framework.org/api-guide/relations/#primarykeyrelatedfield

Or more concretely, something like this:
```python
    signups = serializers.PrimaryKeyRelatedField(
        queryset=SignUp.objects.all().exclude(email=None), many=True, required=False,
    )
```

With this, you can also get rid of `validate_signups(...)`. You'd also need to refactor your view a bit, replacing lines 285 to 294 with:
```python
signups = serializer.validated_data.get("signups", registration.signups.exclude(email=None))
```

registrations/tests/test_send_message.py
```diff
@@ -0,0 +1,247 @@
...
+
+@pytest.mark.django_db
+def test_send_message_to_selected_signups(
+    api_client, registration, signup, signup2, user
+):
+    api_client.force_authenticate(user)
+    send_message_data = {
+        "subject": "Message subject",
+        "body": "Message body",
+        "signups": [signup.id],
```
danipran, line 79: Might make sense to have multiple IDs here. Or separate tests/parameters for selected signups and a selected signup.

registrations/templates/message_to_signup.html
```diff
@@ -0,0 +1,20 @@
...
+{% block content %}
+  <tr>
+    <td class="content-cell" style="padding:0;margin:0;text-align:left;">
+      <div class="content-container" style="padding:0 16px;">
+        <div style="margin: 0px 0px 40px">
+          <p style="color:#1a1a1a;font-size:16px;font-weight:normal;line-height:24px;margin:0px;">{{body |safe}}</p>
+        </div>
+        <div class="cta-button-wrapper" style="margin: 0px 0px 40px">
+          <a href="{{linked_registrations_ui_url}}/fi/registration/{{registration_id}}/enrolment/{{signup_id}}/{{cancellation_code}}/edit" target="_blank" rel="noopener" class="cta-button" style="background:#0000bf;border-radius:0px;border:none;height:56px;text-align:center;text-decoration:none;vertical-align:middle;padding:16px 24px;box-sizing:border-box;display:inline-block;">
+            <span style="color:#ffffff;font-size:16px;font-weight:500;line-height:24px;">{% blocktranslate %}Tarkastele ilmoittautumistasi täällä{% endblocktranslate %}</span>
```
danipran, line 13: Does this actually get translated? 🤔 so if I send this email to a Swedish user and a Finnish user, do they get the email in their own language? Or is this outside of the scope of this PR?

danipran (commented):
Well done with the serializer! I've some questions and one change suggestion.

registrations/api.py
```diff
@@ -254,6 +269,78 @@ def none_to_unlim(val):
...
+            logger.exception("Couldn't send mass HTML email.")
+
+            return Response(
+                str(e),
+                status=status.HTTP_409_CONFLICT,
+            )
+
+        return Response(
+            {
+                "emails": [su.email for su in signups],
```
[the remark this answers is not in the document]
jorilindell: UI doesn't use it for anything. Let's remove emails from the response as it feels little bit risky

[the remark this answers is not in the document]
danipran: Could also return a list of signup IDs if you see a need for some kind of confirmation.

registrations/templates/message_to_signup.html
jorilindell: There are quite many missing translations for registration in general. This could be solved in separate PR.

registrations/tests/test_send_message.py
jorilindell: Added another signup `[signup.id, signup3.id]` to this test to make sure that email is sent to multiple participants

registrations/api.py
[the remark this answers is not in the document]
jorilindell: Changed `emails` to list of signup IDs

registrations/serializers.py
jorilindell: Changed as suggested. I also added MassEmailSignupsField class to have custom get_queryset method. Otherwise user could send emails to signups of other registrations
```
class MassEmailSignupsField(serializers.PrimaryKeyRelatedField):
    def get_queryset(self):
        registration = self.context["request"].data["registration"]

        return registration.signups.exclude(email=None)


class MassEmailSerializer(serializers.Serializer):
    subject = serializers.CharField()
    body = serializers.CharField()
    signups = MassEmailSignupsField(
        many=True,
        required=False,
    )
  ```

Comment (danipran) [reply registrations/serializers.py]: Ah... very true. Good catch!

File: registrations/api.py [hunk 8]
```diff
@@ -254,6 +269,78 @@ def none_to_unlim(val):
...
+            messages.append(message)
+
+        try:
+            send_mass_html_mail(messages, fail_silently=False)
+        except SMTPException as e:
+            logger.exception("Couldn't send mass HTML email.")
+
+            return Response(
+                str(e),
+                status=status.HTTP_409_CONFLICT,
```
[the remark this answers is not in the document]
danipran: Had a short discussion about this in Slack. While 409 isn't technically correct, it's fine for now.

danipran (approved):
LGTM :+1:

PR #584, opened 2023-05-03
LINK-1274 | Local changes from current production environment

charn:
This PR contains local changes which are present in a diff from the current production environment (and some minor refactoring). These changes need to be checked if they are still relevant.

### DIFF
--- a/events/api.py
+++ b/events/api.py
@@ -2500,15 +2500,21 @@ def parse_duration_string(duration):
     return int(val) * mul
 
 
-def _terms_to_regex(terms, operator, fuzziness=3):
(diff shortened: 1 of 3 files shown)


danipran (approved):
Although I have no idea why these changes are done, the changes themselves look fine.

The changes in `local_ongoing_text` filter raises some question marks; why are we fetching non-local events here? What even is this parameter? It's not documented, and I can't find any other references to it here or anywhere within the City of Helsinki organization.

Regardless, these are in production use, sooo... better to just have them in the repository as well? ¯\\\_(ツ)\_/¯

PR #590, opened 2023-05-11
LINK-1172 | Allow searching multiple event statuses in API

tonipel:
# Description

Allows passing multiple parameters to `event_status` -field separated by a comma. Returns events that match any of the given statuses. i.e. `?event_status=EventCancelled,EventRescheduled`

## Closes
[LINK-1172](https://helsinkisolutionoffice.atlassian.net/browse/LINK-1172)

[LINK-1172]: https://helsinkisolutionoffice.atlassian.net/browse/LINK-1172?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/events/api.py
+++ b/events/api.py
@@ -3026,15 +3026,19 @@ def _filter_event_queryset(queryset, params, srs=None):  # noqa: C901
         queryset = queryset.filter(publication_status=PublicationStatus.PUBLIC)
 
     # Filter by event status
-    val = params.get("event_status", None)
-    if val and val.lower() == "eventscheduled":
-        queryset = queryset.filter(event_status=Event.Status.SCHEDULED)
-    elif val and val.lower() == "eventrescheduled":
-        queryset = queryset.filter(event_status=Event.Status.RESCHEDULED)
-    elif val and val.lower() == "eventcancelled":
-        queryset = queryset.filter(event_status=Event.Status.CANCELLED)
-    elif val and val.lower() == "eventpostponed":
-        queryset = queryset.filter(event_status=Event.Status.POSTPONED)
+    if status_param := params.get("event_status"):
+        statuses = []
+        for value in (s.lower() for s in status_param.split(",")):
+            if value == "eventscheduled":
+                statuses.append(Event.Status.SCHEDULED)
+            elif value == "eventrescheduled":
+                statuses.append(Event.Status.RESCHEDULED)
+            elif value == "eventcancelled":
+                statuses.append(Event.Status.CANCELLED)
+            elif value == "eventpostponed":
+                statuses.append(Event.Status.POSTPONED)
+
+        queryset = queryset.filter(event_status__in=statuses)
 
     # Filter by language, checking both string content and in_language field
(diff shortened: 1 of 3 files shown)


events/api.py
```diff
@@ -3025,15 +3025,21 @@ def _filter_event_queryset(queryset, params, srs=None):  # noqa: C901
...
+            if val and val.lower() == "eventscheduled":
+                searched_statuses.append(Event.Status.SCHEDULED)
+            elif val and val.lower() == "eventrescheduled":
+                searched_statuses.append(Event.Status.RESCHEDULED)
+            elif val and val.lower() == "eventcancelled":
+                searched_statuses.append(Event.Status.CANCELLED)
+            elif val and val.lower() == "eventpostponed":
+                searched_statuses.append(Event.Status.POSTPONED)
+
+        queryset = queryset.filter(event_status__in=searched_statuses)
```
tuomas777, line 3042: Something like
```
if status_param := params.get("event_status"):
    statuses = []
    for value in (s.lower() for s in status_param.split(",")):
        if value == "eventscheduled":
            statuses.append(Event.Status.SCHEDULED)
        elif value == "eventrescheduled":
            statuses.append(Event.Status.RESCHEDULED)
        elif value == "eventcancelled":
            statuses.append(Event.Status.CANCELLED)
        elif value == "eventpostponed":
            statuses.append(Event.Status.POSTPONED)

    queryset = queryset.filter(event_status__in=statuses)
```
could be a bit more pythonic and DRY.

events/tests/test_event_get.py
```diff
@@ -578,6 +578,10 @@ def test_event_status_filter(
     get_list_and_assert_events("event_status=eventrescheduled", [event2])
     get_list_and_assert_events("event_status=eventcancelled", [event3])
     get_list_and_assert_events("event_status=eventpostponed", [event4])
+    get_list_and_assert_events(
+        "event_status=eventscheduled,eventrescheduled,eventcancelled,eventpostponed",
+        [event, event2, event3, event4],
+    )
```
tuomas777, line 584: In addition to this, I guess we could also test that multiple but not all work (like 2 statuses), as it is trivial to add that test.

tuomas777 (commented):
Would be nice to also have this new feature mentioned in the "documentation" [around here](https://github.com/City-of-Helsinki/linkedevents/blob/092fc2a46a62873d7718ce700fbdd31846eb67a7/helevents/templates/rest_framework/event_list.html#L166) (yes, there is no way you could have known that 😄)

events/api.py
tonipel: Added note to the documentation and changed code as suggested.

events/tests/test_event_get.py
tonipel: Added new test for this case.

tuomas777 (approved):
Good job! 👍

PR #591, opened 2023-05-13
LINK-1139 | Saving data for external user

charn:
This PR adds new fields for event. Some fields are considered just extra information for an event and some fields contain personal information of an external user sending the event. New fields which are considered personal information of the external user should be hidden from the public API.

External users have more strict validation rules.


tuomas777 (approved):
Looking good! 👍 🐝

PR #600, opened 2023-05-25
LINK-1327, LINK-1328, LINK-1030 | Endpoint to update existing seats reservation

jorilindell:
## Description
- Endpoint to update existing seats reservation
- Move endpoints to create/update seats reservation to a separate route
- Allow to reserve seats to waiting list only when all seats in event are filled

## Closes
[LINK-1327](https://helsinkisolutionoffice.atlassian.net/browse/LINK-1327)
[LINK-1328](https://helsinkisolutionoffice.atlassian.net/browse/LINK-1328)
[LINK-1030](https://helsinkisolutionoffice.atlassian.net/browse/LINK-1030)

[LINK-1327]: https://helsinkisolutionoffice.atlassian.net/browse/LINK-1327?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LINK-1328]: https://helsinkisolutionoffice.atlassian.net/browse/LINK-1328?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LINK-1030]: https://helsinkisolutionoffice.atlassian.net/browse/LINK-1030?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


registrations/serializers.py
```diff
@@ -274,14 +273,131 @@ def validate(self, data):
...
+                _("This field must be specified."), code="required"
+            )
+
+        if isinstance(instance, SeatReservationCode):
+            non_editable_fields = ["registration", "timestamp"]
+
+            # It's not allowed to change registration or timestamp values
+            for field in non_editable_fields:
+                if field in data and getattr(instance, field) != data[field]:
+                    errors[field] = _("You may not change the value.")
```
danipran, line 314: While you can make `timestamp` read-only, `registration` seems to be something that needs to be "create-only", no? Instead of this, you could override `to_internal_value` something like this:
```python
    def to_internal_value(self, data):
        if self.instance:
            data["registration"] = self.instance.registration.id
        return super().to_internal_value(data)
```
This way, it doesn't really matter what the value of the field is or if it even exists during update, it always gets replaced by the reservation's registration.

registrations/serializers.py
```diff
@@ -274,14 +273,131 @@ def validate(self, data):
...
+        errors = {}
+
+        # Amount of seats is always required
+        if data.get("seats") is None:
+            errors["seats"] = ErrorDetail(
+                _("This field must be specified."), code="required"
+            )
+
+        if isinstance(instance, SeatReservationCode):
+            non_editable_fields = ["registration", "timestamp"]
```
charn, line 309: You can define read-only fields in the serializer meta https://www.django-rest-framework.org/api-guide/serializers/#specifying-read-only-fields

charn: Ah, but I guess you need `registration` here to be able to create an seat reservation. :thinking: One option probably would be to have `validate_registration` or define `registration` dynamically as readonly in `__init__` :thinking:

registrations/serializers.py
```diff
@@ -274,14 +273,131 @@ def validate(self, data):
...
+        # Validate attendee capacity only if maximum_attendee_capacity is defined
+        if maximum_attendee_capacity is not None:
+            serialized_registration = RegistrationBaseSerializer(
+                data["registration"], context=self.context
+            ).data
+            attendee_count = serialized_registration["current_attendee_count"]
+            reserved_seats_amount = registration.reserved_seats_amount
+
+            if isinstance(instance, SeatReservationCode):
+                reserved_seats_amount = max(reserved_seats_amount - instance.seats, 0)
```
danipran, line 337:
```suggestion
            if instance:
                reserved_seats_amount = max(reserved_seats_amount - instance.seats, 0)
            else:
                reserved_seats_amount = registration.reserved_seats_amount
```

registrations/serializers.py
```diff
@@ -274,14 +273,131 @@ def validate(self, data):
...
+                        capacity_left=max(
+                            waiting_list_capacity_left - reserved_seats_amount, 0
+                        )
+                    )
+
+        if errors:
+            raise serializers.ValidationError(errors)
+
+        super().validate(data)
+        return data
```
danipran, line 381:
```suggestion
        return super().validate(data)
```

registrations/serializers.py
```diff
@@ -274,14 +273,131 @@ def validate(self, data):
...
+
+    def update(self, instance, validated_data):
+        if localtime() > instance.expiration:
+            raise ConflictException(_("Cannot update expired seats reservation."))
+
+        super().update(instance, validated_data)
+        return instance
```
danipran, line 388:
```suggestion
        return super().update(instance, validated_data)
```

registrations/tests/test_seatsreservation_put.py
```diff
@@ -0,0 +1,225 @@
...
+@pytest.mark.django_db
+def test_cannot_update_registration(api_client, event, registration, registration2):
+    reservation = SeatReservationCode.objects.create(seats=1, registration=registration)
+
+    reservation_data = {
+        "seats": 1,
+        "registration": registration.id,
+        "code": reservation.code,
+        "registration": registration2.id,
+    }
```
danipran, line 108:
```suggestion
    reservation_data = {
        "seats": 1,
        "code": reservation.code,
        "registration": registration2.id,
    }
```

registrations/serializers.py
```diff
@@ -274,14 +273,131 @@ def validate(self, data):
...
+
+        if maximum_attendee_capacity is not None:
+            attendee_count = registration.signups.filter(
+                attendee_status=SignUp.AttendeeStatus.ATTENDING
+            ).count()
+
+            if maximum_attendee_capacity - attendee_count <= 0:
+                return True
+
+        return False
```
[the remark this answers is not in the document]
jorilindell: Changed as suggested. I think your suggestion is more clear

registrations/serializers.py
```diff
@@ -274,14 +273,131 @@ def validate(self, data):
...
+        instance = self.instance
+        errors = {}
+
+        # Amount of seats is always required
+        if data.get("seats") is None:
+            errors["seats"] = ErrorDetail(
+                _("This field must be specified."), code="required"
+            )
+
+        if isinstance(instance, SeatReservationCode):
```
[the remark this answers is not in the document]
jorilindell: Changed as suggested

registrations/serializers.py
```diff
@@ -274,14 +273,131 @@ def validate(self, data):
...
+        registration = data["registration"]
+        maximum_attendee_capacity = registration.maximum_attendee_capacity
+        waiting_list_capacity = registration.waiting_list_capacity
+
+        # Validate attendee capacity only if maximum_attendee_capacity is defined
+        if maximum_attendee_capacity is not None:
+            serialized_registration = RegistrationBaseSerializer(
+                data["registration"], context=self.context
+            ).data
+            attendee_count = serialized_registration["current_attendee_count"]
```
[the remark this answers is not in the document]
jorilindell: Very good point. Added `current_attendee_count`, `current_waiting_list_count`, `remaining_attendee_capacity` and `remaining_waiting_list_capacity` properties to Registration model instead of having similar methods in RegistrationSerializer

registrations/serializers.py
```diff
@@ -274,14 +273,131 @@ def validate(self, data):
...
+                if field in data and getattr(instance, field) != data[field]:
+                    errors[field] = _("You may not change the value.")
+
+            # The code must be defined and match instance code when updating existing seats reservation
+            if self.initial_data.get("code") is None:
+                errors["code"] = ErrorDetail(
+                    _("This field must be specified."), code="required"
+                )
+            elif str(instance.code) != self.initial_data["code"]:
+                errors["code"] = _("The value doesn't match.")
```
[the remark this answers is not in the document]
jorilindell: Good point. Changed so that it returns validations error if code is invalid

registrations/serializers.py
```diff
@@ -274,14 +273,131 @@ def validate(self, data):
 
 class SeatReservationCodeSerializer(serializers.ModelSerializer):
     timestamp = DateTimeField(default_timezone=pytz.UTC, required=False)
```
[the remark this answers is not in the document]
jorilindell: Changed as suggested

registrations/serializers.py
jorilindell: Changed as suggested

registrations/serializers.py
jorilindell: Changed as suggested

registrations/serializers.py
jorilindell: I ended up to have timestamp as read_only field and preventing to change registration in to_internal_value method

```
timestamp = DateTimeField(default_timezone=pytz.UTC, required=False, read_only=True)
```

```
def to_internal_value(self, data):
        if self.instance:
            data["registration"] = self.instance.registration.id
        return super().to_internal_value(data)
```

registrations/serializers.py
```diff
@@ -273,15 +238,118 @@ def validate(self, data):
...
+        attendee_count = registration.signups.filter(
+            attendee_status=SignUp.AttendeeStatus.ATTENDING
+        ).count()
+
+        return maximum_attendee_capacity - attendee_count <= 0
+
+    def to_internal_value(self, data):
+        if self.instance:
+            data["registration"] = self.instance.registration.id
+        return super().to_internal_value(data)
```
charn, line 268: `to_intermal_value` is responsible for deserialization. So in that sense this would be better done elsewhere.

I think `__init__` might work the best.

```
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)

        if self.instance:
            self.fields["registration"].read_only = True
```

danipran: Personally, I wouldn't touch `__init__` in cases like these 😅 I think `to_internal_value` makes sense exactly because it's responsible for deserialization (i.e. write operations) *and* validation. It's the method that's used to return the values needed for building the instance.

danipran: And FWIW it was something I suggested: https://github.com/City-of-Helsinki/linkedevents/pull/600#discussion_r1213055315

danipran (approved):
LGTM, but need to resolve [Juha's review](https://github.com/City-of-Helsinki/linkedevents/pull/600#discussion_r1213664268) first.

registrations/serializers.py
jorilindell: Other options could be:
Do it in `validate_registration` method as suggested in https://github.com/City-of-Helsinki/linkedevents/pull/600#discussion_r1213088698
e.g.
```
def validate_registration(self, value):
        instance = self.instance
        if instance and instance.registration != value:
            raise serializers.ValidationError(
                {
                    _( "You may not change the registration of an existing object.")
                }
            )
        return value
```

or do in in update method as similar validation is done in SignUpSerializer and KeywordSetSerializer
e.g. 
```
def update(self, instance, validated_data):
        if (
            "registration" in validated_data
            and instance.registration != validated_data["registration"]
        ):
            raise serializers.ValidationError(
                {
                    "registration": _(
                        "You may not change the registration of an existing object."
                    )
                }
            )
            ...
```

charn: My reasoning for doing it in the `__init__` was that `self.fields["registration"].read_only = True` is almost self documenting way to do it. I also like `validate_registration` since there you can easily raise the validation error and API user is not left wondering why his update is not updating the value. :shrug:

charn: I can accept `to_internal_value` if a comment is added.

danipran: Now that I think of it, the `__init__` option wouldn't work; setting the field to read-only during init would make the serializer skip the field completely during an update, i.e. it's not included in the data.

danipran: Nope, the `to_internal_value` implementation doesn't skip the field, it just replaces it and also ensures that the data is there, regardless of what happens. Here's what happens if I remove the `to_internal_value` implementation and use the `__init__` implementation instead:
```
===================================================================================== short test summary info ======================================================================================
FAILED registrations/tests/test_seatsreservation_put.py::test_update_seats_reservation - KeyError: 'registration'
FAILED registrations/tests/test_seatsreservation_put.py::test_seats_amount_has_not_limit_if_maximum_attendee_capacity_is_none - KeyError: 'registration'
FAILED registrations/tests/test_seatsreservation_put.py::test_cannot_update_registration - KeyError: 'registration'
FAILED registrations/tests/test_seatsreservation_put.py::test_cannot_update_expired_reservation - KeyError: 'registration'
FAILED registrations/tests/test_seatsreservation_put.py::test_cannot_update_timestamp - KeyError: 'registration'
FAILED registrations/tests/test_seatsreservation_put.py::test_cannot_reserve_seats_if_there_are_not_enough_seats_available - KeyError: 'registration'
FAILED registrations/tests/test_seatsreservation_put.py::test_update_seats_reservation_in_waiting_list - KeyError: 'registration'
FAILED registrations/tests/test_seatsreservation_put.py::test_waiting_list_seats_amount_has_not_limit_if_waiting_list_capacity_is_none - KeyError: 'registration'
FAILED registrations/tests/test_seatsreservation_put.py::test_cannot_reserve_seats_waiting_list_if_there_are_not_enough_seats_available - KeyError: 'registration'
```

danipran: About the other points; you're correct that my `to_internal_value` suggestion doesn't really tell the API user why the registration doesn't get updated and I think it's a valid point. Personally, I prefer convenience over convention in this case; telling the user that you can't change this value doesn't really bring any value.

charn (approved):

PR #601, opened 2023-05-30
Missing cache error

tonipel:
Add check to not result in an error if cache is missing.
Refactor away duplicate code.

Refs LINK-1319

### DIFF
--- a/events/api.py
+++ b/events/api.py
@@ -1,4 +1,5 @@
 import base64
+import logging
 import re
 import struct
 import time
@@ -108,6 +109,7 @@ from helevents.serializers import UserSerializer
 from linkedevents.registry import register_view, viewset_classes_by_model
 from registrations.serializers import RegistrationBaseSerializer
 
+logger = logging.getLogger(__name__)
 LOCAL_TZ = pytz.timezone(settings.TIME_ZONE)
 
 
@@ -2511,6 +2513,43 @@ def _terms_to_regex(terms, operator, fuzziness=3):
     return regex.compile(expr, regex.IGNORECASE)
 
 
+def _get_queryset_from_cache(params, param, cache_name, operator, queryset):
+    val = params.get(param, None)
+    if not val:
+        return queryset
+
+    cache_values = cache.get(cache_name)
+    if not cache_values:
+        logger.error(f"Missed cache {cache_name}")
+        return queryset
+
+    rc = _terms_to_regex(val, operator)
+    ids = {k for k, v in cache_values.items() if rc.search(v, concurrent=True)}
+    queryset = queryset.filter(id__in=ids)
+
+    return queryset
+
+
+def _get_queryset_from_cache_many(params, param, cache_name, operator, queryset):
+    val = params.get(param, None)
+    if not val:
+        return queryset
+
+    cache_values = cache.get(cache_name)
+    if not cache_values:
+        logger.error(f"Missed cache {cache_name}")
+        return queryset
+
+    rc = _terms_to_regex(val, operator)
+    cached_ids = {k: v for i in cache_values.values() for k, v in i.items()}
+
+    ids = {k for k, v in cached_ids.items() if rc.search(v, concurrent=True)}
+
+    queryset = queryset.filter(id__in=ids)
+
+    return queryset
+
+
 def _filter_event_queryset(queryset, params, srs=None):  # noqa: C901
     """
     Filter events queryset by params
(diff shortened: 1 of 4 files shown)


events/api.py
```diff
@@ -2578,76 +2578,64 @@ def _filter_event_queryset(queryset, params, srs=None):  # noqa: C901
...
+    def get_queryset_from_cache(params, param, cache_name, operator, queryset):
+        cache_values = cache.get(cache_name)
+        if cache_values:
+            val = params.get(param, None)
+            if val:
+                rc = _terms_to_regex(val, operator)
+                ids = {
+                    k for k, v in cache_values.items() if rc.search(v, concurrent=True)
+                }
+                queryset = queryset.filter(id__in=ids)
```
danipran, line 2590: Could do this to reduce nesting:
```suggestion
    def get_queryset_from_cache(params, param, cache_name, operator, queryset):
        cache_values = cache.get(cache_name)
        val = params.get(param, None)
        if not val or not cache_values:
            return queryset

        rc = _terms_to_regex(val, operator)
        ids = {
             k for k, v in cache_values.items() if rc.search(v, concurrent=True)
        }
        return queryset.filter(id__in=ids)
```
Makes it a bit easier to read. Same for the other function, `get_queryset_from_cache_many`.

events/api.py
```diff
@@ -2578,76 +2578,64 @@ def _filter_event_queryset(queryset, params, srs=None):  # noqa: C901
...
-    val = params.get("local_ongoing_OR", None)
-    if val:
-        rc = _terms_to_regex(val, "OR")
-        ids = {
-            k
-            for k, v in cache.get("local_ids").items()
-            if rc.search(v, concurrent=True)
-        }
-        queryset = queryset.filter(id__in=ids)
+    def get_queryset_from_cache(params, param, cache_name, operator, queryset):
```
danipran, line 2581: Any reason for this and `get_queryset_from_cache_many` being located in the middle of the function? 🤔 Personally, I'd prefer them being in the beginning of the function. Even better if they're outside the function, since then you can write unit tests for them.

danipran (changes requested):
Definitely better than before!

events/api.py
tonipel: Changed as suggested.

events/api.py
tonipel: Not other than that they're local functions. But moved them away from the parent function. Any tips how to use the cache in tests? Didn't find any examples using it.

danipran: Of course, I'm not quite sure if changing the cache back-end affects other tests. I sure hope not, but if it does, then mocking might make more sense.

danipran: Or maybe a fixture like this...
```python
from django.core.cache import cache

@pytest.fixture(autouse=True)
def teardown_cache():
    yield
    cache.clear()
```
Then you could just use Django's cache instead of requesting a fixture. Should be safer too. Assuming that caches aren't automatically cleared between tests already.

tonipel: Thanks for the good tips! Added tests for those new functions.

danipran (approved):
LGTM :+1:

events/api.py
```diff
@@ -2511,6 +2511,37 @@ def _terms_to_regex(terms, operator, fuzziness=3):
...
+    val = params.get(param, None)
+
+    if not val or not cache_values:
+        return queryset
+
+    rc = _terms_to_regex(val, operator)
+    ids = {k for k, v in cache_values.items() if rc.search(v, concurrent=True)}
+    queryset = queryset.filter(id__in=ids)
+
+    return queryset
```
charn, line 2525: I think checking `val` first and then calling the cache if needed could be a nice micro optimization. Also I think it might be a good idea to maybe send a message to sentry if we miss a cache, since that basically means that some service might be getting a bad result? :thinking: Or at least we'd then have more info when the service admins start asking us why linked events is returning a bad result. Though calling Sentry might be too much if we can expect to get cache misses, but I have a feeling that we shouldn't expect that. 

```suggestion
def _get_queryset_from_cache(params, param, cache_name, operator, queryset):
    val = params.get(param, None)
    if not val:
        return queryset
    
    cache_values = cache.get(cache_name)
    
    if not cache_values:
        send_a_message_to_sentry(f"Missed cache {cache_name}")  # Can't remember what function to actually call from the top of my head. :(
        return queryset

    rc = _terms_to_regex(val, operator)
    ids = {k for k, v in cache_values.items() if rc.search(v, concurrent=True)}
    queryset = queryset.filter(id__in=ids)

    return queryset
```

val = params.get(param, None)

charn: Same applies for the `_get_queryset_from_cache_many`

charn (changes requested):
Added a suggestion for a minor optimization and also a request to send a notification to sentry about a missed cache.

events/api.py
charn: Well, I guess the following should go nicely to sentry:

```python
import logging
logger = logging.getLogger(__name__)

...

logger.error(f"Missed cache {cache_name}")

```

tonipel: Changed as suggested.

charn (approved):
LGTM! :+1: :shipit:

PR #604, opened 2023-06-01
LINK-1164: Allow event creation, modifying and deleting for external users

danipran:
## Description
This PR allows event creation, modifying and deletion for external users, i.e. users that do not belong to an organization. This means that this also adds support for non-organization users. There's also a lot of refactoring involved.

## Related issue(s)
Related issue: [LINK-1164](https://helsinkisolutionoffice.atlassian.net/browse/LINK-1164)

## Change list
(in no order of importance)
- New users are no longer automatically assigned to a default organization
- `publisher` field in event payload is no longer required
  - If missing or null, it gets automatically replaced
  - If creating, it uses the default organization
  - If updating, it uses the target event's organization
- Remove the organization logic from `DataSourceResourceEditPermission`
- New permissions:
  - `IsReadOnly`, check if HTTP method is in safe methods aka read-only methods
  - `UserBelongsToOrganization`, check if user belongs to an organization
  - `IsObjectEditableByOrganizationUser`, check if user has rights to edit an object within their organization
  - `OrganizationUserEditPermission`, a composite of the previous permissions
- Replace view permission classes to `DataSourceResourceEditPermission & OrganizationUserEditPermission`
- Reorganize code to get rid of circular imports and to decouple code
  - `events.permissions.UserModelPermissionsMixin` -> `helevents.models.UserModelPermissionsMixin`
  - `events.api.organization_can_be_edited_by` -> `events.utils.organization_can_be_edited_by`
  - Code from `EventViewSet.user_data_source_and_organization` to `utils.get_user_data_source_and_organization_from_request`
- Rename methods in `UserModelPermissionsMixin` to better reflect their purpose
  - `is_admin(publisher)` -> `is_admin_of(publisher)`
  - `is_regular_user(publisher)` -> `is_regular_user_of(publisher)`
- Add `is_external` property in `UserModelPermissionsMixin` which tells whether the user is an external user or not
- Split `UserModelPermissionsMixin.can_create_event` from `UserModelPermissionsMixin.can_edit_event`
- Fix some of the tests and test data

[LINK-1164]: https://helsinkisolutionoffice.atlassian.net/browse/LINK-1164?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/events/tests/test_event_auth_api.py
+++ b/events/tests/test_event_auth_api.py
@@ -370,17 +370,6 @@ class TestEventAPI(APITestCase):
         response = self.client.get(url)
         self.assertEqual(response.status_code, status.HTTP_404_NOT_FOUND)
 
-    def test_random_user_create_event_denied(self):
-        url = reverse("event-list")
-        location_id = reverse("place-detail", kwargs={"pk": self.place.id})
-        data = self.make_minimal_event_dict(
-            self.system_data_source, self.org_1, location_id
-        )
-
-        self.client.force_authenticate(self.user)
-        response = self.client.post(url, data, format="json")
-        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
-
     def test_random_user_update_public_event_denied(self):
         url = reverse("event-detail", kwargs={"pk": self.event_4.id})
         location_id = reverse("place-detail", kwargs={"pk": self.place.id})
@@ -411,17 +400,6 @@ class TestEventAPI(APITestCase):
 
         self.client.force_authenticate(self.user)
         response = self.client.post(url, [data_1, data_2], format="json")
-        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
-
-    def test_random_user_bulk_update(self):
-        url = reverse("event-list")
-        location_id = reverse("place-detail", kwargs={"pk": self.place.id})
-        data = self.make_complex_event_dict(
-            self.system_data_source, self.org_1, location_id, self.languages
-        )
-
-        self.client.force_authenticate(self.user)
-        response = self.client.put(url, [data], format="json")
         self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
 
     def test_admin_get_own_public_event(self):
--- a/helevents/models.py
+++ b/helevents/models.py
@@ -1,6 +1,84 @@
+from functools import reduce
+
+from django_orghierarchy.models import Organization
 from helusers.models import AbstractUser
 
-from events.permissions import UserModelPermissionMixin
+from events.models import PublicationStatus
+
+
+class UserModelPermissionMixin:
+    """Permission mixin for user models
+
+    A mixin class that provides permission check methods
+    for user models.
+    """
+
+    def is_admin(self, publisher):
+        """Check if current user is an admin user of the publisher organization"""
+        raise NotImplementedError()
+
+    def is_regular_user(self, publisher):
+        """Check if current user is a regular user of the publisher organization"""
+        raise NotImplementedError()
+
+    @property
+    def admin_organizations(self):
+        raise NotImplementedError()
+
+    @property
+    def organization_memberships(self):
+        raise NotImplementedError()
+
+    def can_edit_event(self, publisher, publication_status):
+        """Check if current user can edit (create, change, modify)
+        event with the given publisher and publication_status"""
+        if self.is_admin(publisher):
+            return True
+        if (
+            self.is_regular_user(publisher)
+            and publication_status == PublicationStatus.DRAFT
+        ):
+            return True
+        return False
+
+    def get_editable_events(self, queryset):
+        """Get editable events queryset from given queryset for current user"""
+        # distinct is not needed here, as admin_orgs and memberships should not overlap
+        return queryset.filter(
+            publisher__in=self.get_admin_organizations_and_descendants()
+        ) | queryset.filter(
+            publication_status=PublicationStatus.DRAFT,
+            publisher__in=self.organization_memberships.all(),
+        )
+
+    def get_admin_tree_ids(self):
+        # returns tree ids for all normal admin organizations and their replacements
+        admin_queryset = self.admin_organizations.filter(
+            internal_type="normal"
+        ).select_related("replaced_by")
+        admin_tree_ids = admin_queryset.values("tree_id")
+        admin_replaced_tree_ids = admin_queryset.filter(
+            replaced_by__isnull=False
+        ).values("replaced_by__tree_id")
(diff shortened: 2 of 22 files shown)


helevents/models.py
```diff
@@ -1,6 +1,128 @@
...
+    for user models.
+    """
+
+    @property
+    def is_external(self):
+        """Check if the user is an external user"""
+        return (
+            not self.organization_memberships.exists()
+            and not self.admin_organizations.exists()
+        )
```
tuomas777, line 23: I guess this could be a cached property. 🤔 The performance difference is probably pretty negligible, but it would also prevent the value to be changed during a request, which could lead to hard-to-debug errors. Of course, the probability of that happening is pretty close to 0, but still 😄

I think the other methods, like `get_admin_organizations_and_descendants()` could use memoization as well, but that is not in the scope of this PR.

tuomas777 (approved):
Really superb work with the code cleaning and all!! 👍 🥇

I'm not very familiar with this LE permission stuff, so it would be good if @charn could also check this to make sure we are not missing anything.

helevents/models.py
```diff
@@ -1,6 +1,128 @@
...
+        in a subclass.
+        """
+        raise NotImplementedError()
+
+    def can_create_event(self, publisher, publication_status):
+        """Check if current user can create an event with the given publisher and publication_status"""
+        if self.is_admin_of(publisher):
+            return True
+        # Non-admins can only create drafts.
+        return publication_status == PublicationStatus.DRAFT
```
charn, line 59: If I'm just looking at this method, it now looks like it should be possible for anyone to create a draft event for any publisher. I wonder if it would make sense to do almost the same check as in the `can_edit_event` though here we would just not check `created_by`. I guess you could refactor `can_edit_event` and `can_create_event` to use a shared function or something in that case.

I might also add a way (setting) to disable this external event create/edit feature in case we need to deploy the backend withouth this feature enabled or in the case that later on we need to quickly disable it.

```suggestion
    def can_create_event(self, publisher, publication_status):
        """Check if current user can create an event with the given publisher and publication_status"""
        if self.is_admin_of(publisher):
            return True

        # Non-admins can only create drafts.
        if publication_status == PublicationStatus.DRAFT:
            if (
                settings.ENABLE_EXTERNAL_USER_EVENT_CREATE and
                ( publisher is None or publisher.id == settings.USER_DEFAULT_ORGANIZATION_ID)
            ):
                # External users can only create drafts for the default organization.
                return self.is_external and created_by == self
            else:
                # Regular users can edit drafts from organizations they are members of.
                return self.is_regular_user_of(publisher)

        return False
```

helevents/models.py
```diff
@@ -1,6 +1,128 @@
...
+
+    def get_editable_events(self, queryset):
+        """Get editable events queryset from given queryset for current user"""
+        if self.is_external:
+            # External users can only edit their own drafts from the default organization.
+            return queryset.filter(
+                created_by=self,
+                publisher__id=settings.USER_DEFAULT_ORGANIZATION_ID,
+                publication_status=PublicationStatus.DRAFT,
+            )
```
charn, line 88: I'd add `settings.ENABLE_EXTERNAL_USER_EVENT_CREATE`

```suggestion
        if self.is_external:
        
            if not settings.ENABLE_EXTERNAL_USER_EVENT_CREATE:
                return queryset.none()
                
            # External users can only edit their own drafts from the default organization.
            return queryset.filter(
                created_by=self,
                publisher__id=settings.USER_DEFAULT_ORGANIZATION_ID,
                publication_status=PublicationStatus.DRAFT,
            )
```

linkedevents/settings.py
```diff
@@ -293,8 +292,6 @@ def get_git_revision_hash() -> str:
 LOGOUT_REDIRECT_URL = "/"
 ACCOUNT_LOGOUT_ON_GET = True
 
-# User will be assigned to a default organization if user doesn't have any
-ENABLE_USER_DEFAULT_ORGANIZATION = env("ENABLE_USER_DEFAULT_ORGANIZATION")
 # Default organization to which a user will be assigned if user doesn't
 # have any organization. Default organization is created if it doesn't exist.
 USER_DEFAULT_ORGANIZATION_ID = env("USER_DEFAULT_ORGANIZATION_ID")
```
charn, line 297: This comment should be slightly changed, since the default organization is not set for the user anymore. I wonder if the setting variable should also be changed to something more descriptive. :thinking: Some quickly made up alternatives: `EVENT_DEFAULT_ORGANIZATION`, `EXTERNAL_USER_EVENT_ORGANIZATION`

events/tests/test_event_auth_api.py
```diff
@@ -370,17 +370,6 @@ def test_random_user_get_draft_event_not_found(self):
         response = self.client.get(url)
         self.assertEqual(response.status_code, status.HTTP_404_NOT_FOUND)
 
-    def test_random_user_create_event_denied(self):
```
charn, line 373: I'm not saying that this test shouldn't be deleted. Just thought that something similar could be tested when adding the setting `settings.ENABLE_EXTERNAL_USER_EVENT_CREATE`. So a test making sure that disabling `settings.ENABLE_EXTERNAL_USER_EVENT_CREATE` will work as intended.

helevents/models.py
danipran: Good point! Yeah, noticed that myself as well that some of the properties/methods/functions might as well be cached... but trying to stick to the scope here, it's way too easy to get sidetracked in this project 😅 

Turned this into a cached property.

helevents/models.py
danipran: Yup, you're right; had a brainfart here. Made it use `can_edit_event` with `created_by` set to `self`. Also added an `ENABLE_EXTERNAL_USER_EVENTS` flag.

helevents/models.py
```diff
@@ -1,6 +1,128 @@
...
+        """Check if current user can edit an event with the given publisher and publication_status"""
+        if self.is_admin_of(publisher):
+            return True
+
+        # Non-admins can only edit drafts.
+        if publication_status == PublicationStatus.DRAFT:
+            if (
+                publisher is None
+                or publisher.id == settings.USER_DEFAULT_ORGANIZATION_ID
+            ):
```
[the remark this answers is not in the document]
danipran: Done.

events/tests/test_event_auth_api.py
danipran: Added three new tests in `test_external_user.py` for create/update/delete when the flag is set to `False`.

charn (approved):
LGTM! Great work! :100: :tada: :+1: :shipit:

PR #605, opened 2023-06-01
LINK-1335 | Hide admin user endpoints from api root view

jorilindell:
## Description
- Check that user is admin user in any organization when trying to get list of organization classes or data sources
- Hide organization class, data source and feedback routes from api root view

## Closes
[LINK-1335](https://helsinkisolutionoffice.atlassian.net/browse/LINK-1335)


events/permissions.py
```diff
@@ -93,6 +93,11 @@ def has_permission(self, request, view):
             return False
 
 
+class GuestRetrieve(BasePermission):
+    def has_permission(self, request, view):
+        return request.method == "GET" and view.kwargs.get("pk")
```
charn, line 98: Afaik this can also be done with:
```suggestion
        return view.action == "retrieve"
```

jorilindell: Ah, makes sense. Changed and tested that is works as expected.

linkedevents/tests/test_api_root_view.py
```diff
@@ -0,0 +1,29 @@
...
+    assert response.data["keyword_set"] == "http://testserver/v1/keyword_set/"
+    assert response.data["place"] == "http://testserver/v1/place/"
+    assert response.data["language"] == "http://testserver/v1/language/"
+    assert response.data["organization"] == "http://testserver/v1/organization/"
+    assert response.data["image"] == "http://testserver/v1/image/"
+    assert response.data["event"] == "http://testserver/v1/event/"
+    assert response.data["search"] == "http://testserver/v1/search/"
+    assert response.data["user"] == "http://testserver/v1/user/"
+    assert response.data["registration"] == "http://testserver/v1/registration/"
+    assert response.data["signup"] == "http://testserver/v1/signup/"
```
charn, line 25: You could add `assert len(response.data) == 11`  here. It will fail when a new api route is added, but at least it needs to be set here then that should the new route be visible or not.

jorilindell: Good point. I was just adding an assert for `seats_reservation` route which was added after this PR and added an assert for route length as well.

charn (approved):
LGTM! :+1:

PR #609, opened 2023-06-08
LINK-1322 | Save admin and regular users when creating an organization

jorilindell:
## Description
At the moment `admin_users` and `regular_users` of an organization are not saved when creating a new organization. Fix this by using OrganizationDetailSerializer also when updating an organization instead of having separate serializer for update and having all logic to create/update organization in a single serializer. 
Change format of users in payload from
```
"admin_users": {
    "username": [
      "u-ui745qfmxmi63dvsbjmavaacea"
    ]
  },
```
to
```
"admin_users": ["u-ui745qfmxmi63dvsbjmavaacea"],
```
Show `admin_users` and `regular_users` field only for admin users and hide them if user has insufficient permissions.

PR for UI implementation: https://github.com/City-of-Helsinki/linkedcomponents-ui/pull/189

## Closes
[LINK-1322](https://helsinkisolutionoffice.atlassian.net/browse/LINK-1322)

[LINK-1322]: https://helsinkisolutionoffice.atlassian.net/browse/LINK-1322?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/events/api.py
+++ b/events/api.py
@@ -17,7 +17,6 @@ import django_filters
 import pytz
 import regex
 from django.conf import settings
-from django.contrib.auth.models import AnonymousUser
 from django.contrib.gis.geos import Point
 from django.contrib.gis.measure import D
 from django.contrib.postgres.search import SearchQuery, TrigramSimilarity
@@ -1483,18 +1482,58 @@ class OrganizationListSerializer(OrganizationBaseSerializer):
         return obj.regular_users.count() > 0
 
 
+class OrganizationUsersField(serializers.SlugRelatedField):
+    def __init__(self, **kwargs):
+        super().__init__(queryset=User.objects.all(), slug_field="username", **kwargs)
+
+    def to_representation(self, obj):
+        return UserSerializer(obj).data
+
+
 class OrganizationDetailSerializer(OrganizationListSerializer):
-    regular_users = serializers.SerializerMethodField()
-    admin_users = serializers.SerializerMethodField()
+    user_fields = ["admin_users", "regular_users"]
 
-    def validate(self, data):
-        if Organization.objects.filter(id=str(self.request.data["id"])):
-            raise DRFPermissionDenied(
-                f"Organization with id {self.request.data['id']} already exists."
-            )
-        return super().validate(data)
+    admin_users = OrganizationUsersField(
+        many=True,
+        required=False,
+    )
 
-    def connected_organizations(self, connected_orgs, created_org):
+    regular_users = OrganizationUsersField(
+        many=True,
+        required=False,
+    )
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        if self.instance:
+            self.fields["data_source"].read_only = True
+            self.fields["origin_id"].read_only = True
+
+    def to_representation(self, obj):
+        ret = super().to_representation(obj)
+
+        user = self.context["user"]
+        # Show admin users and regular users only to admins
+        user_fields = set(self.user_fields)
+        if user.is_anonymous or not user.is_admin_of(obj):
+            for field in user_fields:
+                del ret[field]
+
+        return ret
+
+    def validate_parent_organization(self, value):
+        if value:
+            user = self.request.user
+
+            if user.is_anonymous or not utils.organization_can_be_edited_by(
+                value, user
+            ):
+                raise DRFPermissionDenied(_("User has no rights to this organization"))
+
+        return value
+
+    def connect_organizations(self, connected_orgs, created_org):
         internal_types = {
             "sub_organizations": Organization.NORMAL,
             "affiliated_organizations": Organization.AFFILIATED,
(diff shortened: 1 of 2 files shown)


events/api.py
```diff
@@ -1483,18 +1482,58 @@ def get_has_regular_users(obj):
...
+
+    def to_representation(self, obj):
+        ret = super().to_representation(obj)
+
+        user = self.context["user"]
+        # Show admin users and regular users only to admins
+        user_fields = set(self.user_fields)
+        if user.is_anonymous or not user.is_admin_of(obj):
+            for field in user_fields:
+                del ret[field]
```
charn, line 1521: Imho you can remove this and put the following to `__init__`.
```python
        user = self.context["user"]
        # Show admin users and regular users only to admins
        if user.is_anonymous or not user.is_admin_of(obj):
            for field in self.user_fields:
                self.fields.pop(field, None)
```

events/api.py
```diff
@@ -1519,44 +1563,21 @@ def create(self, validated_data):
...
-            ).data
-        else:
-            return []
+    def update(self, instance, validated_data):
+        # Prevent user to accidentally remove himself from admin
+        if validated_data.get("admin_users") is not None:
+            admin_users = validated_data.pop("admin_users")
+            instance.admin_users.clear()
+            instance.admin_users.add(*admin_users)
+            instance.admin_users.add(self.user)
```
charn, line 1574: This part can be replaced with:

```suggestion
            instance.admin_users.set([*admin_users, self.user])
```

events/api.py
jorilindell: Fixed. Added also unit tests to make sure that regular user and anonymous user cannot see admin_users and regular_users

events/api.py
jorilindell: Changed as suggested

events/tests/test_organization_api.py
```diff
@@ -1,247 +1,399 @@
...
+
+    response = assert_update_organization(api_client, organization.id, payload)
+    assert response.data["name"] == payload["name"]
 
 
 @pytest.mark.django_db
-def test_delete_user_has_no_rights(user, user2, organization, api_client):
-    url = reverse("organization-list")
-    api_client.force_authenticate(user2)
+def test_anonymous_user_can_update_organization(api_client, organization):
```
charn, line 296: Probably should be `test_anonymous_user_cannot_update_organization`

jorilindell: Good catch. Fixed test name

events/api.py
```diff
@@ -1506,6 +1541,11 @@ def connected_organizations(self, connected_orgs, created_org):
             created_org.children.add(*conn_org)
 
     def create(self, validated_data):
+        # Add current user to admin users
+        if "admin_users" not in validated_data:
+            validated_data["admin_users"] = []
+        validated_data["admin_users"].append(self.user)
```
tuomas777, line 1547: I guess it is possible that after this the current user is duplicated in the list. Probably not an issue in practice, but a bit meh 🙂

events/tests/test_organization_api.py
```diff
@@ -1,247 +1,399 @@
...
+def assert_update_organization(api_client, pk, organization_data):
+    response = update_organization(api_client, pk, organization_data)
+    assert response.status_code == status.HTTP_200_OK
+    return response
+
+
+@pytest.mark.django_db
+def test_admin_user_can_see_organization_users(api_client, organization, user):
+    organization.regular_users.add(user)
     api_client.force_authenticate(user)
```
tuomas777, line 64: There is also `user_api_client` fixture that has the `user` already force authenticated 🙂

events/api.py
```diff
@@ -1483,18 +1482,54 @@ def get_has_regular_users(obj):
         return obj.regular_users.count() > 0
 
 
+class OrganizationUsersField(serializers.SlugRelatedField):
```
tuomas777, line 1485: As there is nothing that makes this field actually bound to multiple users (one can use it without `many=True` and that is nice) I would rather name it in singular form `OrganizationUserField`. Huge deal, I know.

events/tests/test_organization_api.py
jorilindell: Great, I didn't know about that. Changed `api_client` to `user_api_client` in all the possible tests in `events/tests/test_organization_api.py` file.

Fixed in https://github.com/City-of-Helsinki/linkedevents/pull/609/commits/e28447789e8f99dad0823b7496a6f3538b8a5484

events/api.py
jorilindell: Fixed in https://github.com/City-of-Helsinki/linkedevents/pull/609/commits/f641a100ef900d0db69ef6f32661ff695389b917

events/api.py
jorilindell: Fixed in https://github.com/City-of-Helsinki/linkedevents/pull/609/commits/25d0c77b46dbc5e272a40d47f18618f54b4a43ef

tuomas777 (approved):
LGTM 👍 🐃

PR #610, opened 2023-06-12
LINK-1367 | Support multilingual confirmation_message and instructions

jorilindell:
## Description
At the moment confirmation_message and instructions fields in SignUp model are single text fields. Change them to multilingual text fields.

## Closes
[LINK-1367](https://helsinkisolutionoffice.atlassian.net/browse/LINK-1367)

[LINK-1367]: https://helsinkisolutionoffice.atlassian.net/browse/LINK-1367?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


registrations/models.py
```diff
@@ -347,24 +347,44 @@ def can_be_edited_by(self, user):
...
         email_variables = {
             "linked_events_ui_url": settings.LINKED_EVENTS_UI_URL,
             "linked_registrations_ui_url": settings.LINKED_REGISTRATIONS_UI_URL,
             "username": self.name,
-            "event": self.registration.event.name_fi,
+            "event": getattr(
+                self.registration.event,
+                f"name_{service_language}",
+                self.registration.event.name,
+            ),
```
tuomas777, line 366: Unless there is a specific reason why not, it would be cleaner to use django-modeltranslation's standard way of accessing the translated fields, ie.
```
with translation.override(service_language):
    email_variables = {
        ...
        "event": self.registration.event
        ...
    }
```
One thing to consider is what kind of fallback language logic we want to use with these fields, and [configure modeltranslation](https://django-modeltranslation.readthedocs.io/en/latest/usage.html#fallback-languages) accordingly.

tuomas777 (approved):
Looking good! 👍 🦖

PR #613, opened 2023-06-14
LIIKUNTA-450 | Fetch data directly from Elis in kulke importer

tuhola:
### Description
Instead of reading event/category data from XML files exported from Elis using a script, directly call Elis API in the KulKe importer.

Also fix logging warnings for differing groups.

### DIFF
--- a/events/importer/kulke.py
+++ b/events/importer/kulke.py
@@ -301,8 +301,6 @@ class KulkeImporter(Importer):
     languages_to_detect = []
 
     def setup(self):
-        self.eventcodes_set = set()
-        self.file = open("eventcodes.txt", "w")
         self.languages_to_detect = [
             lang[0].replace("-", "_")
             for lang in settings.LANGUAGES
@@ -493,7 +491,6 @@ class KulkeImporter(Importer):
             return clean(text(k))
 
         eid = int(event_el.attrib["id"])
-        self.eventcodes_set.add(text_content("servicecode"))
         if text_content("servicecode") != "Pelkkä ilmoitus" and not is_course:
             # Skip courses when importing events
             return False
@@ -938,10 +935,6 @@ class KulkeImporter(Importer):
     def import_events(self):
         logger.info("Importing Kulke events")
         self._import_events()
-        for i in self.eventcodes_set:
-            self.file.write(i)
-            self.file.write("\n")
-        self.file.close()
 
     def import_courses(self):
         logger.info("Importing Kulke courses")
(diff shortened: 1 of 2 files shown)


events/importer/kulke.py
```diff
@@ -31,6 +34,14 @@
 # Per module logger
 logger = logging.getLogger(__name__)
 
+
+EVENT_API_BASE_URL = os.getenv("EVENT_API_BASE_URL", "http://elis.helsinki1.hki.local/event-api/")
```
tuomas777, line 38: Other env-based configs in this project go through `settings.py` so this one should as well. 

Also, as the env variable is global, it would be good if it was named a bit more specifically to Elis/Kulke (there is already a lot of "event" stuff in this project 😄). Maybe `ELIS_EVENT_API_BASE_URL` ?

events/importer/kulke.py
```diff
@@ -930,15 +940,28 @@ def import_courses(self):
         logger.info("Importing Kulke courses")
         self._import_events(importing_courses=True)
 
+    def _iter_elis_events(self, language: str, begin_date: datetime) -> Iterator[etree.Element]:
+        begin_date = begin_date.strftime('%d.%m.%Y')
+        offset = 0
+        parser = etree.XMLParser(recover=True)
+        while True:
+            logger.debug("Fetching events: %s - %d", language, offset)
+            response = requests.get(EVENTS_URL_TEMPLATE.format(begin_date=begin_date, offset=offset, language=language))
```
tuomas777, line 949: Ditto.

events/importer/kulke.py
tuhola: 👍 Moved to settings and named it `ELIS_EVENT_API_URL`, which seems to be in line with other URLs there.

events/importer/kulke.py
```diff
@@ -382,12 +393,11 @@ def setup(self):
         except License.DoesNotExist:
             self.event_only_license = None
 
-    def parse_kulke_categories(self):
+    def fetch_kulke_categories(self) -> dict[str, Union[str, int]]:
+        response = requests.get(CATEGORY_URL)
```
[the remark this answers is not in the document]
tuhola: Added a timeout.

events/importer/kulke.py
tuhola: Added one here too.

tuomas777 (approved):
LGTM, good job! 👍

PR #616, opened 2023-06-19
YSO importer: Only accept supported languages when updating

charn:
In a similar way to importing alt labels, only accept supported languages when updating keywords. If a language is not supported (in this case configured in Django) the default language will apparently be used, which caused Finnish language to have mixed translations.

Refs LINK-1299


jorilindell (approved):
Very nice. Tested locally and it works as expected.

PR #617, opened 2023-06-19
LINK-1299 | YSO importer keyword deprecation handles invalid YSO ids

charn:
Deprecate and replace functionality failed if an invalid YSO id was encountered.


tuomas777 (approved):
LGTM 👍

PR #619, opened 2023-06-20
LINK-1368 | Add service_language field to Language model

jorilindell:
## Description
- Support service_language query param in language list endpoint
- Show service_language field in Django admin
- Allow to select only languages which has service_language as signup service_language

PRs for UIs:
https://github.com/City-of-Helsinki/linkedcomponents-ui/pull/191
https://github.com/City-of-Helsinki/linkedregistrations-ui/pull/56

## Closes
[LINK-1368](https://helsinkisolutionoffice.atlassian.net/browse/LINK-1368)

[LINK-1368]: https://helsinkisolutionoffice.atlassian.net/browse/LINK-1368?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


events/api.py
```diff
@@ -1417,6 +1417,17 @@ class LanguageViewSet(JSONAPIViewMixin, viewsets.ReadOnlyModelViewSet):
...
+        query_params = self.request.query_params
+
+        val = query_params.get("service_language", None)
+        if val is not None:
+            queryset = queryset.filter(
+                service_language=parse_bool(val, "service_language")
+            )
+
+        return queryset
+
```
tuomas777, line 1430: While this works and is consistent with many currently existing filters in LE, it also represents the kind of DIY style we want to get replaced by a more generic django-filter based solution, so ideally new filter code should use that. In this case I **think** just adding `filterset_fields` to the viewset should work (not sure at all, pls test 😄):
```suggestion
    filterset_fields = ("service_language",)
```

jorilindell: Nice suggestion. Changed as suggested.

tuomas777 (approved):
LGTM 👍 🦭

PR #622, opened 2023-06-26
Use transactions in kulke importer

tuhola:
Currently the kulke importer does not use transactions, meaning that if it's interrupted, the result might contain inconsistencies.

Also a couple of other changes:
- handle super event metadata during super event creation instead of doing it separately afterwards
- default to using the first event's name for the super event's name in cases where no common part among the member event names is found
- use bulk create/update instead of looped creates/updates where possible
- delete existing super events if they only contain one event

N.B. the transaction change to `save_event` and `save_place` will affect all importers, not just kulke.

### DIFF
--- a/events/importer/kulke.py
+++ b/events/importer/kulke.py
@@ -10,7 +10,7 @@ import dateutil
 import requests
 from django.conf import settings
 from django.core.exceptions import ObjectDoesNotExist
-from django.db import IntegrityError
+from django.db import transaction
 from django_orghierarchy.models import Organization
 from lxml import etree
 from pytz import timezone
@@ -132,6 +132,11 @@ CATEGORIES_TO_IGNORE = [
     55,
 ]
 
+CATEGORY_TYPES_TO_IGNORE = [
+    2,
+    3,
+]
+
 # Events having one of these categories are courses - they are excluded when importing events
 # and only they are included when importing courses.
 COURSE_CATEGORIES = {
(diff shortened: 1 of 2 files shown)


events/importer/kulke.py
```diff
@@ -851,88 +871,84 @@ def simple(field):
...
+            self._update_super_event(super_event, events)
+            super_event.save()
+            aggregate.super_event = super_event
+            aggregate.save()
+            event_aggregates = [
+                EventAggregateMember(event=event, event_aggregate=aggregate)
+                for event in events
+            ]
+            EventAggregateMember.objects.bulk_create(event_aggregates)
+        elif n_super_events == 1:
```
tuomas777, line 934: Here `n_super_events` must be `1` because we have already checked that it cannot be `>1` or `0`.
```suggestion
        else:
```

events/importer/kulke.py
```diff
@@ -851,88 +871,84 @@ def simple(field):
...
-                aggregate.super_event = super_event
-                aggregate.save()
+        aggregate = None
+        if n_super_events == 0:
+            # Don't create a super event if there is only one event in the database
+            if len(events) < 2:
+                return False
+
+            aggregate = EventAggregate()
+            aggregate.save()
```
tuomas777, line 918:
```suggestion
            aggregate = EventAggregate.objects.create()
```

events/importer/kulke.py
```diff
@@ -851,88 +871,84 @@ def simple(field):
...
-                super_event = Event(
-                    publisher=self.organization,
-                    super_event_type=Event.SuperEventType.RECURRING,
-                    data_source=DataSource.objects.get(pk="kulke"),  # TODO
-                    id="linkedevents:agg-{}".format(aggregate.id),
-                )
-                super_event.save()
-                aggregate.super_event = super_event
-                aggregate.save()
+        aggregate = None
```
tuomas777, line 911:
```suggestion
```

events/importer/kulke.py
tuhola: Fixed

tuomas777 (approved):
LGTM, good job! 👍 🐯

PR #623, opened 2023-06-27
Add doc generation to Kulke importer

tuhola:
Generate docs about the Kulke importer with the target audience being someone who creates events and needs to understand how the event will be represented in LE.

The docs will contain information about the various data transformations and mappings that the importer performs.

### DIFF
--- a/events/importer/kulke.py
+++ b/events/importer/kulke.py
@@ -6,7 +6,7 @@ import re
 from datetime import datetime, time, timedelta
 from posixpath import join as urljoin
 from textwrap import dedent
-from typing import Iterator, Optional, Union
+from typing import Iterator, Union
 
 import dateutil
 import requests
@@ -1032,14 +1032,10 @@ class KulkeImporter(Importer):
                 )
 
     @staticmethod
-    def generate_documentation_md(
-        output_file: Optional[str] = None,
-    ) -> None:  # noqa: C901
+    def generate_documentation_md() -> str:
         """
         Generate MarkDown document covering Kulke importer logic and mappings.
-        :param output_file: (optional) Output file to write MarkDown document into.
-                            Writes to stdout if not given
-        :return: nothing
+        :return: documentation string
         """
 
         from snakemd import new_doc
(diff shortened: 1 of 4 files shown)


events/importer/kulke.py
```diff
@@ -1019,3 +1023,328 @@ def import_keywords(self):
                     data_source=self.data_source,
                     publisher=self.organization,
                 )
+
+    @staticmethod
+    def generate_documentation_md(
+        output_file: Optional[str] = None,
+    ) -> None:  # noqa: C901
```
tuomas777, line 1030: From reusability and testing POV it would be better if there was a function that returned the documentation string instead of printing it.

events/management/commands/event_importer_doc.py
```diff
@@ -0,0 +1,65 @@
...
+        name = "generate_documentation_md"
+        method = getattr(importer, name, None)
+        if not method:
+            raise CommandError(
+                "Importer {} does not support documentation generation".format(
+                    importer.name
+                )
+            )
+
+        method(options["output_file"])
```
tuomas777, line 65: I guess using `translation.override()` would be a bit more robust. I also like walruses.
```suggestion
        method_name = "generate_documentation_md"
        if not method := getattr(importer, method_name, None):
            raise CommandError(
                "Importer {} does not support documentation generation".format(
                    importer.name
                )
            )
        
        # Activate the default language
        # to make sure translated fields are populated correctly.
        with translation.override(settings.LANGUAGES[0][0]):
            method(options["output_file"])
```

events/importer/kulke.py
Cap-JaTu: If returning the document as a string, where the printing or output to file should happen?

tuhola: That could happen in the doc generation management command. I agree that it's better for reusability, so I'll change it.

events/management/commands/event_importer_doc.py
tuhola: 👍

tuomas777 (approved):
LGTM 👍 💯 📜

PR #625, opened 2023-06-29
Delete events not found in Elis in kulke importer

tuhola:
If there are events that have previously been imported, but have since been removed in Elis, there's currently no mechanism to remove them from LE. This PR adds such a mechanism. Basically we check for any events that we have previously imported that no longer exist in Elis. We won't touch events that are older than the start date we use for the Elis search though.

Also, the fix for the empty name issue in my previous PR was not sufficient, so I fixed that. The logic for taking the common part when creating a super event name should now work in all supported languages.

When reviewing, please pay special attention to the logic for selecting which events to delete.

### DIFF
--- a/events/importer/kulke.py
+++ b/events/importer/kulke.py
@@ -13,7 +13,7 @@ import requests
 from django.conf import settings
 from django.core.exceptions import ObjectDoesNotExist
 from django.db import transaction
-from django.db.models import Count
+from django.db.models import Count, Q
 from django_orghierarchy.models import Organization
 from lxml import etree
 from pytz import timezone
@@ -925,7 +925,8 @@ class KulkeImporter(Importer):
                 # The imported event is not part of an aggregate but one was found it in the db.
                 # Remove the super event. This is the only case when an event is removed from
                 # a recurring aggregate.
-                aggregate.super_event.delete()
+                aggregate.super_event.soft_delete()
+                aggregate.super_event.sub_events.all().update(super_event=None)
                 return False
             else:
                 for event in events:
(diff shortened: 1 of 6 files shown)


events/importer/kulke.py
```diff
@@ -838,33 +834,34 @@ def simple(field):
...
+
+                if name:
+                    setattr(super_event, name_attr, name.rstrip())
+                else:
+                    # If a common part was not found, default to the first event's name
+                    setattr(
+                        super_event,
+                        name_attr,
+                        getattr(first_event, name_attr),
+                    )
```
danipran, line 864: Not necessarily a better way of doing this, but an alternative that's a bit easier to read:
```suggestion
                for word in words:
                    if all(
                        member_event_name and
                        member_event_name.startswith(name + word)
                        for member_event_name in member_event_names
                    ):
                        name += word + " "
                    else:
                        name = name.rstrip()
                        break

                # If a common part was not found, default to the first event's name
                setattr(super_event, name_attr, name or getattr(first_event, name_attr))
```
Assumes that `words` is always a list. Might require some extra formatting from black.

events/importer/kulke.py
```diff
@@ -939,12 +936,46 @@ def _save_recurring_superevent(self, recurring_group):
...
+            .filter(aggregate_member_count__lt=2)
+            .delete()
+        )
+        if count:
+            logger.debug(
+                "Deleted %d empty super events and associated objects: %s",
+                count,
+                deleted,
+            )
+
```
danipran, line 978: AFAIK we don't delete events, but soft delete them instead. E.g. `some_event.soft_delete()`. Not sure what the case is with importers though, but I can't find any references in the code base to anything that actually deletes events, so I suppose it's the safer option.

Don't think there's a queryset method for soft deletion though (e.g. `Event.objects.all().soft_delete()`), so you'd have to either add one or do it locally. For reference, here's the `Event` model's `soft_delete` method:
```
    def soft_delete(self, using=None):
        self.deleted = True
        self.save(update_fields=("deleted",), using=using, force_update=True)
```

danipran (changes requested):

events/tests/importers/test_kulke.py
```diff
@@ -236,9 +236,9 @@ def test__handle_removed_events(self):
             begin_date=now - timedelta(days=60),
         )
 
-        self.assertFalse(Event.objects.filter(id=event_1.id).exists())
-        self.assertTrue(Event.objects.filter(id=event_2.id).exists())
-        self.assertTrue(Event.objects.filter(id=event_3.id).exists())
+        self.assertFalse(Event.objects.filter(id=event_1.id, deleted=False).exists())
```
charn, line 239: This is a soft deleted event, right? If so would it be safer to test that:

```python
self.assertTrue(Event.objects.filter(id=event_1.id, deleted=True).exists())
```

events/tests/importers/test_kulke.py
```diff
@@ -269,10 +269,18 @@ def test__handle_removed_events_superevent(self):
...
+        )
+        self.assertFalse(Event.objects.filter(id=super_1.id, deleted=False).exists())
+        self.assertTrue(
+            Event.objects.filter(id=super_2_event_1.id, deleted=False).exists()
+        )
+        self.assertTrue(
+            Event.objects.filter(id=super_2_event_2.id, deleted=False).exists()
+        )
+        self.assertTrue(Event.objects.filter(id=super_2.id, deleted=False).exists())
+        self.assertFalse(Event.objects.filter(id=super_3.id, deleted=False).exists())
```
charn, line 286: Same here, if there's any soft deleted events, maybe we could test that `self.assertTrue(Events.objects.filter(..., deleted=True).exists())`

charn (approved):
Couldn't spot anything serious. :+1: :shipit: A minor comment about testing those soft deleted events.

events/tests/importers/test_kulke.py
danipran: Yeah, very true.

PR #631, opened 2023-07-06
LINK-1353 | Registration specific permissions

jorilindell:
## Description
- RegistrationUserAccess model is used to grant permissions to view signups of a single registration. Registration user access is saved as an email and logged in user can view signups of the registration is email addresses match.
- An email notification is sent when new registration users are added or email is updated
- Also support resending invitation email to a registration user access
- Support updating registration users accesses in registration form (Django admin)

PR for FE: https://github.com/City-of-Helsinki/linkedcomponents-ui/pull/198

## Closes
[LINK-1353](https://helsinkisolutionoffice.atlassian.net/browse/LINK-1353)

## Screenshots
Screenshot 2023-07-07 at 17 11 36

[LINK-1353]: https://helsinkisolutionoffice.atlassian.net/browse/LINK-1353?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


events/api.py
```diff
@@ -1840,16 +1841,48 @@ class RegistrationSerializer(LinkedEventsSerializer, RegistrationBaseSerializer)
         queryset=Event.objects.all(),
     )
 
+    only_admin_visible_fields = ("created_by", "last_modified_by", "registration_users")
+
+    def _create_registration_users(self, registration, registration_users):
+        for registration_user in registration_users:
+            [user_instance, created] = RegistrationUser.objects.get_or_create(
```
charn, line 1848: I'd remove the brackets
```suggestion
            user_instance, created = RegistrationUser.objects.get_or_create(
```

registrations/models.py
```diff
@@ -236,6 +239,59 @@ def is_user_editable_resources(self):
         )
 
 
+class RegistrationUser(models.Model):
+    email = models.EmailField(verbose_name=_("E-mail"), null=True, blank=True)
```
charn, line 243: Since this is the only new information the model holds, I would think that removing `null=True, blank=True` would probably make sense.

registrations/serializers.py
```diff
@@ -182,16 +190,16 @@ def get_signups(self, obj):
         if "signups" in params.get("include", "").split(","):
             #  only the organization admins should be able to access the signup information
             user = self.context["user"]
-            event = obj.event
-            if not isinstance(user, AnonymousUser) and user.is_admin_of(
-                event.publisher
+            registration_user_emails = [u.email for u in obj.registration_users.all()]
```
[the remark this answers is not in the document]
jorilindell: Fixed in https://github.com/City-of-Helsinki/linkedevents/pull/631/commits/22a53edd3def57c8004e3b1787b9c0942d3b1f1f

events/api.py
jorilindell: Fixed in https://github.com/City-of-Helsinki/linkedevents/pull/631/commits/22a53edd3def57c8004e3b1787b9c0942d3b1f1f

registrations/models.py
jorilindell: Very good point: Fixed in https://github.com/City-of-Helsinki/linkedevents/pull/631/commits/a82f25d59e6988d2771614afe85cd35f7aa5e045

events/api.py
```diff
@@ -1840,16 +1841,48 @@ class RegistrationSerializer(LinkedEventsSerializer, RegistrationBaseSerializer)
         queryset=Event.objects.all(),
     )
 
+    only_admin_visible_fields = ("created_by", "last_modified_by", "registration_users")
+
+    def _create_registration_users(self, registration, registration_users):
+        for registration_user in registration_users:
+            user_instance, created = RegistrationUser.objects.get_or_create(
```
harriris-vincit, line 1848: Is it possible that existing RegistrationUsers' emails are updated? In that case `update_or_create` would be more suitable, and perhaps there could be another private method `_update_registration_users` for those RegistrationUsers that already exist in the DB?

harriris-vincit: Though the current logic in these cases recreates the users with new emails and in a way already "updates" them. Updating the usual way as suggested in my previous comment should be a bit more efficient than deleting and the recreating.

registrations/models.py
```diff
@@ -236,6 +239,59 @@ def is_user_editable_resources(self):
...
+    )
+
+    def can_be_edited_by(self, user):
+        """Check if current registration user can be edited by the given user"""
+        if user.is_superuser:
+            return True
+        return user.is_admin_of(self.registration.event.publisher)
+
+    def send_invitation(self):
+        locale = "fi"
```
harriris-vincit, line 256: Would it be possible to send the UI locale when doing the API request when creating or updating a RegistrationUser, store that too into RegistrationUser and use it here to override the invitation email language (as long as it would be one of the supported langues in which case a fallback like this current one could be used)?

registrations/permissions.py
```diff
@@ -52,3 +52,16 @@ def has_object_permission(self, request: Request, view, obj):
...
+    def has_permission(self, request, view):
+        # Only authenticated users can get object
+        return view.action == "retrieve" and request.user.is_authenticated
+
+    def has_object_permission(self, request: Request, view, obj):
+        user = request.user
+        if view.action != "retrieve":
+            return False
+
+        return len(obj.registration.registration_users.filter(email=user.email))
```
harriris-vincit, line 67:
```suggestion
        return obj.registration.registration_users.filter(email=user.email).exists()
```

registrations/models.py
jorilindell: I just want to confirm before commiting that I get this rights. Is the idea to add `language` field to RegistrationUser model and then use it as email language? e.g.

```
class RegistrationUser(models.Model):
    email = models.EmailField(verbose_name=_("E-mail"))

        Registration, on_delete=models.CASCADE, related_name="registration_users"
    )

    language = models.ForeignKey(
        Language,
        on_delete=models.SET_NULL,
        null=True,
        blank=True,
        related_name="registration_user_language",
    )

    def _get_language_pk(self):
        if self.language:
            return self.language.pk
        return "fi"

    def send_invitation(self):
        locale = self._get_language_pk()

        with translation.override(locale):
            event_name = self.registration.event.name
            event_type_id = self.registration.event.type_id

            email_variables = {
                "email": self.email,
                "event": event_name,
                "event_type_id": event_type_id,
                "linked_events_ui_locale": locale,
                "linked_events_ui_url": settings.LINKED_EVENTS_UI_URL,
                "linked_registrations_ui_locale": locale,
                "linked_registrations_ui_url": settings.LINKED_REGISTRATIONS_UI_URL,
                "registration_id": self.registration.id,
            }

            invitation_template = "signup_list_rights_granted.html"
            invitation_subject = _(
                "Rights granted to the participant list - %(event_name)s"
            ) % {"event_name": event_name}

            rendered_body = render_to_string(invitation_template, email_variables)

            try:
                send_mail(
                    invitation_subject,
                    rendered_body,
                    get_email_noreply_address(),
                    [self.email],
                    html_message=rendered_body,
                )
            except SMTPException:
                pass
```

In that case it would be nice to allow only Languages where `service_languauge` is set to True. Similar logic is used in email sending to signups
```
class RegistrationUserSerializer(serializers.ModelSerializer):
    language = serializers.PrimaryKeyRelatedField(
        queryset=Language.objects.filter(service_language=True),
        many=False,
        required=False,
    )

    class Meta:
        model = RegistrationUser
        fields = ["id", "email"]
```

registrations/permissions.py
jorilindell: Fixed in https://github.com/City-of-Helsinki/linkedevents/pull/631/commits/11834ffbec8e2c83886750f3d7760a08351729cf

registrations/models.py
harriris-vincit: Yes, this is what I had in mind in terms of the logic (didn't think too much into the details of how it would be best implemented, but your suggestions looks good)

events/api.py
jorilindell: My Friday-brains didn't find perfect solution to implement what you suggested, but I changed
```
user_instance, created = RegistrationUser.objects.get_or_create(
                registration=registration, **registration_user
)
```
to
```
user_instance, created = RegistrationUser.objects.update_or_create(
                email=registration_user["email"],
                registration=registration,
                defaults=registration_user,
            )
```
so user can update `language` value via API.

harriris-vincit: You would need to do it like this to implement what I suggested:
```
RegistrationUser.objects.update_or_create(id=<id>, registration=registration, defaults={"email": <email>, "language": <language>})
```
where `id` can be `None` for a new user.

harriris-vincit: I am thinking that for robust data integrity the `unique_together` should probably be kept, but in the case where there would be another registration user with the same email address as in the `update` action data, somebody (API client, admin, etc.) would need to delete the existing entry based on the validation error and recreate it with the updated email. So perhaps just return the validation error in this case and let the API user handle the error?

registrations/models.py
```diff
@@ -250,6 +253,69 @@ def is_user_editable_resources(self):
...
+        Registration, on_delete=models.CASCADE, related_name="registration_users"
+    )
+
+    language = models.ForeignKey(
+        Language,
+        on_delete=models.SET_NULL,
+        null=True,
+        blank=True,
+        related_name="registration_user_language",
+    )
```
danipran, line 269: This is my personal preference for using empty lines:
```suggestion
    email = models.EmailField(verbose_name=_("E-mail"))

    registration = models.ForeignKey(
        Registration, on_delete=models.CASCADE, related_name="registration_users"
    )
    language = models.ForeignKey(
        Language,
        on_delete=models.SET_NULL,
        null=True,
        blank=True,
        related_name="registration_user_language",
    )
```
An empty line between every single field/attribute makes the code a bit more frustrating to read, especially if you have a lot of one-liners (which isn't the case here). Better to use them for grouping, e.g. regular fields, then an empty line, then all the foreign keys, etc.

For example:
```py
class SomeModel(models.Model):
    email = models.EmailField(...)
    first_name = models.CharField(...)
    last_name = models.CharField(...)

    foo = models.ForeignKey(...)
    bar = models.ForeignKey(...)

    some_custom_attribute = ...

    def method():
        ...
```
is much more pleasant to read than:

```py
class SomeModel(models.Model):
    email = models.EmailField(...)

    first_name = models.CharField(...)

    last_name = models.CharField(...)

    foo = models.ForeignKey(...)

    bar = models.ForeignKey(...)

    some_custom_attribute = ...

    def method():
        ...
```

Of course, this is completely a matter of personal preference; I just don't like having too many empty lines. Do what you will.

registrations/models.py
```diff
@@ -250,6 +253,69 @@ def is_user_editable_resources(self):
         )
 
 
+class RegistrationUser(models.Model):
```
danipran, line 256: Is this model supposed to represent an user or an participant/attendee? The name used here is a bit confusing for me at least; if I see `FooUser`, I assume it's an actual user model that's been inherited from `AbstractUser` or something 😅

registrations/serializers.py
```diff
@@ -159,6 +159,26 @@ class Meta:
         model = SignUp
 
 
+class RegistrationUserIdField(serializers.PrimaryKeyRelatedField):
+    def get_queryset(self):
+        registration_id = self.context["request"].parser_context["kwargs"]["pk"]
+        return RegistrationUser.objects.filter(registration__pk=registration_id)
```
danipran, line 165: Why does this exist? I don't quite get it 😅

registrations/serializers.py
```diff
@@ -196,16 +218,18 @@ def get_signups(self, obj):
...
-            event = obj.event
-            if not isinstance(user, AnonymousUser) and user.is_admin_of(
-                event.publisher
+            registration_user_emails = obj.registration_users.all().values_list(
+                "email", flat=True
+            )
+
+            if not user.is_anonymous and (
+                user.is_admin_of(obj.event.publisher)
+                or user.email in registration_user_emails
```
danipran, line 227:
```suggestion
            if not user.is_anonymous and (
                user.is_admin_of(obj.event.publisher)
                or obj.registration_users.filter(email=user.email).exists()
```

registrations/utils.py
```diff
@@ -49,3 +50,7 @@ def send_mass_html_mail(
         messages.append(mail)
 
     return connection.send_messages(messages)
+
+
+def get_email_noreply_address():
+    return "noreply@%s" % Site.objects.get_current().domain
```
danipran, line 56: Pretty magical. Would be nice if you can override this through a setting.

danipran (changes requested):
BIt of a massive PR. This should've been at least two separate PRs (one for permissions, one for the rest), if not more. Took a while to review.

events/api.py
jorilindell: Good point. Updating registration users so that it will cause integrity errors is kind of edge case. Removed the migration file to remove `unique_together`. Fixed in https://github.com/City-of-Helsinki/linkedevents/pull/631/commits/025bfe395b5fe917bdc02ca0a4d29814470c29bd

events/api.py
```diff
@@ -1839,18 +1840,101 @@ class RegistrationSerializer(LinkedEventsSerializer, RegistrationBaseSerializer)
...
+
         return instance
 
+    def validate_registration_users(self, value):
+        errors = []
+        emails = []
+
+        for data in value:
+            email = data["email"]
+            if emails.__contains__(email):
```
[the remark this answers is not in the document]
jorilindell: Fixed in https://github.com/City-of-Helsinki/linkedevents/pull/631/commits/0c09190f408bafde8d1e1bc6d66d6872c8c19180

events/api.py
```diff
@@ -1839,18 +1840,101 @@ class RegistrationSerializer(LinkedEventsSerializer, RegistrationBaseSerializer)
...
+                                code="unique",
+                            )
+                        ]
+                    }
+                )
+            else:
+                emails.append(email)
+                errors.append({})
+
+        if len([e for e in errors if e != {}]):
```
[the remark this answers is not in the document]
jorilindell: `if errors` didn't quite work but caused couple of failing tests. Added `raise_errors` variables which is set to True when there is a need to raise errors
 
Fixed in https://github.com/City-of-Helsinki/linkedevents/pull/631/commits/0c09190f408bafde8d1e1bc6d66d6872c8c19180

registrations/models.py
jorilindell: Purpose of this model is to grant read permissions to signup list of a single registration/event. RegistrationUsers are the persons who check the participants present while the event is going on. 
We use email to check that current user has permissions to read signups so make it easier for event/registration admin to grant permissions.  

I agree with you that `RegistrationUser` is quite confusing name. Do you have any suggestion for better name? 
Would something like `RegistrationAccess` make more sense? 

I was also thinking RegistrationPermissions but that's confusing as well.

events/api.py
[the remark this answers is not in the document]
danipran: Right you are. Forgot that the valid list items need to be represented as well in the error list. The error flag is a very nice solution, I like it!

registrations/utils.py
jorilindell: Good point. At the moment noreply address is defined in several files. Added a separate JIRA ticket to have only one method to define noreply address and to have environment variable to override it: https://helsinkisolutionoffice.atlassian.net/browse/LINK-1451

registrations/serializers.py
jorilindell: The reason to have RegistrationUserIdField class is to validate registration user id values. They have to exist and to be in same registration that user is updating. So the user cannot accidentally edit registration_users of other registration and move them to the registration which he is updating.

jorilindell:
> BIt of a massive PR. This should've been at least two separate PRs (one for permissions, one for the rest), if not more. Took a while to review.

Yeah, sorry about the massive size of the PR. I opened it just before my vacation and wanted to get it out of my hands. Should have done several smaller PRs.

registrations/models.py
danipran: Hm... okay, so if I read the code correctly, you need to be an authenticated user to see all of the signups, right? And we're using this through the Registration model to check if a user's email matches any of the emails in `registration_users`... so isn't this basically something that could be done as a many-to-many relation between `Registration` and `User`? 😅 

As for the name, if this can't be done as a many-to-many relation instead then I suppose it acts like a read access table? So something like `RegistrationAccess` or `RegistrationUserAccess` would be more descriptive 🤔

jorilindell: This would be nice to be done with many-to-many relation between `Registration` and `User`. But the problem with that approach is that the user might not exist when admin user is creating/updating `Registration`. Also, at the moment only superadmin users are allowed to get list of users, so admin user wouldn't be able to search users that he want to add to a registration.

My personal favourite for the name is `RegistrationUserAccess`?

@harriris-vincit What do you think?

registrations/serializers.py
danipran: Makes sense.

registrations/models.py
harriris-vincit: `RegistrationUserAccess` seems good enough considering also the terminology that we have agreed (should try to match the commonly known names with the ones we use in the code, and we have agreed "Registration user" as the name for this role).

registrations/api.py
```diff
@@ -274,7 +299,12 @@ def filter_queryset(self, queryset):
                         _("Registration with id {pk} doesn't exist.").format(pk=pk)
                     )
 
-                if registration.publisher not in admin_organizations:
+                if not (
```
harriris-vincit, line 302: Should we also allow superusers to have access?

registrations/api.py
```diff
@@ -284,9 +314,12 @@ def filter_queryset(self, queryset):
                 registrations.append(registration)
             queryset = queryset.filter(registration__in=registrations)
         elif self.action == "list":
```
harriris-vincit, line 316: Also here: perhaps superusers should be allowed to perform the `list` action?

registrations/models.py
jorilindell: Ok, I renamed `RegistrationUser` model as `RegistrationUserAccess`. Changes are committed in https://github.com/City-of-Helsinki/linkedevents/pull/631/commits/bfc5875f9bf8b91cc78a1bb390231d39859f16b9

registrations/api.py
jorilindell: Allowed superusers to have access in https://github.com/City-of-Helsinki/linkedevents/pull/631/commits/2a60a351dca92e2849af5afd8ace752c381dda44

registrations/api.py
jorilindell: Allowed superusers to have access in https://github.com/City-of-Helsinki/linkedevents/pull/631/commits/2a60a351dca92e2849af5afd8ace752c381dda44

registrations/migrations/0022_rename_registrationuser_as_registrationuseraccess.py
harriris-vincit, line 1: All of these three migrations are introduced in this same PR. It would be good to have just a single migration file for all of them so then locally you can do this:
1. `./manage.py migrate registrations 0019` (unapplies the three new ones)
2. Delete these three migrations files.
3. `./manage.py makemigrations`
4. Migrate again locally.

jorilindell: Migration files are merged in https://github.com/City-of-Helsinki/linkedevents/pull/631/commits/e363ce50c5d4a5b46db22ebf7e9e4ce6d2b00c27
I had to do something to migration files anyway since `0020_signup_presence_status.py` was added to `main` brach after this PR was opened.

harriris-vincit (approved):

danipran (approved):

PR #635, opened 2023-07-20
LINK-1392 | Free used memory from uWSGI processes

charn:
Reload workers regularly to keep memory fresh and ease potential memory leaks.


danipran (approved):
LGTM :+1:

PR #638, opened 2023-07-25
Fix: build translations for production container

charn:
This change builds the translations for the production container in `Dockerfile`. Translations were previously only built for the local development container, which has some issues since the `docker-compose.yml` mounts over the ` /app` folder gets. For compiling translations in development container a `COMPILE_TRANSLATIONS` variable was added.


danipran (approved):
LGTM :+1:

PR #642, opened 2023-07-28
LINK-1384 | Expose User.is_external property in the API

charn:
This adds the is_external flag to the API.


danipran (approved):
LGTM :+1:

PR #643, opened 2023-07-28
LINK-1080: Miscellaneous importer changes

danipran:
- Add exception information in some of the kulke/harrastushaku importer error logs
  - Also change the error messages for harrastushaku error loggers.
  - These both are mostly for making Sentry more helpful.
  - I also considered making `HarrastushakuException` related log messages into warnings, since they are essentially warnings, but wasn't quite sure how Sentry handles warnings, so left them as errors.
- Add YSO replacement for Suvilahti (deprecated: [Suvilahti](https://finto.fi/yso/fi/page/p27158), replaced by: [Suvilahti (Helsinki)](https://finto.fi/yso-paikat/fi/page/p508707))
  - Due to this error: `Exception: Deprecating YSO keyword Suvilahti that is referenced in events <MultilingualBaseTreeQuerySet [<Event: D-A-D & Melrose (helsinki:af25a6ji4u) 2022-08-25 15:00:00+00:00>]>. No replacement keyword was found in YSO. Please manually map the keyword to a new keyword in YSO_DEPRECATED_MAPS.`
  - Not sure why the importer can't find the replacement by itself since it's very apparent in the website at least, but then again, I didn't really take a stab at the API side of things

Probably doesn't fix everything, but should give more info in Sentry and hopefully resolve at least one hurdle for the YSO importer.

### DIFF
--- a/events/importer/harrastushaku.py
+++ b/events/importer/harrastushaku.py
@@ -1,5 +1,4 @@
 import logging
-import traceback
 from collections import namedtuple
 from copy import deepcopy
 from datetime import datetime, timedelta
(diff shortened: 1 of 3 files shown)


charn (approved):
Looks awesome to me! :+1: :shipit:

PR #644, opened 2023-08-01
LINK-1394 | User doing AD authentication is never considered to be an external user

charn:
User.is_external checks the AMR claim value and and returns False if AD authentication is used.

AMR (Authentication Method Reference) should be a list of string that are identifiers for authentication methods used in the authentication. Tunnistamo however sets a simple string here, which is assumed in this implementation also.

Refs LINK-1394

### DIFF
--- a/helevents/models.py
+++ b/helevents/models.py
@@ -1,3 +1,4 @@
+import logging
 from functools import reduce
 
 from django.conf import settings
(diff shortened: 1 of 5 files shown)


helevents/models.py
```diff
@@ -18,6 +21,17 @@ class UserModelPermissionMixin:
     @cached_property
     def is_external(self):
         """Check if the user is an external user"""
+
+        if token_amr_claim := getattr(self, "token_amr_claim", None):
```
danipran, line 25: Might make sense to add this as an attribute to the User model.

danipran (approved):
Couple small notes, LGTM otherwise :+1:

PR #645, opened 2023-08-01
Fix memory leak with cache

voneiden:
Use of @cache caused each request to be permanently cached leading to rapid memory usage increase.

Ref LINK-1395

### DIFF
--- a/events/permissions.py
+++ b/events/permissions.py
@@ -27,14 +27,34 @@ class IsReadOnly(BasePermission):
         return request.method in permissions.SAFE_METHODS
 
 
-class UserBelongsToOrganization(BasePermission):
+class UserDataFromRequestMixin:
+    """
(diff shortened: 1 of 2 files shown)


events/permissions.py
```diff
@@ -27,14 +27,28 @@ def has_permission(self, request, view):
         return request.method in permissions.SAFE_METHODS
 
 
-class UserBelongsToOrganization(BasePermission):
+class UserDataFromRequestMixin:
```
danipran, line 30: A short docstring could be useful for this. I'm sure this is the kind of thing that raises the question "but why?" for future software archaeologists, so an answer for this would be nice.

danipran (approved):
LGTM :+1: a short docstring answering the question "why" would be nice. Can merge afterwards.

PR #646, opened 2023-08-01
Optimize related for event?include

voneiden:
Optimize related searches for event include (location and keywords).

Ref LINK-1396

### DIFF
--- a/events/api.py
+++ b/events/api.py
@@ -3248,14 +3248,15 @@ class EventViewSet(
         Event.objects.all()
         .select_related("location", "publisher")
         .prefetch_related(
-            "offers",
-            "keywords",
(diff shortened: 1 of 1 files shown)


events/api.py
```diff
@@ -3248,14 +3248,15 @@ class EventViewSet(
         Event.objects.all()
         .select_related("location", "publisher")
         .prefetch_related(
-            "offers",
-            "keywords",
             "audience",
+            "external_links",
+            "keywords",
+            "offers",
```
danipran, line 3254: If this is meant to be in an alphabetical order, this is in the wrong spot.

danipran (approved):
LGTM :+1:

PR #649, opened 2023-08-03
Remove use of distinct in event listing

voneiden:
The use of distinct is extremely slow in the event listing. ~The need for distinct can be replaced in one-to-many and many-to-many relations by using
`exclude(~Q(relation=something)`. When excluding relations, Django automatically uses SQL EXISTS. Negating the Q statement will essentially create a NOT NOT EXISTS which the database then optimizes into plain EXISTS.~

DISTINCT can be avoided by utilizing EXISTS queries in filters that can create duplicate rows. Additionally some filter queries on m2m relations can be made still significantly faster if we target the through table directly to avoid unnecessary INNER JOIN by django.

### DIFF
--- a/events/api.py
+++ b/events/api.py
@@ -24,7 +24,7 @@ from django.contrib.postgres.search import SearchQuery, TrigramSimilarity
 from django.core.cache import cache
 from django.core.exceptions import PermissionDenied
 from django.db import transaction
-from django.db.models import Count, F, Prefetch, Q
+from django.db.models import Count, Exists, F, OuterRef, Prefetch, Q, QuerySet
 from django.db.models.functions import Greatest
 from django.db.utils import IntegrityError
 from django.http import Http404, HttpResponsePermanentRedirect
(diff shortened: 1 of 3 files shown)


voneiden:
Yritetty tarkistaa seuraavat relaatiot

# Events many-to-many
in_language
images
keywords
audience

# Events one-to-many
videos
external_links
offers
aliases
sub_events

events/api.py
```diff
@@ -2511,13 +2511,39 @@ def _get_queryset_from_cache_many(params, param, cache_name, operator, queryset)
     return queryset
 
 
+def _param_to_keywords(param: str):
+    """
+    Convert comma separated param str into keyword objects with
+    replacements applied.
+    """
```
danipran, line 2518: I think it would make more sense to have this take a list of keyword ids instead of a very specific string format, and call it `find_keyword_replacements` or something. Could also split this into two functions, `find_keyword_replacements` and `_param_to_keywords`, where the latter would call the former.

This is more of a matter of opinion than anything actually important, so feel free to leave it as it is; I just like having my functions with as little responsibility as possible.

events/api.py
```diff
@@ -2511,13 +2511,39 @@ def _get_queryset_from_cache_many(params, param, cache_name, operator, queryset)
     return queryset
 
 
+def _param_to_keywords(param: str):
```
danipran, line 2514: Are there any tests for this? A unit test would be nice now that there's an actual small unit to test (rather than a massive single function that does ***EVERYTHING***)

events/api.py
voneiden: Addressed in c3f71d1d44f1b551acf3dd5491c613547bcd9141

Renamed and changed to accept a list of keyword ids

events/api.py
voneiden: Very good idea, considering there's a little bit of logical branching. Added also `list(set())` to drop out duplicates (like a case where user tries to search both with the new keyword and the replaced keyword.

Addressed in c3f71d1d44f1b551acf3dd5491c613547bcd9141

events/api.py
```diff
@@ -2511,13 +2511,39 @@ def _get_queryset_from_cache_many(params, param, cache_name, operator, queryset)
...
+    """
+    Convert comma separated param str into keyword objects with
+    replacements applied.
+    """
+    keyword_ids = param.split(",")
+    keywords = Keyword.objects.filter(id__in=keyword_ids).select_related("replaced_by")
+    found_all_keywords = len(keyword_ids) == len(keywords)
+    replaced_keywords = [
+        kw_replacement if (kw_replacement := keyword.get_replacement()) else keyword
```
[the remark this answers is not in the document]
voneiden: Comedy gold. 

Addressed in c3f71d1d44f1b551acf3dd5491c613547bcd9141

danipran (approved):
The negated excludes are definitely not intuitive and hurt readability, but I'd say the benefits outweigh the cons. LGTM :+1:

events/api.py
```diff
@@ -2511,13 +2511,47 @@ def _get_queryset_from_cache_many(params, param, cache_name, operator, queryset)
...
+    replaced_keywords = [keyword.get_replacement() or keyword for keyword in keywords]
+    return list(set(replaced_keywords)), found_all_keywords
+
+
+def _filter_events_keyword_or(queryset: QuerySet, keyword_ids: list[str]) -> QuerySet:
+    """
+    Given an Event queryset, apply OR filter on keyword ids
+    :param queryset:
+    :param keyword_ids:
+    :return:
```
voneiden, line 2533: Remove unnecessary param docstrings

events/api.py
```diff
@@ -3119,8 +3152,13 @@ def _filter_event_queryset(queryset, params, srs=None):  # noqa: C901
     # Filter by free offer
     val = params.get("is_free", None)
     if val and val.lower() in ["true", "false"]:
+        # Include events that have at least one free offer
         if val.lower() == "true":
-            queryset = queryset.filter(offers__is_free=True)
+            queryset = queryset.filter(
+                Exists(Offer.objects.filter(event=OuterRef("pk"), is_free=True))
+            )
+            queryset = queryset.exclude(~Q(offers__is_free=True))
```
charn, line 3160: Isn't this line the same as the one above, just with the NOT NOT? :thinking:

voneiden: whoops!

charn (approved):
As far as I can tell, this looks awesome! :cookie: :shipit:

tuomas777 (approved):
LGTM 👍 🌞

PR #650, opened 2023-08-03
LINK-1080: Harrastushaku importer fixes

danipran:
This fixes the "erroneous timetables" error and a DataError ("value too long for type character varying(255)").

`provider_contact_info` (a combination of registration and organizer details) can get quite lengthy with Harrastushaku ([one of the longer ones is over 1024 chars long](https://harrastushaku.fi/view/17094)), so I increased its max length to 10000 from 255.

The erroneous timetables errors were happening were happening in one-day "recurring" events because of timezones. Converting the timestamps to Europe/Helsinki instead of UTC fixes this.

E.g. [if an event starts and ends on 2023-11-23](https://www.harrastushaku.fi/view/17226), the timestamps [from the API](https://www.harrastushaku.fi/api/activity?id=17226) tell that it starts/ends on 2023-11-22T22:00:00Z. This causes errors if using UTC timezone, since then the start/end date is on 2023-11-22, not 2023-11-23.


voneiden:
thou shall add migrations

danipran:
🤡

voneiden (approved):
Looks good to me! :eyes:

charn (approved):
LGTM! :+1: :shipit:

danipran:
SonarCloud is whining about the migration code not being tested. I am not going to write a test for that.

PR #653, opened 2023-08-09
LINK-1080: Add missing imports (kulke keywords, harrastushaku places)

danipran:
Add kulke keywords import and harrastushaku places import in `hourly_tasks.sh` and `daily_tasks.sh`.

Some of the kulke/harrastushaku events (or at least Harrastushaku) require these. Causes `Keyword.DoesNotExist` in kulke & `Place.DoesNotExist` in HH otherwise. Tested with Harrastushaku importer locally; Kulke importer doesn't seem to work locally for me, but the error seems to have a similar root cause.

These weren't in the old production environment's crontab, so makes me wonder whether there's a good reason for it or not.

### DIFF
--- a/helevents/service_tasks/hourly_tasks.sh
+++ b/helevents/service_tasks/hourly_tasks.sh
@@ -11,9 +11,14 @@ fi
 
 echo "--- Starting kulke importer ---"
 
+timeout --preserve-status -s INT 30m python manage.py event_import kulke --keywords
+if [ $? -ne 0 ]; then
+    echo "kulke importer keyword import signaled failure"
+fi
+
 timeout --preserve-status -s INT 30m python manage.py event_import kulke --events
 if [ $? -ne 0 ]; then
-    echo "kulke importer signaled failure"
+    echo "kulke importer event import signaled failure"
 fi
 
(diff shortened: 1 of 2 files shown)


helevents/service_tasks/hourly_tasks.sh
```diff
@@ -11,9 +11,14 @@ fi
 
 echo "--- Starting kulke importer ---"
 
+timeout --preserve-status -s INT 30m python manage.py event_import kulke --keywords
```
voneiden, line 14: Not strictly related to this PR but the total max timeouts in the hourly cron are now 10 + 30 + 30 + 45 + 75 plus a bunch of commands without timeouts. That's like 190 minutes. 

In any case, Kubernetes/Openshift has `activeDeadlineSeconds` which allows setting a max time for the whole (cron)job. Currently not set. Could consider.

voneiden (approved):

helevents/service_tasks/hourly_tasks.sh
danipran: 🤔 good point. Should probably do something about this.

PR #654, opened 2023-08-10
Feature/link 1237 only signed users can signup

harriris-vincit:
### Description

- Adds the new fields `created_at`, `last_modified_at`, `created_by`, `last_modified_by` to the model `SignUp` and to the serializer `SignUpSerializer`
- Automatically sets values for the `created_at` and `last_modified_at` date time fields; `created_by` and `last_modified_by` are set to by the serializers `SignUpSerializer` and `CreateSignUpsSerializer` (value will be the request user / current active user)
- Allows only authenticated users to create, read, update or delete `SignUp` instances; when reading, updating or deleting, the user must additionally be a superuser, an admin of the publishing organization or the user who created the sign-up
- The authentication requirement is mostly handled by the two new permission classes `CanCreateSignUp` and `CanReadUpdateDeleteSignup`
- Completely removes the field `cancellation_code` and the related functionality as permissions are now used instead
### Closes
[LINK-1237](https://helsinkisolutionoffice.atlassian.net/browse/LINK-1237)

[LINK-1237]: https://helsinkisolutionoffice.atlassian.net/browse/LINK-1237?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


voneiden (changes requested):
OPTIONS request must be permitted, otherwise a preflight/CORS error will occur if the API is called from a browser app, right? Otherwise looks good.

registrations/models.py
```diff
@@ -67,40 +67,53 @@ def formfield(self, **kwargs):
...
-class Registration(models.Model):
-    event = models.OneToOneField(
-        Event,
-        on_delete=models.CASCADE,
-        related_name="registration",
-        null=False,
+class CreatedModifiedBaseModel(models.Model):
+    created_at = models.DateTimeField(
+        verbose_name=_("Created at"),
+        null=True,
```
voneiden, line 73: Is there a purpose for nullable auto_now_add field?

registrations/models.py
```diff
@@ -67,40 +67,53 @@ def formfield(self, **kwargs):
...
-    )
-    audience_max_age = models.PositiveSmallIntegerField(
-        verbose_name=_("Maximum recommended age"), blank=True, null=True, db_index=True
-    )
-
-    created_at = models.DateTimeField(verbose_name=_("Created at"), auto_now_add=True)
     last_modified_at = models.DateTimeField(
-        verbose_name=_("Modified at"), null=True, blank=True, auto_now=True
+        verbose_name=_("Modified at"),
+        null=True,
```
voneiden, line 79: I see this was already nullable but at least I would rather set  it non-nullable and set a default in the migration. Unless there was some good reason to do this originally..  A lot of fields in linkedevents are indeed nullable :sweat_smile:

registrations/models.py
harriris-vincit: In this case it's because the SignUp model has existed before these changes and there can be already created instances in the database. For those it is probably most accurate to have a null value and due to this, we allow this field to be nullable.

registrations/models.py
harriris-vincit: Indeed, some of these would be better off being non-nullable. I think the reasoning for allowing this to be nullable is the same as for the created_at timestamp: the existing data where it having a null valuable is perhaps the most appropriate to not give a false default value for them.

registrations/permissions.py
```diff
@@ -1,54 +1,38 @@
...
-        if isinstance(user, ApiKeyUser):
+class CanReadUpdateDeleteSignup(permissions.BasePermission):
+    message: str = _(
+        "Only authenticated users that are admins in the publishing organization or"
+        "that have created the sign-up are allowed to view and edit it"
+    )
+
+    def has_permission(self, request: Request, view: APIView) -> bool:
+        return (
+            request.method in ("GET", "PUT", "PATCH", "DELETE")
```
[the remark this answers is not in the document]
harriris-vincit: Now allowed in the most recent commits.

registrations/models.py
tuomas777: False default values should not really matter as the service is not in production yet, so all those nullability things could and should be fixed now IMO.

registrations/migrations/0017_auto_20230810_1303.py
```diff
@@ -0,0 +1,63 @@
...
+            ),
+        ),
+        migrations.AlterField(
+            model_name="registration",
+            name="created_at",
+            field=models.DateTimeField(
+                auto_now_add=True, null=True, verbose_name="Created at"
+            ),
+        ),
+    ]
```
tuomas777, line 63: It would be nicer for the future if the migration had some human-readable name.

harriris-vincit: Migration name has been changed in the most recent commit.

registrations/models.py
harriris-vincit: The datetimefields have now been made non-nullable.

registrations/models.py
harriris-vincit: This field has now been made non-nullable.

voneiden (approved):
LGTM :cake:

PR #657, opened 2023-08-14
Disable harrastushaku fuzzy keyword search

voneiden:
Full text search + trigram similarity can result in inappropriate keywords to be selected in events.

https://helsinkisolutionoffice.atlassian.net/browse/LINK-1420


tuomas777 (approved):
:shipit:

PR #662, opened 2023-08-17
Rename primary branch from develop to main

voneiden:
Update the various pipelines to use main as the primary branch. Simplify and remove staging build tags

Ref LINK-1438


charn (approved):
LGTM! :+1: :shipit:

PR #663, opened 2023-08-18
Increase the buffer-size for uWSGI

charn:
Increasing the buffer-size will allow bigger requests which will allow e.g. to include big lists of AD-group memberships in a authentication token.

Refs LINK-1448


tuomas777 (approved):
👍

PR #664, opened 2023-08-18
LINK-1447 | Make Image alt_text non_translatable

nicobav:
Reverts`Image` model's `alt_text` to be non-translatable. 
Keeping the field as translatable would mean breaking api change.

Refs LINK-1447


tuomas777 (approved):
LGTM 👍 🚀

PR #666, opened 2023-08-20
Fix api root view name

tuomas777:
Fixed API root view name which was erroneous "Custom Api Root" due to a regression introduced in 8ac8ec78c34aba7ee90159f64b7797a2a5dadd91.

Spotted by @charn 🦅 👁️

Screenshot 2023-08-21 at 0 20 52

Screenshot 2023-08-21 at 0 26 48


voneiden (approved):
very nice

PR #667, opened 2023-08-22
Support changing the URL prefix used by Django

vikoivun:
Support changing the URL prefix for Django by the way of the "mount" and "manage-script-name" directives in uwsgi

URL prefix is read from environment with default being "/".

Also includes a slight optimization for container shutdown speed


tuomas777 (approved):
LGTM 👍

PR #668, opened 2023-08-23
TILA-1071 | Add filters for event max- and min_attendee_capacity

nicobav:
Adds possibility to filter max- and min_attendee_capacity fields through gte, lte lookups. Such as `?max_attendee_capacity__gte=12`.

Also adds ordering capabilities for those fields

Refs LINK-1071

### DIFF
--- a/events/tests/test_event_get.py
+++ b/events/tests/test_event_get.py
@@ -1155,3 +1155,179 @@ def test_get_event_list_verify_registration_filter(
     event3.save()
     get_list_and_assert_events("registration=true", [event, event2])
     get_list_and_assert_events("registration=false", [event3])
+
+
+@pytest.mark.django_db
+def test_get_event_list_max_attendee_capacity_filter_gte(
+    api_client,
+    event,
+    event2,
+    event3,
+):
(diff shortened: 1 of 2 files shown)


events/tests/test_event_get.py
```diff
@@ -1155,3 +1155,111 @@ def test_get_event_list_verify_registration_filter(
...
+    event3,
+):
+    event.maximum_attendee_capacity = 10
+    event.save()
+    event2.maximum_attendee_capacity = 9
+    event2.save()
+    event3.maximum_attendee_capacity = 8
+    event3.save()
+
+    get_list_and_assert_events("max_attendee_capacity_gte=10", [event])
```
danipran, line 1174: This seems to test only the "equal" part of "greater than or equal", which means that this test should pass if, for whatever reason, I changed the `lookup_expr` parameter to `exact`. Applies to the other gte/lte tests as well.

danipran (approved):
Just a tiny issue with the tests, but otherwise LGTM :+1: finally some actual filters 😭

PR #674, opened 2023-08-25
LINK-1351, part 2 | Registration admin permissions

harriris-vincit:
### Description
Makes the following changes to registration permissions:
* Only a registration admin, a superuser or the "created user" is able to perform all CRUD operations on the endpoint `signup`.
* In addition to a regular admin, the registration admin will be able to perform all CRUD operations on the endpoint `registration`.
* A registration user (`RegistrationUserAccess`) is allowed to get a `registration`, but not allowed to do any other CRUD operations on that endpoint.
* A registration user (`RegistrationUserAccess`) is allowed to get or update a `signup`, but not allowed to do any other CRUD operations on that endpoint (updating will be limited to patching just `presence_status` in LINK-1429).

This PR is the second one of two PRs - the first PR (part 1) will bring the Django admin and organization API endpoint changes described in the Jira task.
### Partially closes
[LINK-1351](https://helsinkisolutionoffice.atlassian.net/browse/LINK-1351)

[LINK-1351]: https://helsinkisolutionoffice.atlassian.net/browse/LINK-1351?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/registrations/api.py
+++ b/registrations/api.py
@@ -34,10 +34,7 @@ from registrations.models import (
     SignUp,
     SignUpGroup,
 )
-from registrations.permissions import (
-    CanCreateEditDeleteSignup,
-    RegistrationUserAccessRetrievePermission,
-)
+from registrations.permissions import CanAccessRegistration, CanAccessSignup
 from registrations.serializers import (
     CreateSignUpsSerializer,
     MassEmailSerializer,
@@ -65,9 +62,7 @@ class RegistrationViewSet(
     serializer_class = RegistrationSerializer
     queryset = Registration.objects.all()
 
-    permission_classes = [
-        DataSourceResourceEditPermission & OrganizationUserEditPermission
-    ]
+    permission_classes = [CanAccessRegistration & DataSourceResourceEditPermission]
 
     def perform_create(self, serializer):
         # Check object level permissions for event which has the relevant data_source.
@@ -121,9 +116,7 @@ class RegistrationViewSet(
     @action(
         methods=["post"],
         detail=True,
-        permission_classes=[
-            DataSourceResourceEditPermission & OrganizationUserEditPermission
-        ],
+        permission_classes=[CanAccessRegistration & DataSourceResourceEditPermission],
     )
     def send_message(self, request, pk=None, version=None):
         registration = self.get_object()
@@ -210,11 +203,7 @@ class SignUpViewSet(
 ):
     serializer_class = SignUpSerializer
     queryset = SignUp.objects.all()
-
-    permission_classes = [
-        (CanCreateEditDeleteSignup & DataSourceResourceEditPermission)
-        | RegistrationUserAccessRetrievePermission
-    ]
+    permission_classes = [CanAccessSignup & DataSourceResourceEditPermission]
 
     def create(self, request, *args, **kwargs):
         context = super().get_serializer_context()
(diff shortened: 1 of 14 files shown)


registrations/permissions.py
```diff
@@ -1,46 +1,91 @@
...
-    )
-
+class CanAccessRegistration(UserDataFromRequestMixin, permissions.BasePermission):
     def has_permission(self, request: Request, view: APIView) -> bool:
-        return (
-            request.method
-            in permissions.SAFE_METHODS + ("POST", "PUT", "PATCH", "DELETE")
-            and request.user.is_authenticated
-        )
+        if not request.user.is_authenticated:
```
voneiden, line 12: Well, I've learned something new here. django-cors-headers middleware will NOT pass an OPTIONS request further down the chain, but instead returns a response immediately.

( https://github.com/adamchainz/django-cors-headers/blob/main/src/corsheaders/middleware.py#L71 )

So my earlier comment a few weeks ago about the OPTIONS was not accurate. 

So this will work just fine, although in principle it would block all OPTIONS requests should they ever arrive here since OPTIONS don't carry authentication information. But yeah, goes in the "this should never happen" category.

registrations/permissions.py
```diff
@@ -1,46 +1,91 @@
...
         if isinstance(request.user, ApiKeyUser):
             user_data_source, _ = view.user_data_source_and_organization
             # allow only if the api key matches instance data source
             if obj.data_source != user_data_source:
                 return False
 
+        if request.method == "GET":
+            return (
+                obj.can_be_edited_by(request.user)
+                or obj.registration_user_accesses.filter(
```
voneiden, line 47: Minor risk of privilege escalation if the user is able to manipulate their email address. But maybe nothing to worry about considering the use case? 

Minimizing the risk would probably need something that binds the RegistrationUserAccess permanently to a User object (as soon as such object becomes available).

voneiden: Yeah I don't see any easy way spoofing email with the current authentication methods in tunnistamo.

voneiden (approved):
LGTM. Looks like certain checks (sonarcloud, codecov) don't run for this PR as it's not against `main`.  

Running it manually @ https://github.com/City-of-Helsinki/linkedevents/actions/runs/5999328503/job/16269244043 but dunno if it syncs the results here.

voneiden:
Yeeah sonarcloud lost the plot, ignore.

registrations/api.py
```diff
@@ -304,22 +296,24 @@ def filter_queryset(self, queryset):
...
                         email=user.email
                     ).exists()
-                    or registration.publisher in admin_organizations
+                    or registration.publisher in registration_admin_orgs
                 ):
                     raise DRFPermissionDenied(
                         _(
-                            "Only the admins of the organization {organization} have access rights."
-                        ).format(organization=registration.publisher)
+                            f"Only the admins of the organization {registration.publisher} "
```
charn, line 303: I seem to remember that you cannot mix `gettext` and f-strings. I think the f-string value would be evaluated before `gettext` is run so the sting to translate would already contain the organization name in this case.

harriris-vincit: Good point and definitely applies here -> I changed this back to how it was.

harriris-vincit:
@voneiden & @charn: re-requesting review as the following changes have been made since your previous reviews:
- commits have been squashed into one
- f-string usage in the gettext string "Only the admins of the organization {organization} have access rights." has been removed and the string is now formatted with `format` again
- certain signup group tests have been updated so that the user is a registration admin (otherwise the tests would fail)

voneiden (approved):
Changes look good to me. One trivial conflict in merge.

PR #680, opened 2023-08-30
LINK-1075, part 3 | Update signup group

harriris-vincit:
### Description
Adds the possibility to update `SignUpGroup` instances and `SignUp` instances belonging to the groups. It is only allowed to edit the related `SignUp` instances' data - adding new sign-ups or deleting existing ones from groups needs to be done through the `signup` endpoint.
### Partially closes
[LINK-1075](https://helsinkisolutionoffice.atlassian.net/browse/LINK-1075)

[LINK-1075]: https://helsinkisolutionoffice.atlassian.net/browse/LINK-1075?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/registrations/api.py
+++ b/registrations/api.py
@@ -331,6 +331,7 @@ class SignUpGroupViewSet(
     mixins.CreateModelMixin,
     mixins.RetrieveModelMixin,
     mixins.DestroyModelMixin,
+    mixins.UpdateModelMixin,
     viewsets.GenericViewSet,
 ):
     serializer_class = SignUpGroupSerializer
@@ -342,14 +343,34 @@ class SignUpGroupViewSet(
             return SignUpGroupCreateSerializer
         return super().get_serializer_class()
 
+    def _ensure_shared_request_data(self, data):
+        if self.action == "partial_update":
+            registration_id = data.get(
+                "registration", self.get_object().registration_id
+            )
+        else:
+            registration_id = data.get("registration")
+
+        if not registration_id or not data.get("signups"):
+            return
+
+        for signup_data in data["signups"]:
+            signup_data["registration"] = registration_id
+
     @transaction.atomic
     def create(self, request, *args, **kwargs):
-        data = request.data
-        registration = data.get("registration")
-        if registration and data.get("signups"):
-            for signup_data in data["signups"]:
-                signup_data["registration"] = registration
+        self._ensure_shared_request_data(request.data)
         return super().create(request, *args, **kwargs)
+
+    @transaction.atomic
+    def update(self, request, *args, **kwargs):
+        self._ensure_shared_request_data(request.data)
+        return super().update(request, *args, **kwargs)
+
+    @transaction.atomic
+    def partial_update(self, request, *args, **kwargs):
+        self._ensure_shared_request_data(request.data)
+        return super().partial_update(request, *args, **kwargs)
 
     @transaction.atomic
     def destroy(self, request, *args, **kwargs):
--- a/registrations/serializers.py
+++ b/registrations/serializers.py
@@ -380,6 +380,7 @@ class SignUpGroupSerializer(serializers.ModelSerializer):
 
         if (
             "registration" in validated_data
+            and self.instance
             and self.instance.registration != validated_data["registration"]
         ):
             errors["registration"] = _(
--- a/registrations/serializers.py
+++ b/registrations/serializers.py
@@ -34,6 +34,7 @@ def validate_registration_enrolment_times(registration):
 
 class SignUpSerializer(serializers.ModelSerializer):
     view_name = "signup"
+    id = serializers.IntegerField(required=False)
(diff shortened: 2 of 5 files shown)


nicobav (changes requested):
Would be nice to get this rebased against [feature/LINK-1075-1_create-signup-group-endpoint](https://github.com/City-of-Helsinki/linkedevents/tree/feature/LINK-1075-1_create-signup-group-endpoint) branch to get ridoff the merge commits etc.

harriris-vincit:
> Would be nice to get this rebased against [feature/LINK-1075-1_create-signup-group-endpoint](https://github.com/City-of-Helsinki/linkedevents/tree/feature/LINK-1075-1_create-signup-group-endpoint) branch to get ridoff the merge commits etc.

@nicobav done - I've picked only the relevant commits and squashed them into a single one.

registrations/tests/test_signup_post.py
```diff
@@ -24,6 +29,7 @@ def create_signups(api_client, signups_data):
 
 def assert_create_signups(api_client, signups_data):
     response = create_signups(api_client, signups_data)
+    print(response.json())
```
charn, line 32: This should be removed :eyes:

registrations/serializers.py
```diff
@@ -360,7 +363,50 @@ class Meta:
...
+        if (
+            "registration" in validated_data
+            and instance.registration != validated_data["registration"]
+        ):
+            errors["registration"] = _(
+                "You may not change the registration of an existing object."
+            )
+
+        if errors:
+            raise serializers.ValidationError(errors)
```
charn, line 401: You could have this kind of logic in `validate_registration` or `validate`. :thinking:

registrations/api.py
```diff
@@ -383,15 +384,33 @@ def get_serializer_class(self):
             return SignUpGroupCreateSerializer
         return super().get_serializer_class()
 
+    def _ensure_shared_request_data(self, data):
+        if self.action in ("update", "partial_update"):
+            registration_id = (
+                data.get("registration") or self.get_object().registration_id
```
nicobav, line 390: is there some particual reason not to use dict's default value? `data.get("registration", self.get_object().registration_id)` would make my pythonic eyes more sense. 
Obivously on creation it will blow up since the `get_object` returns `None` but that has been tackled in the above so should be the case.

nicobav (commented):
Looks nice, adding my debatable opinions about the some line breaks and use of dictionary.

registrations/api.py
harriris-vincit: Here I was thinking the case where `registration` would be an empty string. In that case the usage of `or` would lead to the `self.get_object().registration_id` being used and thus a somewhat unnecessary validation error would be avoided. If it was defined as the dict's default value, it would not be used as the empty string would be considered a usable value.

Thinking about this again I realized that there is going to be a validation error regardless of this as the `self.get_object().registration_id` is only going to be passed to the "children" of the group (sign-ups). The parent/group is going to have the empty string in its data regardless of this code and a validation error will occur.

The code could be easily improved to "fix" also the parent validation error, but perhaps it is best to raise the error and not try to alter the request data for a smoother API experience. So I'll commit this change soon.

harriris-vincit: Changed - will re-request review soon

nicobav (approved):
Looks A OK to me 👍 

(Left some minor comments for you sir to consider before merge)

registrations/serializers.py
```diff
@@ -362,7 +365,55 @@ class Meta:
...
+
+        return validated_data
+
+    def _update_signups(self, instance, validated_signups_data):
+        for signup_data in validated_signups_data:
+            if not signup_data.get("id"):
+                continue
+            if not self.partial:
```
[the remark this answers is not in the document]
harriris-vincit: Added

registrations/serializers.py
```diff
@@ -362,7 +365,55 @@ class Meta:
...
+                signup_data["signup_group"] = instance
+            signup = SignUp.objects.get(pk=signup_data["id"])
```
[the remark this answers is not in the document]
harriris-vincit: Added here as well

registrations/serializers.py
```diff
@@ -362,7 +365,57 @@ class Meta:
...
+        return fields
+
+    def validate(self, data):
+        validated_data = super().validate(data)
+
+        errors = {}
+
+        if (
+            "registration" in validated_data
+            and self.instance.registration != validated_data["registration"]
```
charn, line 383: To be on the safe side I think you should also check that `self.instance` exists i.e. is not `None`.
```suggestion
            and self.instance and self.instance.registration != validated_data["registration"]
```

PR #681, opened 2023-08-30
LINK-1075, part 4 | List signup groups

harriris-vincit:
### Description
Adds the `list` action and filtering by certain parameters to the `signup_group` endpoint. The filtering parameters and logic are shared between the `signup_group` and `signup` endpoints.
### Partially closes
[LINK-1075](https://helsinkisolutionoffice.atlassian.net/browse/LINK-1075)

[LINK-1075]: https://helsinkisolutionoffice.atlassian.net/browse/LINK-1075?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/registrations/api.py
+++ b/registrations/api.py
@@ -7,6 +7,7 @@ from django.contrib.auth.models import AnonymousUser
 from django.db import transaction
 from django.db.models import ProtectedError, Q, Value
 from django.db.models.functions import Concat
+from django.utils.functional import cached_property
 from django.utils.translation import gettext as _
 from rest_framework import mixins, status, viewsets
 from rest_framework.decorators import action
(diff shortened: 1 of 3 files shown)


registrations/api.py
```diff
@@ -384,6 +398,16 @@ def create(self, request, *args, **kwargs):
                 signup_data["registration"] = registration
         return super().create(request, *args, **kwargs)
 
+    def filter_queryset(self, queryset):
+        request = self.request
+
+        if registration_param := request.query_params.get("registration", None):
```
nicobav: then here just one condition to return fast and avoid nesting

```
if registration_param:
    return self.filter_by_registration(queryset, registration_param)
    
return self.filter_by_user(queryset, self.request.user)
```

registrations/tests/test_signup_group_get.py
```diff
@@ -143,3 +166,119 @@ def test_api_key_from_wrong_data_source_cannot_get_signup(
...
+    default_org.admin_users.remove(user)
+
+    api_client.force_authenticate(user)
+
+    response = get_list(api_client, f"registration={registration.id}")
+    assert response.status_code == status.HTTP_403_FORBIDDEN
+
+
+@pytest.mark.django_db
+def test__get_all_signup_groups_to_which_user_has_admin_role(
```
nicobav: test name got double underscore `test__....` could be changed to `test_`

registrations/tests/test_signup_group_get.py
```diff
@@ -143,3 +166,119 @@ def test_api_key_from_wrong_data_source_cannot_get_signup(
...
+    default_org.regular_users.add(user)
+    default_org.admin_users.remove(user)
+
+    api_client.force_authenticate(user)
+
+    get_list_and_assert_signup_groups(api_client, "", [])
+
+
+@pytest.mark.django_db
+def test__user_from_other_organization_cannot_get_signup_group_list(
```
nicobav: double underscore here as well

registrations/tests/test_signup_group_get.py
```diff
@@ -143,3 +166,119 @@ def test_api_key_from_wrong_data_source_cannot_get_signup(
...
+    SignUpGroupFactory(registration=registration)
+
+    api_client.force_authenticate(user2)
+
+    response = get_list(api_client, f"registration={registration.id}")
+    assert response.status_code == status.HTTP_403_FORBIDDEN
+
+
+@pytest.mark.django_db
+def test__cannot_get_signup_groups_of_nonexistent_registration(
```
nicobav: and here

registrations/tests/test_signup_group_get.py
```diff
@@ -143,3 +166,119 @@ def test_api_key_from_wrong_data_source_cannot_get_signup(
...
+
+    api_client.force_authenticate(user2)
+
+    response = get_list(api_client, "registration=not-exist")
+    assert response.status_code == status.HTTP_400_BAD_REQUEST
+    assert response.data["detail"] == "Registration with id not-exist doesn't exist."
+
+
+@pytest.mark.django_db
+def test__api_key_with_organization_can_get_signup_group_list(
```
nicobav: ☝️

registrations/tests/test_signup_group_get.py
```diff
@@ -143,3 +166,119 @@ def test_api_key_from_wrong_data_source_cannot_get_signup(
...
+
+    api_client.credentials(apikey=data_source.api_key)
+
+    get_list_and_assert_signup_groups(
+        api_client, f"registration={registration.id}", (signup_group0, signup_group1)
+    )
+
+
+@pytest.mark.django_db
+def test__api_key_with_wrong_organization_cannot_get_signup_group_list(
```
nicobav: here too

registrations/api.py
```diff
@@ -220,8 +221,124 @@ def send_invitation(self, request, pk=None, version=None):
...
+        relation = self._list_filter_relation_name
+        filters = {
+            f"{relation}first_last_name__icontains": text_param,
+            f"{relation}last_first_name__icontains": text_param,
+            f"{relation}email__icontains": text_param,
+            f"{relation}extra_info__icontains": text_param,
+            f"{relation}membership_number__icontains": text_param,
+            f"{relation}phone_number__icontains": text_param,
+        }
+        q_set = Q()
```
nicobav, line 309: there could be one empty line above this to let the code breath a bit => improves readability

registrations/api.py
```diff
@@ -220,8 +221,124 @@ def send_invitation(self, request, pk=None, version=None):
...
+        for item in filters.items():
+            q_set |= Q(item)
+        return q_set
```
nicobav, line 312: would suggest adding one empty line above of this as well.

registrations/api.py
```diff
@@ -220,8 +221,124 @@ def send_invitation(self, request, pk=None, version=None):
...
+
+    def filter_by_text(self, queryset, text_param):
+        return queryset.annotate(**self._build_text_annotations()).filter(
+            self._build_text_filter(text_param)
+        )
+
+    def _build_attendee_status_filter(self, vals):
+        relation = self._list_filter_relation_name
+        filters = {f"{relation}attendee_status__in": vals}
+        return Q(**filters)
```
nicobav, line 322: Well, I mentioned adding line breaks, could here as well 🤓

registrations/api.py
```diff
@@ -384,6 +398,16 @@ def create(self, request, *args, **kwargs):
                 signup_data["registration"] = registration
         return super().create(request, *args, **kwargs)
 
+    def filter_queryset(self, queryset):
+        request = self.request
```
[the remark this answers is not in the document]
nicobav: hmph, I see this is changed in next commit... would be nice squash those commits into one. 

So my upper comment is outdated anyhow 👍

registrations/api.py
nicobav: seems to be rewritten elsewhere in next commit. Nevermind.

nicobav (changes requested):
Looks good overall 👍 Some minor readability improvement notes.

Some commit squashing would be nice to get the code churn to minimum but this is fine.

registrations/api.py
harriris-vincit: Done

registrations/api.py
harriris-vincit: Done

harriris-vincit:
> Looks good overall 👍 Some minor readability improvement notes.
> 
> Some commit squashing would be nice to get the code churn to minimum but this is fine.

Commits have been squashed for this PR too now.

registrations/api.py
```diff
@@ -220,8 +220,53 @@ def send_invitation(self, request, pk=None, version=None):
...
+                or registration.publisher in admin_organizations
+            ):
+                raise DRFPermissionDenied(
+                    _(
+                        "Only the admins of the organization {organization} have access rights."
+                    ).format(organization=registration.publisher)
+                )
+
+            registrations.append(registration)
+        return queryset.filter(registration__in=registrations)
```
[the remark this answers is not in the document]
harriris-vincit: Done

registrations/api.py
```diff
@@ -220,8 +221,131 @@ def send_invitation(self, request, pk=None, version=None):
 register_view(RegistrationUserAccessViewSet, "registration_user_access")
 
 
+class SignUpListMixin:
```
charn, line 224: Instead of having these filters as these custom methods you could consider refactoring this logic into a `django_filters.rest_framework.FilterSet`

tuomas777: I second this. We should move toward automatic documentation generation, and therefore having all new features support that is highly desirable.

harriris-vincit: This one's done now - re-requesting reviews soon.

registrations/api.py
```diff
@@ -220,14 +221,165 @@ def send_invitation(self, request, pk=None, version=None):
...
     UserDataSourceAndOrganizationMixin,
     viewsets.ModelViewSet,
 ):
     serializer_class = SignUpSerializer
     queryset = SignUp.objects.all()
+    filter_backends = [SignUpFilterBackend]
+    filterset_class = SignUpFilter
     permission_classes = [CanAccessSignup & DataSourceResourceEditPermission]
 
+    def get_filterset_kwargs(self):
```
nicobav, line 378: Why we need this? I'm probably missing something 🤔

registrations/api.py
```diff
@@ -220,14 +221,165 @@ def send_invitation(self, request, pk=None, version=None):
 register_view(RegistrationUserAccessViewSet, "registration_user_access")
 
 
+class SignUpFilterBackend(django_filters.rest_framework.DjangoFilterBackend):
+    def get_filterset_kwargs(self, request, queryset, view):
+        kwargs = super().get_filterset_kwargs(request, queryset, view)
+
+        if hasattr(view, "get_filterset_kwargs"):
```
nicobav, line 228: this because we have `get_filterset_kwargs` defined in the `SignUpViewset`?

nicobav (commented):
Looks overall good.
Just having hard time to understand the use of the `SignUpFilterBackend` and its `get_filterset_kwargs`

registrations/api.py
harriris-vincit: @nicobav I predicted that this would be questioned :-) This is because we are passing `action` from the View as a kwarg to the Filterset and the reason for that is that I wanted to keep business logic untouched after this refactoring. The existing code gave a 403 error if any other action that `list` was done and if the user wasn't allowed to do the action on the objects. It did that by checking the `action` firstly so I kept it as it was (as otherwise the endpoint would give a 404 as the object would not be included in the queryset). The 403 is perhaps semantically better to the user than 404 in this case.

charn: I think you can also do this purely in `SignUpFilterBackend.get_filterset_kwargs` since it gets `view` as an argument and that is the instance of the view, so in `SignUpFilterBackend` you could just have:

```
    def get_filterset_kwargs(self, request, queryset, view):
        kwargs = super().get_filterset_kwargs(request, queryset, view)
        kwargs["action"] = view.action

        return kwargs
```

harriris-vincit: Good point -> now changed to work like that

PR #684, opened 2023-08-30
LINK-1451 | Add env variable for noreply email address

harriris-vincit:
### Description
- Makes it possible to define the `DEFAULT_FROM_EMAIL` using an environment variable. The main purpose of the variable will be to define a "noreply" address.
- Makes code where the noreply address has been determined individually use a common utility function for getting the address.
- The utility function will fall back to `Site.objects.get_current().domain` if `settings` do not have an address value.
### Closes
[LINK-1451](https://helsinkisolutionoffice.atlassian.net/browse/LINK-1451)

[LINK-1451]: https://helsinkisolutionoffice.atlassian.net/browse/LINK-1451?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/linkedevents/settings.py
+++ b/linkedevents/settings.py
@@ -52,6 +52,7 @@ env = environ.Env(
     COOKIE_PREFIX=(str, "linkedevents"),
     DATABASE_URL=(str, "postgis:///linkedevents"),
     DEBUG=(bool, False),
+    DEFAULT_FROM_EMAIL=(str, "noreply@linkedevents.hel.fi"),
     ELASTICSEARCH_URL=(str, None),
     ELIS_EVENT_API_URL=(
         str,
@@ -591,6 +592,9 @@ FULLTEXT_SEARCH_LANGUAGES = {"fi": "finnish", "sv": "swedish", "en": "english"}
 
 # Email address used to send feedback forms
 SUPPORT_EMAIL = env("SUPPORT_EMAIL")
+
+# Email address used as default "from" email
+DEFAULT_FROM_EMAIL = env("DEFAULT_FROM_EMAIL")
 
 SESSION_SERIALIZER = "django.contrib.sessions.serializers.PickleSerializer"
 
--- a/registrations/utils.py
+++ b/registrations/utils.py
@@ -65,4 +65,6 @@ def send_mass_html_mail(
 
 
 def get_email_noreply_address():
-    return "noreply@%s" % Site.objects.get_current().domain
+    return (
+        settings.DEFAULT_FROM_EMAIL or "noreply@%s" % Site.objects.get_current().domain
+    )
--- a/events/models.py
+++ b/events/models.py
@@ -28,7 +28,6 @@ from django.contrib.gis.db import models
 from django.contrib.postgres.fields import HStoreField
 from django.contrib.postgres.indexes import Index
 from django.contrib.postgres.search import SearchVectorField
-from django.contrib.sites.models import Site
 from django.core.mail import send_mail
 from django.db import transaction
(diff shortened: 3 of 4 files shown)


jorilindell (commented):
Code looks good. NOREPLY_ADDRESS needs to be added to https://dev.azure.com/City-of-Helsinki/linkedevents/_git/linkedevents-pipelines?path=/devops/linkedevents-api-config-map.yml to get it to work on Platta environments. 

After that it can be defined in environment specific library. e.g. https://dev.azure.com/City-of-Helsinki/linkedevents/_library?itemType=VariableGroups&view=VariableGroupView&variableGroupId=2822&path=linkedevents-api-development

linkedevents/settings.py
```diff
@@ -600,3 +601,6 @@ def dummy_haystack_connection_for_lang(language_code):
 # Enkora course importer
 ENKORA_API_USER = env("ENKORA_API_USER")
 ENKORA_API_PASSWORD = env("ENKORA_API_PASSWORD")
+
+
+NOREPLY_ADDRESS = env("NOREPLY_ADDRESS")
```
charn, line 606: Django already has `DEFAULT_FROM_EMAIL`. Using that might be a good alternative? https://docs.djangoproject.com/en/4.2/ref/settings/#std-setting-DEFAULT_FROM_EMAIL

registrations/utils.py
```diff
@@ -53,4 +53,7 @@ def send_mass_html_mail(
 
 
 def get_email_noreply_address():
-    return "noreply@%s" % Site.objects.get_current().domain
+    return (
+        getattr(settings, "NOREPLY_ADDRESS", None)
```
charn, line 57: No need to have getattr here since the setting exists.

harriris-vincit: @charn This is now changed in https://github.com/City-of-Helsinki/linkedevents/pull/684/commits/fb464f1fa591da600215be5f39665142999e6801.

linkedevents/settings.py
harriris-vincit: Yes, because it's not empty by default it should be ensured that it's set for each environment with the correct value while a variable like `NOREPLY_ADDRESS` can be used only when necessary. And do we want in all cases to have the default "from" email to be the "noreply" address as would happen if the Django variable would be used?

Personally I would lean towards sticking with the `NOREPLY_ADDRESS` variable, but perhaps we need a final word here. What does @tuomas777  think of this?

tuomas777: Discussed this with @harriris-vincit and we decided to go with Django's `DEFAULT_FROM_EMAIL` mostly because it is used for this purpose in several other Helsinki's Django projects already.

harriris-vincit:
> Code looks good. NOREPLY_ADDRESS needs to be added to https://dev.azure.com/City-of-Helsinki/linkedevents/_git/linkedevents-pipelines?path=/devops/linkedevents-api-config-map.yml to get it to work on Platta environments.
> 
> After that it can be defined in environment specific library. e.g. https://dev.azure.com/City-of-Helsinki/linkedevents/_library?itemType=VariableGroups&view=VariableGroupView&variableGroupId=2822&path=linkedevents-api-development

Variable was renamed to "DEFAULT_FROM_EMAIL" (Django's existing variable) and using that name, was added to the configmap and defined in the "dev" environments library as "noreply@linkedevents.dev.hel.ninja".

charn (approved):
LGTM! :+1: :shipit:

PR #687, opened 2023-08-31
Fix user_created_notification failing on empty emails

charn:
If a superuser with an empty exists. `user_created_notification` will fail with `AnymailInvalidAddress`.

### DIFF
--- a/events/signals.py
+++ b/events/signals.py
@@ -40,10 +40,10 @@ def user_created_notification(sender, instance, created, **kwargs):
     if created:
         user_model = get_user_model()
         recipient_list = [
-            item[0]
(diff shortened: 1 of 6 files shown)


voneiden (approved):
looks m'kay to me

events/signals.py
```diff
@@ -44,6 +44,7 @@ def user_created_notification(sender, instance, created, **kwargs):
             for item in user_model.objects.filter(is_superuser=True).values_list(
                 "email"
             )
```
jorilindell, line 46: This is not really related to this PR. But at some point there was problems when trying to send emails to admin users and some of them didn't have email address. I'm not really sure is that fixed in code level or was it fixed by adding dummy email for each superuser.

PR #691, opened 2023-09-05
Make division filtering use exists instead of distinct

voneiden:
Division filtering ends up doing DISTINCT which ends up using on disk sort which kills the DB perf. Replace with EXISTS to avoid silly sorting


charn (approved):
LGTM :shipit:

PR #692, opened 2023-09-05
LINK-1527 | Log uWSGI errors to sentry

charn:
This PR adds uwsgi-sentry plugin for uWSGI so that errors from uWSGI can be logged into sentry. This PR also adds `harakiri` to uWSGI configuration, which should take care of actually killing a long running uWSGI process.


danipran (approved):
LGTM :+1:

PR #694, opened 2023-09-05
Bump pyjwt, cryptography, social-auth-core

nicobav:
Needed to bump social-auth-core (🤞 ) to get latest pyjwt. Previously used version of pyjwt had some security issues.

Refs RAT-18

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -211,7 +211,7 @@ pygments==2.15.1
     # via ipython
 pyjwkest==1.4.2
     # via drf-oidc-auth
-pyjwt==1.7.1
+pyjwt==2.8.0
     # via social-auth-core
 pyparsing==3.0.9
     # via rdflib
@@ -274,11 +274,10 @@ six==1.16.0
     #   pyjwkest
     #   python-dateutil
     #   social-auth-app-django
-    #   social-auth-core
     #   url-normalize
 social-auth-app-django==4.0.0
     # via -r requirements.in
-social-auth-core==4.0.2
+social-auth-core==4.0.3
     # via
     #   -r requirements.in
     #   social-auth-app-django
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -147,8 +147,6 @@ icalendar==5.0.4
     # via -r requirements.in
 idna==3.4
     # via requests
-importlib-metadata==6.0.0
```
charn, line 150: This should not get removed. You've probably run `pip-compile` with Python=<3.9.

requirements.txt
```diff
@@ -312,8 +309,6 @@ webencodings==0.5.1
     # via bleach
 whitenoise==6.4.0
     # via -r requirements.in
-zipp==3.15.0
```
charn, line 315: This should also not get removed.

requirements.txt
```diff
@@ -274,11 +272,10 @@ six==1.16.0
     #   pyjwkest
     #   python-dateutil
     #   social-auth-app-django
-    #   social-auth-core
     #   url-normalize
 social-auth-app-django==4.0.0
     # via -r requirements.in
-social-auth-core==4.0.2
+social-auth-core==4.0.3
```
charn, line 278: I wonder if we could bump this even more. Though I guess this is enough since it has the support for the new pyjwt.

nicobav: Perhaps we could. I just wanted to play sure. What we get if we upgrade higher, that I don't know.

nicobav: So if we stay for now at this 4.0.3?

requirements.txt
nicobav: Nicely catched 🎣

requirements.txt
charn: Fine by me! :+1:

charn (approved):
LGTM! :+1: :shipit:

PR #696, opened 2023-09-07
LINK-1503 | Azure pipeline: use production pool for production

charn:
Changing to production pool should take care of some resources not being available in the pipeline.


voneiden (approved):
Learned something new today

PR #698, opened 2023-09-07
LINK-1531: Optimize place and keyword listings

voneiden:
Fix duplicate queries for both place and keyword endpoints, move multiple select_related fields to prefetch_related because they are mostly null rows anyway, so joining is inefficient (especially with large page offsets), and add tests to keep an eye on the query numbers for default listings.

Ref https://helsinkisolutionoffice.atlassian.net/browse/LINK-1531


danipran (approved):
LGTM :+1:

PR #701, opened 2023-09-11
LINK-1490 | Add possibility to send emails to specific registration signup groups

harriris-vincit:
### Description
Makes it possible to send messages to specific signup groups (to their responsible signups).

Also changes the individual signup messaging so that if any individual signups - that is, signups that don't belong to any group - are given to the endpoint, messages will be sent to them regardless of if they are responsible for a group or not (unless they were already processed through the group messaging).
### Closes
[LINK-1490](https://helsinkisolutionoffice.atlassian.net/browse/LINK-1490)

[LINK-1490]: https://helsinkisolutionoffice.atlassian.net/browse/LINK-1490?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/registrations/serializers.py
+++ b/registrations/serializers.py
@@ -500,6 +500,13 @@ class SeatReservationCodeSerializer(serializers.ModelSerializer):
         model = SeatReservationCode
 
 
+class MassEmailSignupGroupsField(serializers.PrimaryKeyRelatedField):
+    def get_queryset(self):
+        registration = self.context["request"].data["registration"]
+
+        return registration.signup_groups.exclude(signups__email=None)
+
+
 class MassEmailSignupsField(serializers.PrimaryKeyRelatedField):
     def get_queryset(self):
         registration = self.context["request"].data["registration"]
@@ -510,6 +517,10 @@ class MassEmailSignupsField(serializers.PrimaryKeyRelatedField):
 class MassEmailSerializer(serializers.Serializer):
     subject = serializers.CharField()
     body = serializers.CharField()
+    signup_groups = MassEmailSignupGroupsField(
+        many=True,
+        required=False,
+    )
     signups = MassEmailSignupsField(
         many=True,
         required=False,
--- a/registrations/serializers.py
+++ b/registrations/serializers.py
@@ -504,7 +504,7 @@ class MassEmailSignupGroupsField(serializers.PrimaryKeyRelatedField):
     def get_queryset(self):
         registration = self.context["request"].data["registration"]
 
-        return registration.signup_groups.exclude(signups__email=None)
+        return registration.signup_groups.all()
 
 
 class MassEmailSignupsField(serializers.PrimaryKeyRelatedField):
--- a/registrations/tests/test_send_message.py
+++ b/registrations/tests/test_send_message.py
@@ -66,7 +66,7 @@ def test_email_is_sent_to_all_if_no_groups_or_signups_given(
 
     # Individual
     third_signup = SignUpFactory(registration=registration, email="test3@test.com")
-    fourth_signup = SignUpFactory(registration=registration, email="test3@test.com")
+    fourth_signup = SignUpFactory(registration=registration, email="test4@test.com")
 
     api_client.force_authenticate(user)
     send_message_data = {"subject": "Message subject", "body": "Message body"}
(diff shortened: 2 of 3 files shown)


voneiden (approved):
Looks solid to me.

registrations/serializers.py
```diff
@@ -500,6 +500,13 @@ class Meta:
         model = SeatReservationCode
 
 
+class MassEmailSignupGroupsField(serializers.PrimaryKeyRelatedField):
+    def get_queryset(self):
+        registration = self.context["request"].data["registration"]
+
+        return registration.signup_groups.exclude(signups__email=None)
```
charn, line 507: If a single signup inside a signup group has email as `None` the signup group will be excluded, I guess that is intended?

registrations/tests/test_send_message.py
```diff
@@ -60,77 +63,114 @@ def test_email_is_sent_to_responsible_signups(api_client, registration, user):
...
     assert_send_message(
         api_client,
         registration.id,
         send_message_data,
-        [first_signup.email, third_signup.email, fourth_signup.email],
+        [
+            responsible_signup.email,
+            non_responsible_signup.email,
+            individual_signup.email,
+        ],
```
danipran, line 173: The test is named `test_email_is_sent_to_selected_signups_only`; however, there seems to be a lot more going than that (multiple signup groups, responsible/non-responsible/individual signups). What's happening here? Is the name correct?

registrations/tests/test_send_message.py
```diff
@@ -60,77 +63,114 @@ def test_email_is_sent_to_responsible_signups(api_client, registration, user):
...
+        "body": "Message body",
+        "signup_groups": [signup_group1.pk],
+    }
 
     assert_send_message(
-        api_client,
-        registration.id,
-        send_message_data,
-        [first_signup.email, second_signup.email, third_signup.email],
+        api_client, registration.id, send_message_data, [third_signup.email]
```
danipran, line 121: The name of the test is `test_email_is_sent_to_selected_signup_groups_only`; shouldn't the email be sent to `test4@test.com` as well then?

registrations/tests/test_send_message.py
```diff
@@ -60,77 +63,114 @@ def test_email_is_sent_to_responsible_signups(api_client, registration, user):
         responsible_for_group=True,
         email="test2@test.com",
     )
-    SignUpFactory(
-        signup_group=signup_group, registration=registration, email="test3@test.com"
-    )
+
+    # Individual
+    third_signup = SignUpFactory(registration=registration, email="test3@test.com")
+    fourth_signup = SignUpFactory(registration=registration, email="test3@test.com")
```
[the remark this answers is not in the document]
harriris-vincit: Not supposed to be the same so an "Oopsie 💩" has occurred here as well -> fixing soon. "Individual" here means signups that don't belong to any group.

registrations/tests/test_send_message.py
```diff
@@ -60,77 +63,114 @@ def test_email_is_sent_to_responsible_signups(api_client, registration, user):
...
 
 @pytest.mark.django_db
-def test_email_is_sent_to_group_signups_without_responsible_signups(
-    api_client, registration, user
-):
-    signup_group = SignUpGroupFactory(registration=registration)
-    first_signup = SignUpFactory(
-        signup_group=signup_group, registration=registration, email="test@test.com"
+def test_email_is_sent_to_selected_signup_groups_only(api_client, registration, user):
+    signup_group0 = SignUpGroupFactory(registration=registration)
```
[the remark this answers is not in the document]
harriris-vincit: Starting from 0 makes perfect sense to us androids. But I can compute why you prefer the other option.

(I'll change this to use the "first_signup", "second_signup", etc, naming convention that I've used elsewhere).

registrations/tests/test_send_message.py
harriris-vincit: The test name is correct because when giving `signups` in the request body, only those should get the emails. There are multiple other ones here to test that they don't also get the email.

registrations/tests/test_send_message.py
harriris-vincit: Improved the test case name in the commit 34128889b16ed30ae20ff09e74526edd7565e798. It was otherwise correct that `test4@test.com` shouldn't get the email as it isn't a `responsible_for_group` signup (and since the group already has the third signup marked as the responsible one).

registrations/serializers.py
harriris-vincit: This would have been wrong actually => now changed so that all groups for the registration are included, but later the responsible signups are filtered like this.

danipran (approved):
Thanks for the changes and clarifications!

PR #702, opened 2023-09-11
LINK-1498 | No draft notifications for events created with an API key

charn:
Draft events which are created using an API key shouldn't send any notifications for organization admin.

### DIFF
--- a/events/models.py
+++ b/events/models.py
@@ -1079,7 +1079,11 @@ class Event(MPTTModel, BaseModel, SchemalessFieldMixin, ReplacedByMixin):
             old_deleted is False and self.deleted is True
         ):
             self.send_deleted_notification()
-        if created and self.publication_status == PublicationStatus.DRAFT:
+        if (
+            created
+            and self.publication_status == PublicationStatus.DRAFT
+            and not self.is_created_with_apikey
+        ):
             self.send_draft_posted_notification()
 
     def __str__(self):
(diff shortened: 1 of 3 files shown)


events/models.py
```diff
@@ -1165,12 +1165,26 @@ def send_published_notification(self, request=None):
         )
 
     def send_draft_posted_notification(self, request=None):
+        if self.is_created_with_apikey:
+            # Events created with an API key shouldn't send draft notifications.
+            return
```
nicobav, line 1170: Matter of a taste; I think this would benefit a bit of air in a form of line break after returning [None] to make the rest of the logic more "separate"

nicobav (approved):
LGTM 🚀 

(Had one oxygenic comment to be disregarded 🧐 )

events/tests/test_notifications.py
```diff
@@ -130,6 +132,28 @@ def test_draft_posted(draft_posted_notification_template, user, draft_event):
     )
 
 
+@pytest.mark.parametrize("uses_api_key, expected", [(True, 0), (False, 1)])
+@pytest.mark.django_db
+def test_draft_posted_api_key(uses_api_key, expected, user, organization):
```
danipran, line 137: Please use a more descriptive name or split the test into two tests.

events/tests/test_notifications.py
```diff
@@ -130,6 +132,28 @@ def test_draft_posted(draft_posted_notification_template, user, draft_event):
...
+    """No draft notification should be sent when an API key is used."""
```
danipran, line 138: This description for the test is incorrect.

danipran (changes requested):

events/tests/test_notifications.py
danipran: I highly recommend the latter.

PR #704, opened 2023-09-11
LINK-1548 | Log registrations exceptions to Sentry

harriris-vincit:
### Description
Ensures that some caught exceptions in the `registrations` app are logged into Sentry.
### Closes
[LINK-1548](https://helsinkisolutionoffice.atlassian.net/browse/LINK-1548)

[LINK-1548]: https://helsinkisolutionoffice.atlassian.net/browse/LINK-1548?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/registrations/models.py
+++ b/registrations/models.py
@@ -1,3 +1,4 @@
+import logging
 from datetime import timedelta
 from smtplib import SMTPException
 from uuid import uuid4
@@ -25,6 +26,8 @@ from registrations.utils import (
 )
 
 User = settings.AUTH_USER_MODEL
+
+logger = logging.getLogger(__name__)
 
 
 class SignUpNotificationType:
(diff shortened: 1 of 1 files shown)


registrations/api.py
```diff
@@ -166,7 +164,7 @@ def send_message(self, request, pk=None, version=None):
         try:
             send_mass_html_mail(messages, fail_silently=False)
         except SMTPException as e:
-            logger.exception("Couldn't send mass HTML email.")
+            capture_exception(e)
```
charn, line 167: Since we have set up sentry logging in `settings.py`, using `logger.exception` should be enough to send errors to sentry.

harriris-vincit: Yes, that seems to be the default behavior of the Sentry SDK: https://docs.sentry.io/platforms/python/guides/logging/. Learned something new here :-) Now changed to using `logger.exception`

charn (approved):
LGTM! :+1: :shipit:

PR #708, opened 2023-09-14
LINK-1011 | Helsinki Profile GDPR Endpoints

harriris-vincit:
### Description
Adds the [Helsinki Profile GDPR API ](https://github.com/City-of-Helsinki/helsinki-profile-gdpr-api/) to the project, and makes it possible for a user to query and delete their own GDPR data. It has been determined that at this point the data stored in the `helevents_user`, `registrations_signup` and `registrations_signup_group` tables are the ones to be subject to querying and deleting like this.
### Closes
[LINK-1011](https://helsinkisolutionoffice.atlassian.net/browse/LINK-1011)

[LINK-1011]: https://helsinkisolutionoffice.atlassian.net/browse/LINK-1011?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/helevents/tests/test_gdpr_api_delete.py
+++ b/helevents/tests/test_gdpr_api_delete.py
@@ -1,0 +1,123 @@
+from unittest.mock import patch
+from uuid import UUID
+
+import pytest
+import requests_mock
+from django.urls import reverse
+from helusers.settings import api_token_auth_settings
+from rest_framework import status
+
+from events.tests.conftest import APIClient
+from helevents.models import User
+from helevents.tests.conftest import get_api_token_for_user_with_scopes
+from helevents.tests.factories import UserFactory
+from registrations.models import SignUp, SignUpGroup
+from registrations.tests.factories import (
+    RegistrationFactory,
+    SignUpFactory,
+    SignUpGroupFactory,
+)
+
+# === util methods ===
+
+
+def _delete_gdpr_data(api_client: APIClient, user_uuid: UUID):
+    gdpr_profile_url = reverse("helsinki_gdpr:gdpr_v1", kwargs={"uuid": user_uuid})
+    return api_client.delete(gdpr_profile_url)
+
+
+# === tests ===
+
+
+@pytest.mark.django_db
+def test_authenticated_user_can_delete_own_data(api_client, settings):
+    settings.GDPR_API_DELETE_SCOPE = api_token_auth_settings.API_SCOPE_PREFIX
+
+    user = UserFactory()
+
+    registration = RegistrationFactory()
+
+    signup_group = SignUpGroupFactory(registration=registration, created_by=user)
+    SignUpFactory(signup_group=signup_group, registration=registration, created_by=user)
+    SignUpFactory(registration=registration, created_by=user)
+
+    assert User.objects.count() == 1
+    assert SignUpGroup.objects.count() == 1
+    assert SignUp.objects.count() == 2
+
+    with (
+        requests_mock.Mocker() as req_mock,
+        patch(
+            "registrations.signals._signup_or_group_post_delete"
+        ) as mocked_signup_post_delete,
+    ):
+        auth_header = get_api_token_for_user_with_scopes(
+            user.uuid, [settings.GDPR_API_DELETE_SCOPE], req_mock
+        )
+        api_client.credentials(HTTP_AUTHORIZATION=auth_header)
+
+        response = _delete_gdpr_data(api_client, user.uuid)
+        assert response.status_code == status.HTTP_204_NO_CONTENT
+
+        assert mocked_signup_post_delete.called is True
+
+    assert User.objects.count() == 0
+    assert SignUpGroup.objects.count() == 0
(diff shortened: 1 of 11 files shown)


registrations/migrations/0023_signup_and_signup_group_user_data.py
```diff
@@ -0,0 +1,61 @@
+# Generated by Django 3.2.20 on 2023-09-12 13:03
+
+from django.db import migrations, models
```
nicobav, line 3: I think at least the signup model part of this should be in the first  [commit](https://github.com/City-of-Helsinki/linkedevents/pull/708/commits/c839fb6809c823ce87220197b452880cfaa7fa9e) since this overrides many things introduced there

helevents/tests/test_gdpr_api_delete.py
```diff
@@ -1,9 +1,10 @@
 from unittest.mock import patch
+from uuid import UUID
```
nicobav, line 2: Churn happening in this commit. Should combine small logical wholes into own commits => e.g introducing tests for these delete and get in one commit or their own (or even part of the commits where the get function is introduced and where the delete part is introduced)

nicobav (changes requested):
Some commit re-organization todo. Introduction of logical wholes in own commits - avoiding overwrites in following commits.

harriris-vincit:
> Some commit re-organization todo. Introduction of logical wholes in own commits - avoiding overwrites in following commits.

@nicobav Perhaps I could re-arrange the commits into three different ones like this:

1. GDPR API integration (model changes, migrations and the business logic).
2. Changes to the existing tests.
3. Addition of the new GDPR API tests.

Any objections to this?

requirements.txt
```diff
@@ -1,5 +1,5 @@
 #
-# This file is autogenerated by pip-compile with Python 3.9
+# This file is autogenerated by pip-compile with Python 3.10
```
nicobav, line 2: I've learned this from the hard way - using python 3.10 with pip-compile it removes necessary packages. So urge you to run this with python 3.9. Our envs are using 3.9.

nicobav:
> > Some commit re-organization todo. Introduction of logical wholes in own commits - avoiding overwrites in following commits.
> 
> @nicobav Perhaps I could re-arrange the commits into three different ones like this:
> 
>     1. GDPR API integration (model changes, migrations and the business logic).
> 
>     2. Changes to the existing tests.
> 
>     3. Addition of the new GDPR API tests.
> 
> 
> Any objections to this?

Not against, **but** if asking 🤓 => to my taste I'd separate model changes to own commit and introduce usage (api, business logic) of those model changes in the latter own commits. Tests could be introduced in the same commit where the logic to be tested is OR in own commit after the "logical" commit. Same applies with the api tests, in the api commit or actually I would separate the API commits into own commits; GET and DELETE related.

So the stacking of the changes could go like this;
1. Model changes with migrations (and some tests if there's logic that results) one commit
2. Business logic => Helpers to models and signals (and some tests) in own commits if they do some separate things. => so can be multiple commits.
4. API changes which uses those new model stuff 
   a) GET api functionalities (and tests) --------------------Own commit
   b) DELETE api functionalities (and tests) --------------- Own commit
   *(c) api tests --------------------------------------------Own commit if not introduced in above commits)*

harriris-vincit:
@nicobav re-requesting review after re-arranging commits and implementing the other improvement suggestions as well.

registrations/models.py
```diff
@@ -319,6 +321,22 @@ class SignUpGroup(CreatedModifiedBaseModel, SignUpMixin):
         default=None,
     )
 
+    user_data = models.ForeignKey(
```
nicobav, line 324: I see we're assigning this user_data to be the same as the `created_by` user, which raises a question; Is this relation necessary and could we use just the `created_by` relation? In the gdpr api we could fetch the sign up data based with the created_by field 🤔

harriris-vincit: I was thinking this same while implementing this, but then thought that here could be two benefits, though pretty minor ones, to having a model and a relation for the GDPR data:
- looking in the database only for whatever reason like data analysis, it is easy to recognize tables that contain GDPR data when they have a field like this relation
- deletion of the GDPR data happens through the cascade behavior so there is no need to introduce additional code for deleting the separate tables' rows

nicobav: Agree that the CASCADE is handy - of course we could have the cascading on the created_by user but perhaps its reasonable to have it like this for future use cases etc. 👌

nicobav:
Looking a OK after resolving conflicts 👍

harriris-vincit:
> Looking a OK after resolving conflicts 👍

After a brief discussion with Tuomas this morning, we decided to still make the change that you too suggested: to not have the separate model `UserData` but rather fetch and delete the related data in code only. Re-requesting a review once more due to this, @nicobav

danipran (approved):
LGTM :+1:

PR #713, opened 2023-09-20
LINK-1578 | Add no-cache headers for all authenticated requests

charn:
This PR adds a middleware which will add the no-cache headers for all authenticated requests.


nicobav (approved):
🤌

PR #715, opened 2023-09-21
LINK-1576 | Registration admin fields

jorilindell:
## Description
- Allow admin and registration admin users to see `created_by`, `last_modified_by` and `registration_user_accesses` fields
- `has_registration_user_access` field which is set to true if current user is strongly identificated and is in registration_user_accesses

## Closes
[LINK-1576](https://helsinkisolutionoffice.atlassian.net/browse/LINK-1576)

[LINK-1576]: https://helsinkisolutionoffice.atlassian.net/browse/LINK-1576?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/events/api.py
+++ b/events/api.py
@@ -1847,11 +1847,43 @@ class RegistrationSerializer(LinkedEventsSerializer, RegistrationBaseSerializer)
         queryset=Event.objects.all(),
     )
 
-    only_admin_visible_fields = (
-        "created_by",
-        "last_modified_by",
-        "registration_user_accesses",
-    )
+    # LinkedEventsSerializer removes created_by and last_modified_by default, so set this
+    # to empty array and have custom logic to hide admin fields in to_representation
+    only_admin_visible_fields = []
+
+    def __init__(
+        self,
+        *args,
+        **kwargs,
+    ):
+        super().__init__(*args, **kwargs)
+        context = self.context
+
+        if "registration_admin_tree_ids" in context:
+            self.registration_admin_tree_ids = context["registration_admin_tree_ids"]
+
+    def to_representation(self, obj):
+        # These fields are visible only if user is admin, registration admin or registration user access user
+        admin_visible_fields = (
+            "created_by",
+            "last_modified_by",
+            "registration_user_accesses",
+        )
+
+        ret = super().to_representation(obj)
+        user = self.user
+
+        show_admin_fields = not user.is_anonymous and (
+            obj.publisher.tree_id in self.admin_tree_ids
+            or obj.publisher.tree_id in self.registration_admin_tree_ids
+        )
+
+        if not show_admin_fields:
+            for field_name in admin_visible_fields:
+                if field_name in ret:
+                    del ret[field_name]
+
+        return ret
 
     def _create_or_update_registration_user_accesses(
         self, registration, registration_user_accesses
--- a/events/api.py
+++ b/events/api.py
@@ -1859,8 +1859,9 @@ class RegistrationSerializer(LinkedEventsSerializer, RegistrationBaseSerializer)
         super().__init__(*args, **kwargs)
         context = self.context
 
-        if "registration_admin_tree_ids" in context:
-            self.registration_admin_tree_ids = context["registration_admin_tree_ids"]
+        self.registration_admin_tree_ids = context.get(
+            "registration_admin_tree_ids", set()
+        )
 
     def to_representation(self, obj):
         # These fields are visible only if user is admin, registration admin or registration user access user
--- a/events/api.py
+++ b/events/api.py
@@ -533,6 +533,8 @@ class LinkedEventsSerializer(TranslatedModelSerializer, MPTTModelSerializer):
         if skip_fields is None:
             skip_fields = set()
 
+        self.skip_fields = skip_fields
+
         if "request" in context:
             self.request = context["request"]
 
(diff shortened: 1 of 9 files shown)


events/api.py
```diff
@@ -1847,11 +1847,43 @@ class RegistrationSerializer(LinkedEventsSerializer, RegistrationBaseSerializer)
...
     )
 
-    only_admin_visible_fields = (
-        "created_by",
-        "last_modified_by",
-        "registration_user_accesses",
-    )
+    # LinkedEventsSerializer removes created_by and last_modified_by default, so set this
+    # to empty array and have custom logic to hide admin fields in to_representation
+    only_admin_visible_fields = []
```
danipran, line 1852: This is a bit confusing, so instead of doing this and overriding `to_representation`, I'd rather refactor `LinkedEventsSerializer` a bit.

There's this bit of code in `LinkedEventsSerializer.to_representation`:
```py
        # display non-public fields if 1) obj has publisher org and 2) user belongs to the same org tree
        # never modify self.skip_fields, as it survives multiple calls in the serializer across objects
        obj_skip_fields = set(self.skip_fields)
        if (
            self.user
            and hasattr(obj, "publisher")
            and obj.publisher
            and obj.publisher.tree_id in self.admin_tree_ids
        ):
            for field in self.only_admin_visible_fields:
                obj_skip_fields.remove(field)
        for field in obj_skip_fields:
            if field in ret:
                del ret[field]
        return ret
```
...and this bit in its `__init__`:
```py
        # by default, admin fields are skipped
        self.skip_fields = skip_fields | set(self.only_admin_visible_fields)
```

The only code that refers to `only_admin_visible_fields` is `LinkedEventsSerializer`, and, of course, this serializer, while `skip_fields` is only referred in serializer contexts, so they should be safe to modify without causing any behavioral oddity.

What I'd change is:
1. `self.skip_fields` doesn't automatically append `self.only_admin_visible_fields` into itself.
2. Instead of removing `only_admin_visible_fields` from `obj_skip_fields` if the user is an admin, do the inverse (so add the fields if the user is not an admin)
3. Do some code splitting; split the admin check part and/or the part where you get the fields to skip to its own method(s)
4. In `RegistrationSerializer`, override the new method(s) to your liking, e.g. override the admin check part to whatever's written in `to_representation` at the moment
5. In `RegistrationSerializer`, use `only_admin_visible_fields` as usual.

This should make the code a bit more logical and less complex.

danipran (changes requested):

helevents/models.py
```diff
@@ -233,3 +248,10 @@ def is_regular_user_of(self, publisher):
         if publisher is None:
             return False
         return self.organization_memberships.filter(id=publisher.id).exists()
+
+    def is_registration_user_access_user_of(self, registration_user_accesses):
+        """Check if current user can be found in registration user accesses"""
+        return (
+            self.is_strongly_identificated
```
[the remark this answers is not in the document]
jorilindell: Renamed in https://github.com/City-of-Helsinki/linkedevents/pull/715/commits/f0bbcbce0b7e006fd252fdd2d2be1090d281834f

events/api.py
jorilindell: Very nice suggestion. 
I added `are_only_admin_visible_fields_allowed` method to check are admin visible fields allowed. `only_admin_visible_fields` are appended to skip_fields only if the user is admin.
I override that method in `RegistrationSerializer` to allow also registration admin users to see admin fields

Changed are committed in https://github.com/City-of-Helsinki/linkedevents/pull/715/commits/2f71eb4838d2c710885ae548e5705d1d69817c2c

events/api.py
```diff
@@ -520,6 +520,14 @@ class LinkedEventsSerializer(TranslatedModelSerializer, MPTTModelSerializer):
     )
     only_admin_visible_fields = ("created_by", "last_modified_by")
 
+    def are_only_admin_visible_fields_allowed(self, obj):
```
danipran, line 523: Small nitpick; move this method after `__init__`. It's like an unspoken rule, but the constructor should almost always be the first method in a class. Doesn't really affect anything else than readability.

registrations/tests/test_registration_get.py
```diff
@@ -143,12 +145,53 @@ def test_admin_can_get_registration_not_created_by_self(
 @pytest.mark.django_db
 def test_admin_user_can_see_registration_user_accesses(registration, user_api_client):
     RegistrationUserAccess.objects.create(registration=registration)
+    RegistrationUserAccess.objects.create(registration=registration, email=test_email)
```
danipran, line 148: Could use `RegistrationUserAccessFactory` instead of `RegistrationUserAccess.objects.create`. Applies to everything in this file. Honestly, with simple test cases like this, doesn't really matter that much, but personally, I forget the existence of factories every now and then. Seeing them used in other tests reminds me of their existence and encourages me to use them in my own tests as well.

registrations/tests/test_registration_get.py
```diff
@@ -16,6 +16,7 @@
 include_signups_query = "include=signups"
 
 # === util methods ===
+test_email = "test@email.com"
```
danipran, line 19: Move just a couple lines up, right after `include_signups_query`

registrations/tests/test_registration_get.py
```diff
@@ -143,12 +145,53 @@ def test_admin_can_get_registration_not_created_by_self(
...
 
+@pytest.mark.django_db
+def test_registration_user_access_user_can_see_if_he_has_access(
+    registration, user, user_api_client
+):
+    user.get_default_organization().admin_users.remove(user)
+
+    RegistrationUserAccess.objects.create(registration=registration, email=user.email)
+    RegistrationUserAccess.objects.create(
+        registration=registration, email="test@email.com"
```
danipran, line 180:
```suggestion
        registration=registration, email=test_email
```

danipran (approved):
Good work! Came up with some small suggestions, but no need to re-request a review

events/api.py
danipran: Well done!

PR #718, opened 2023-09-25
Espoo Importer

voneiden:
Espoo importer is a rather generic LE importer. Handles various relations between objects etc. although in hindsight nobody has specified if this is necessary. Actually, at this point I still don't know what are the parameters that should be used for importing, so that's configurable via an environment variable.. 

https://helsinkisolutionoffice.atlassian.net/browse/LINK-1375

### DIFF
--- a/events/importer/espoo.py
+++ b/events/importer/espoo.py
@@ -1,800 +1,543 @@
 import logging
-import re
 import time
-from datetime import datetime, timedelta
+import urllib
+from copy import deepcopy
+from typing import Annotated, Any, Callable, Optional, Type, TypeVar, Union
+from urllib.parse import urljoin, urlparse
 
-import bleach
-import dateutil.parser
-import pytz
 import requests
-import requests_cache
-from django.utils.html import strip_tags
+from django.conf import settings
+from django.db import transaction
+from django.db.models import Model
 from django_orghierarchy.models import Organization
-from pytz import timezone
+from requests.adapters import HTTPAdapter
+from requests.exceptions import RetryError
+from rest_framework.fields import empty
+from urllib3 import Retry
 
-from events.models import DataSource, Event, Keyword, Place
+from events.models import DataSource, Event, Keyword, Language, Place
+from events.translation import EventTranslationOptions
 
-from .base import Importer, recur_dict, register_importer
+from ..api import generate_id
+from .base import Importer, register_importer
 from .sync import ModelSyncher
-from .util import clean_text, clean_url
-from .yso import KEYWORDS_TO_ADD_TO_AUDIENCE
 
 # Per module logger
 logger = logging.getLogger(__name__)
 
-# Maximum number of attempts to fetch the event from the API before giving up
-MAX_RETRY = 5
+M = TypeVar("M", bound=Model)
 
-YSO_BASE_URL = "http://www.yso.fi/onto/yso/"
-YSO_KEYWORD_MAPS = {
-    "koululaiset ja opiskelijat": ("p16485", "p16486"),
-    "yhdistykset ja seurat": "p1393",  # both words seems to mean associations
-    "näyttelyt ja tapahtumat": ("p5121", "p2108"),
-    "nuoriso": "p11617",
-    "koulutus, kurssit ja luennot": ("p84", "p9270", "p15875"),
-    "stand up ja esittävä taide": ("p9244", "p2850"),
-    "nuorisotyö": "p1925",
-    "ohjaus, neuvonta ja tuki": ("p178", "p23"),
-    "hyvinvointi ja terveys": ("p38424", "p2762"),  # -> hyvinvointi, terveys
-    "ilmastonmuutos": "p5729",
-    "leirit, matkat ja retket": ("p143", "p366", "p25261"),
-    "kerhot ja kurssit": ("p7642", "p9270"),
-    "internet": "p20405",
-    "tapahtumat": "p2108",
-    "asukastoiminta": "p2250",
-    "rakentaminen": "p3673",
-    "kaavoitus": "p8268",
-    "laitteet ja työtilat": ("p2442", "p546"),  # -> Laitteet, työtilat
-    "museot": "p4934",
-    "museot ja kuvataide": ("p4934", "p2739"),  # -> museot, kuvataide
-    "näyttelyt ja galleriat": ("p5121", "p6044"),  # -> Näyttelyt, galleriat
-    "musiikki": "p1808",
-    "teatteri": "p2625",
-    "kevyt liikenne": "p4288",
-    "liikenne": "p3466",
-    "tiet ja kadut": ("p1210", "p8317"),  # -> Tiet, kadut
-    "liikuntapalvelut": "p9824",
-    "liikuntapaikat": "p5871",
-    "luonto- ja ulkoilureitit": ("p13084", "p5350"),  # -> Luonto, ulkoilureitit
-    "uimahallit": "p9415",
-    "ulkoilualueet": "p4858",
-    "urheilu- ja liikuntajärjestöt": ("p965", "p2042"),  # -> Urheilu, liikuntajärjestöt
-    "virkistysalueet": "p4058",
-    "bändit": "p5072",
-    "nuorisotilat": "p17790",
-    "aikuiskoulutus": "p300",
-    "korkeakouluopetus": "p1246",
-    "perusopetus": "p19327",
-    "päivähoito (lapsille)": "p3523",  # -> Päivähoito
-    "lapsille": "p4354",  # lapset (ikäryhmät)
-    "elokuva": "p16327",  # elokuva (taiteet)
-    "elokuvat": "p16327",  # elokuva (taiteet)
-    "musiikki ja konsertit": ("p1808", "p11185"),  # Musiikki, konsertit
-    "liikunta, ulkoilu ja urheilu": ("p916", "p2771", "p965"),
-    "liikuntalajit": "p916",
-    "ohjattu liikunta": "p916",
-    "harrastus- ja kerhotoiminta": (
-        "p2901",
-        "p7642",
-        "p8090",
-    ),  # Harrastus, Kerho, toiminta
-    "perheet": "p4363",  # perheet (ryhmät)
-    "koko perheelle": "p4363",
-    "yrittäjät ja yritykset": ("p1178", "p3128"),
-    "yrittäjät": "p1178",
-    "lapset": "p4354",
-    "kirjastot": "p2787",
-    "opiskelijat": "p16486",
-    "konsertit ja klubit": ("p11185", "p20421"),  # -> konsertit, musiikkiklubit
-    "kurssit": "p9270",
-    "venäjä": "p7643",  # -> venäjän kieli
-    "seniorit": "p2433",  # -> vanhukset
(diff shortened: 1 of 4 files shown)


events/importer/espoo.py
```diff
@@ -1,800 +1,534 @@
...
+    data: dict,
+    instance: M,
+    instance_map: dict[str, str],
+):
+    """
+    Extension of _post_recreate_m2o for Event.external_links
+    which also require a reference to language object.
+
+    Does not create new Language objects, those that don't exist
+    are skipped.
```
tuomas777, line 214: Maybe those skipped ones could be logged?

events/tests/importers/test_espoo.py
```diff
@@ -0,0 +1,367 @@
...
+    assert [d["id"] for d in origin_data[:1]] == [i.id for i in common_objs]
+    assert origin_data[1:] == origin_objs
+
+
+@pytest.mark.django_db
+def test_importer(settings, requests_mock, sleep, api_client):
+    settings.ESPOO_API_URL = "http://localhost/"
+    settings.ESPOO_API_QUERY = "test=1"
+
+    # Generate the test data in our own API
```
danipran, line 246: It's a good idea to utilize our API for generating the mock responses. However, it's not a good idea to make this dynamic; while Espoo uses Linked Events, it's their own version of Linked Events and is not guaranteed to match with ours. Use a fixture instead.

danipran (changes requested):

events/importer/espoo.py
```diff
@@ -1,800 +1,534 @@
...
+    """
+    if value := obj.get(key):
+        if not isinstance(value, list):
+            value = [value]
+        for row in value:
+            if jsonld:
+                row_id = parse_jsonld_id(row)
+            else:
+                row_id = row["id"]
+            container[row_id] = row
```
tuomas777, line 359: As discussed, unless there is a good reason why these need to be like this, it would be better if these didn't silently ignore missing keys.

events/importer/espoo.py
```diff
@@ -1,800 +1,534 @@
...
-    elif len(groups) == 3:
-        postal_code = groups[1]
-        city = groups[2]
+
+def parse_jsonld_id(data: dict) -> str:
+    ld_reference = data.get("@id")
+    if ld_reference:
+        origin_id = ld_reference.split("/")[-2]
+        if origin_id:
+            return origin_id
```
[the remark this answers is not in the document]
voneiden: ok björner

events/importer/espoo.py
```diff
@@ -1,800 +1,534 @@
...
+def parse_jsonld_ids(data: list) -> list[str]:
+    return [parse_jsonld_id(d) for d in data]
+
+
+def _get_data(url: str) -> dict:
+    for _ in range(settings.ESPOO_MAX_RETRIES):
+        response = requests.get(url)
+        if response.status_code == 200:
+            return response.json()
+        time.sleep(settings.ESPOO_WAIT_BETWEEN)
```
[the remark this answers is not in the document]
voneiden: Not supported by requests_mock ( https://github.com/jamielennox/requests-mock/issues/223 )

events/importer/espoo.py
```diff
@@ -1,800 +1,534 @@
...
+    to existing models in other data sources. For example tprek entries
+    should generally be identical between Helsinki LE and Espoo LE.
+    """
+    origin_obj_ids = set(obj["id"] for obj in origin_objs)
+    common_objs = list(
+        model.objects.filter(id__in=origin_obj_ids)
+        .exclude(data_source=data_source)
+        .only("id")
+    )
+    common_obj_ids = [org.id for org in common_objs]
```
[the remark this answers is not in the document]
voneiden: There it was! I remembered seeing that but failed to find it and gave up thinking maybe I fixed it :joy: Yes, a remnant

events/importer/espoo.py
```diff
@@ -1,800 +1,534 @@
...
+    :param mapping:
+    :param jsonld:
+    :return:
+    """
+
+    def pre_map_to_id(
+        field_name: str, field_data: Union[str, dict], data: dict
+    ) -> dict[str, Any]:
+        if field_data is None:
+            return {**data}
```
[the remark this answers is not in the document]
voneiden: Changed to use {**data} in other mappers instead to be consistent about returning copies of dicts.

events/importer/espoo.py
```diff
@@ -1,800 +1,534 @@
...
+
+M = TypeVar("M", bound=Model)
+
+
+# Pre field actions are functions that receive arguments:
+# - field_name: str (field that is currently being processed)
+# - field_data: Any (origin object data)
+# - data: dict (target object data)
+# and return a new dict of fields to be used for the creation of the target object
+def _pre_reverse_map_tuple(mapping: tuple[Any, Any]):
```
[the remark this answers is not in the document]
voneiden: I'll go with `build`.. also a better type annotation now:

```python
def _build_pre_map_to_id_mapper(mapping: dict[str, str], jsonld=False) -> PreMapper:
```

A wee bit java-esque naming.

events/importer/espoo.py
```diff
@@ -1,800 +1,534 @@
...
+):
     """
-    Extract the url from the html tag if any, and return it cleaned if valid
+    For relations that the API does not provide an ID for, such as
+    Event.offers.
+
+    Delete old ones and recreate.
+    """
+    field = getattr(instance, field_name)
+    field.all().delete()
```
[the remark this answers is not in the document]
voneiden: Suppose I could add a hardcoded allowed_fields there as a safeguard to make the danger more obvious.

voneiden:
OK, I believe I've addressed all the comments. Requesting re-review and the new changes are in separate commits for now.

events/importer/espoo.py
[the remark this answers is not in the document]
danipran: This hurts 🥲 test code should never stop you from implementing stuff in production code. Tests are there to validate that your code works as intended. If you use `HTTPAdapter` & `Retry`, you shouldn't need to test the retrying. They're presumably tested in `requests`/`urllib3` already. You could test for `max_retries` or something too if you want to be extra sure with having those retries in place.

That being said, this change is not exactly needed, it's just if you want to use something existing already. `Retry` does include handy stuff like backoffs and retrying only for certain status codes. A custom implementation is fine too, especially since it's a rather limited and simple use case, so do what you will.

events/importer/espoo.py
```diff
@@ -18,84 +20,101 @@
...
+def _get_data(url: str, params: Optional[dict[str, str]] = None) -> dict:
     for _ in range(settings.ESPOO_MAX_RETRIES):
-        response = requests.get(url)
+        response = requests.get(url, timeout=settings.ESPOO_TIMEOUT, params=params)
         if response.status_code == 200:
             return response.json()
+        elif response.status_code < 500:
+            raise EspooImporterError(
+                f"ESPOO API returned status code {response.status_code}"
+            )
```
danipran, line 74: Probably shouldn't retry on all 4xx/5xx codes. Retrying on 503 (service unavailable) should be fine, as well as 429 (too many requests). Of course, this is Linked Events we're talking about, so you're mileage may vary,

danipran (changes requested):
Some of the earlier comments were not addressed.

events/importer/espoo.py
voneiden: > you shouldn't need to test the retrying

I most certainly do want to have proof that my code will trigger retries as expected regardless of whether I use my own retry logic or external retry logic. Anyway, I still have the HTTPAdapter/Retries code here, so I'll see if there's some reasonable way to test this.

events/importer/espoo.py
```diff
@@ -1,800 +1,556 @@
...
+    Translation field mapper
+    """
+    if field_data is None:
+        return data
+
+    # LANGUAGES is a tuple containing (code, label) pairs
+    # [("fi", "Finnish"), ...]
+    # zip it into a list containing a tuple of codes and a tuple of labels
+    # [("fi", ...), ("Finnish", ...)]
+    language_codes = list(zip(*settings.LANGUAGES))[0]
```
charn, line 135:
```suggestion
    language_codes = [lang[0] for lang in settings.LANGUAGES]
```
I think this would be more readable and doesn't necessarily require a comment.

voneiden: caseclosed

events/importer/espoo.py
danipran: No idea what I've thought here with the retries. Feel free to resolve this.

> I most certainly do want to have proof that my code will trigger retries as expected regardless of whether I use my own retry logic or external retry logic. Anyway, I still have the HTTPAdapter/Retries code here, so I'll see if there's some reasonable way to test this.

I'm guessing you're talking about this thread? https://github.com/City-of-Helsinki/linkedevents/pull/718#discussion_r1339780066

linkedevents/settings.py
```diff
@@ -60,6 +60,15 @@ def get_git_revision_hash() -> str:
     ),
     ENABLE_EXTERNAL_USER_EVENTS=(bool, True),
     ENABLE_REGISTRATION_ENDPOINTS=(bool, False),
+    ESPOO_API_URL=(str, "https://api.espoo.fi/events/"),
+    ESPOO_API_QUERY_PARAMS=(
```
tuomas777, line 64: As these are only used with event resource queries, I would rather name this `ESPOO_API_EVENT_QUERY_PARAMS` (I actually first thought these were used in every query when I saw the setting 🙂)

events/importer/espoo.py
voneiden: Yeah my comments here are not terribly relevant at this point: the HTTPAdapter/Retry is baked in and there's also a test that manages to test the retry functionality so I'm happy. It's using the default retry behaviour, so 429  will retry based on the Retry-After.

events/tests/importers/test_espoo.py
```diff
@@ -0,0 +1,449 @@
...
+    class Sleep:
+        call_count = 0
+
+        def sleep(self, t):
+            self.call_count += 1
+
+    sleep_instance = Sleep()
+
+    monkeypatch.setattr("time.sleep", sleep_instance.sleep)
+    return sleep_instance
```
danipran, line 75: Could use `Mock` instead:
```suggestion
@pytest.fixture
def sleep(monkeypatch):
    mock_sleep = Mock()  # or a MagicMock
    monkeypatch.setattr("time.sleep", mock_sleep)
    return mock_sleep
```

Or if you don't like mixing & matching:
```suggestion
@pytest.fixture
def sleep():
    with patch("time.sleep") as mock:  # not completely sure on the target; maybe "events.importer.sleep" if this one doesn't work?
        yield mock
```

events/tests/importers/test_espoo.py
```diff
@@ -0,0 +1,449 @@
...
+                {
+                    "@id": "https://api.espoo.localhost/linkedevents/v1/place/espoo:asdfasdf/",
+                }
+            ),
+        ]
+    )
+
+
+@pytest.fixture
+def sleep(monkeypatch):
```
danipran, line 65: I'd prefer the name `mock_sleep`, but up to you.
```suggestion
def mock_sleep(monkeypatch):
```

danipran (approved):
Some suggestions for the tests, otherwise LGTM :+1:

PR #720, opened 2023-09-26
LINK-1580: Enable hourly kulke import, disable kulke course import

danipran:
Kulke importer works well enough to be enabled once again.

Commented out course import in code level to avoid any accidental course imports (since we're not quite sure if it works as intended).

Refs LINK-1580

# Note: Do not merge if you don't know what you're doing


tuomas777 (approved):
lets-go-lets-do-this-dwight-schrute-781hy0bq8jb4kr64

PR #721, opened 2023-09-26
LINK-1556 LINK-1427 | Event API: Add filter for excluding sub events of recurring super events

charn:
This PR adds a filter to the event API which allows filtering sub events of recurring super events from the results.


events/tests/test_event_get.py
```diff
@@ -1331,3 +1338,26 @@ def test_sort_events_by_minimum_attendee_capacity(api_client, event, event2, eve
...
+@pytest.mark.django_db
+def test_get_event_list_hide_recurring_children(api_client, hide_children):
+    now = timezone.now()
+    event_1 = EventFactory(origin_id=1, start_time=now)
+    event_2 = EventFactory(data_source=event_1.data_source, origin_id=2, start_time=now)
+    super_1 = create_super_event([event_1, event_2], event_1.data_source)
+
+    response = get_list(
+        api_client=api_client, query_string=f"hide_recurring_children={hide_children}"
+    )
```
danipran, line 1353: I think you can test the filter directly instead of using the API 🤔 integration tests are slower and less accurate and testing this through the API isn't exactly necessary.

E.g. testing with `False`:
```py
    now = timezone.now()
    event_1 = EventFactory(origin_id=1, start_time=now)
    event_2 = EventFactory(data_source=event_1.data_source, origin_id=2, start_time=now)
    super_1 = create_super_event([event_1, event_2], event_1.data_source)
    qs = Event.objects.all()
    filter_set = EventFilter()

    result = filter_set.filter_hide_recurring_children(qs, "hide_recurring_children", False)

    assert qs == result
```

Disclaimer: haven't tested this myself, so might not actually work.

events/tests/test_event_get.py
```diff
@@ -1331,3 +1338,26 @@ def test_sort_events_by_minimum_attendee_capacity(api_client, event, event2, eve
...
     assert results[1]["id"] == event2.id
     assert results[2]["id"] == event.id
+
+
+@pytest.mark.parametrize("hide_children", [True, False])
+@pytest.mark.django_db
+def test_get_event_list_hide_recurring_children(api_client, hide_children):
+    now = timezone.now()
+    event_1 = EventFactory(origin_id=1, start_time=now)
+    event_2 = EventFactory(data_source=event_1.data_source, origin_id=2, start_time=now)
```
danipran, line 1348: Do the `origin_id` and `start_time` values matter in this test? Can you leave them blank?

danipran (changes requested):

danipran (approved):
LGTM! 👍

PR #725, opened 2023-10-04
LINK-1615 | Fix external organisation set as publisher

charn:
When creating an event the publisher field is not required. If the publisher field is not sent it will default to:
- users organization, 
- data sources owner when api key is used or 
- to the external organization when external user sends the event. 

There was an issue where the external organization was set as the publisher if the publisher was not in the event data. The fix is to set the default external organization only if an external user sends the data and the publisher is missing.


voneiden (approved):
Seems good

PR #727, opened 2023-10-05
LINK-1626 | Allow registration user access to update signup created by the user

jorilindell:
## Description
This PR fixes the edge case that user belongs to registration user accesses and user is updating a signup or signup group created by user. Registration user access can update only presence_status field and signup creator can edit all other fields. So in this edge case user should be able to update all signup or signup group fields.

Also added strong identification for the tests that tested registration user access permissions to avoid false positives.

## Closes
[LINK-1626](https://helsinkisolutionoffice.atlassian.net/browse/LINK-1626)

[LINK-1626]: https://helsinkisolutionoffice.atlassian.net/browse/LINK-1626?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/registrations/permissions.py
+++ b/registrations/permissions.py
@@ -62,14 +62,15 @@ class CanAccessSignup(UserDataFromRequestMixin, permissions.BasePermission):
         ):
             return True
 
-        if obj.registration.registration_user_accesses.filter(
-            email=request.user.email
-        ).exists():
-            # Registration user can only change presence_status and nothing else,
-            # and they need to be strongly identified.
+        if request.user.is_registration_user_access_user_of(
+            obj.registration.registration_user_accesses
+        ):
+            # Strongly identified registration user can only change presence_status
+            # and nothing else.
             data_keys = set(request.data.keys())
-            return request.user.is_strongly_identified and (
-                data_keys == {"id", "presence_status"}
+            return (
+                obj.created_by_id == request.user.id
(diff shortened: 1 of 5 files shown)


registrations/permissions.py
```diff
@@ -57,21 +57,28 @@ def has_permission(self, request: Request, view: APIView) -> bool:
...
+            obj.registration.registration_user_accesses
+        ):
+            # Strongly identified registration user can only change presence_status
+            # and nothing else.
             data_keys = set(request.data.keys())
-            return request.user.is_strongly_identified and (
-                data_keys == {"id", "presence_status"}
-                or data_keys == {"presence_status"}
-            )
+            return data_keys == {"id", "presence_status"} or data_keys == {
```
harriris-vincit, line 79: Now the `is_registraton_user_access_user_of` check will be done twice if `request.user` is a registration user, but not the `created_by` user. To prevent this, I would maybe check `obj.created_by_id == request.user.id` here:
```
return obj.created_by_id == request.user.id or (data_keys == <x> or data_keys == <y>)
```

charn (approved):
LGTM! :+1: :shipit:

harriris-vincit (approved):

PR #728, opened 2023-10-05
LINK-1587 | Prevent the HEAD method for registrations API

harriris-vincit:
### Description
OWASP states that only necessary HTTP verbs / methods should be allowed for an API. The HEAD method is seen as such for the registrations API and will not be allowed after these changes.

To ensure a semantically correct HTTP response status code (405), the method has been prevented in the ViewSets instead of in the permissions.
### Closes
[LINK-1587](https://helsinkisolutionoffice.atlassian.net/browse/LINK-1587)

[LINK-1587]: https://helsinkisolutionoffice.atlassian.net/browse/LINK-1587?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/registrations/api.py
+++ b/registrations/api.py
@@ -52,15 +52,15 @@ from registrations.utils import send_mass_html_mail
 logger = logging.getLogger(__name__)
 
 
+class RegistrationsAllowedMethodsMixin:
+    http_method_names = ["post", "put", "patch", "get", "delete", "options"]
+
+
 class RegistrationViewSet(
     UserDataSourceAndOrganizationMixin,
     JSONAPIViewMixin,
-    mixins.RetrieveModelMixin,
-    mixins.UpdateModelMixin,
-    mixins.DestroyModelMixin,
-    mixins.ListModelMixin,
-    mixins.CreateModelMixin,
-    viewsets.GenericViewSet,
+    RegistrationsAllowedMethodsMixin,
+    viewsets.ModelViewSet,
 ):
     serializer_class = RegistrationSerializer
     queryset = Registration.objects.all()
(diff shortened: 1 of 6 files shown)


charn (approved):
LGTM! :+1: :shipit:

harriris-vincit:
> I understand that here we're preventing a potential security issue if there ever happens to be code that for some reason treats HEAD and GET differently from the permission perspective.
> 
> Generally speaking HEAD is just a "short" GET and is also considered to be a safe method (in django-rest-framework). Backend doesn't currently specifically check if someone is using GET, but rather checks if:
> 
> ```
> request.method in permissions.SAFE_METHODS
> ```
> 
> So in that regard there shouldn't be any apparent security issues. From the viewpoint of permissions HEAD and GET seem to be treated in the same manner. [Default django View treats](https://github.com/django/django/blob/main/django/views/generic/base.py#L127-L128) HEAD the same as GET if a separate head method hasn't been defined for the view.
> 
> HEAD requests are processed just like GET requests. The content is removed from the response right before it's sent back.
> 
> This change does make the API a bit inconsistent as a whole. Maybe since registration is somewhat a separate part of the API, it doesn't matter.

Yes, this is mainly hardening the code for potential threats rather than actual known ones. The original [OWASP paper](https://cheatsheetseries.owasp.org/assets/REST_Security_Cheat_Sheet_Bypassing_VBAAC_with_HTTP_Verb_Tampering.pdf) seems to deal with VBAAC configurations and there recommends disabling the HEAD method. This has in the OWASP web docs then been generalized to a common recommendation. I think the most important thing regarding this would be to be as explicit as possible in defining the allowed methods per endpoint throughout the whole project and have that documented well too

PR #733, opened 2023-10-06
LINK-1543 | Remove stale allauth tables and contenttypes

charn:
In commit bf155e2 the dependency to django-allauth was removed. But that commit didn't clean allauth stuff away from the database. This commit does that.

A migration file handles the cleaning. The migration file is added to the linkedevents app.


voneiden (approved):
Oh, yes. Little Bobby Tables, we call him.

PR #738, opened 2023-10-13
Add check_deprecated_yso_keywords command

voneiden:
Utility command for fixing missing replacements from yso keywords

Ref https://helsinkisolutionoffice.atlassian.net/browse/LINK-1606


events/management/commands/check_deprecated_yso_keywords.py
```diff
@@ -0,0 +1,88 @@
...
+        graph = importer.load_graph_into_memory(yso.URL)
+        deprecated_keywords_without_replacement = Keyword.objects.filter(
+            data_source=importer.data_source, deprecated=True, replaced_by__isnull=True
+        )
+
+        self.stdout.write(
+            f"Found {len(deprecated_keywords_without_replacement)} deprecated keywords without replacements"
+        )
+        if len(deprecated_keywords_without_replacement) == 0:
+            self.stdout.write(self.style.SUCCESS("ALl good, nothing to do!"))
```
danipran, line 36:
```suggestion
            self.stdout.write(self.style.SUCCESS("All good, nothing to do!"))
```

events/management/commands/check_deprecated_yso_keywords.py
```diff
@@ -0,0 +1,88 @@
...
+from events.importer import yso
+from events.models import Keyword
+
+
+class DryRun(Exception):
+    pass
+
+
+class Command(BaseCommand):
+    help = "Check replacements of deprecated keywords."
```
danipran, line 13: "Check" would imply a safe operation; however, this is only a check if you run it dry. Maybe rename to `replace_deprecated_yso_keywords`? Or something similar that makes it clear that this *will* alter data.

events/management/commands/check_deprecated_yso_keywords.py
```diff
@@ -0,0 +1,88 @@
...
+            "--apply",
+            action="store_true",
+            help="Apply changes",
+        )
+
+    def handle(self, **options):
+        apply = options["apply"]
+        importer = yso.YsoImporter({})
+        importer.setup()
+        self.stdout.write("Loading graph into memory, this can take a while..")
```
danipran, line 26: [An ellipsis consists of three periods.](https://en.wikipedia.org/wiki/Ellipsis)
```suggestion
        self.stdout.write("Loading graph into memory, this can take a while...")
```

events/management/commands/check_deprecated_yso_keywords.py
```diff
@@ -0,0 +1,88 @@
...
+        self.stdout.write(
+            f"Found {len(deprecated_keywords_without_replacement)} deprecated keywords without replacements"
+        )
+        if len(deprecated_keywords_without_replacement) == 0:
+            self.stdout.write(self.style.SUCCESS("ALl good, nothing to do!"))
+            return
+
+        replacements = []
+        missing_replacements = []
+        self.stdout.write("Finding replacements..")
```
danipran, line 41:
```suggestion
        self.stdout.write("Finding replacements...")
```

events/management/commands/check_deprecated_yso_keywords.py
```diff
@@ -0,0 +1,88 @@
...
+            if replacement:
+                keyword.replaced_by_id = yso.get_yso_id(replacement)
+                replacements.append(keyword)
+            else:
+                missing_replacements.append(keyword)
+
+        if missing_replacements:
+            self.stdout.write(
+                self.style.WARNING(
+                    f"Missing replacements for {len(missing_replacements)} keywords"
```
danipran, line 54: Add a colon to make it clear that the following messages are part of the same message.
```suggestion
                    f"Missing replacements for {len(missing_replacements)} keywords:"
```

danipran (changes requested):
Also a slight problem: some of the manual replacements for YsoImporter in `YSO_DEPRECEATED_MAPS` do not actually work since they're in different dictionaries.

E.g. [yso:p27158](https://finto.fi/yso/fi/page/p27158) is replaced by [yso:p508707](https://finto.fi/yso-paikat/fi/page/p508707), however, the latter one is in `yso-paikat`, not `yso`. so YsoImporter never finds the replacements. YsoImporter's code just pretty much skips them instead and leaves the old one deprecated without a replacement.

It's not a problem in the command itself, but the command *will* fail because of this.

events/management/commands/check_deprecated_yso_keywords.py
voneiden: I've updated the help text to read "Check or apply". The default operation is to check only.

danipran (approved):
LGTM :+1:

PR #739, opened 2023-10-17
Move dysfunctional deprecated kw validation from model.save to serializer

voneiden:
Previously Event.save did m2m validation on keywords, but that does not work as intended, as serializers update m2m relations after the model save is done. Therefore an event with deprecated keywords can not be updated, as the model cannot be saved before the deprecated keywords are  removed. Also validation logic belongs to the serializer, not the model save.

Ref https://helsinkisolutionoffice.atlassian.net/browse/LINK-1605

### DIFF
--- a/events/models.py
+++ b/events/models.py
@@ -1037,25 +1037,6 @@ class Event(MPTTModel, BaseModel, SchemalessFieldMixin, ReplacedByMixin):
                     }
                 )
 
-        if (
-            self.keywords.filter(deprecated=True)
-            or self.audience.filter(deprecated=True)
-        ) and (not self.deleted):
-            raise ValidationError(
-                {
-                    "keywords": _(
-                        "Trying to save event with deprecated keywords "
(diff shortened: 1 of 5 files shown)


events/models.py
```diff
@@ -1037,22 +1037,6 @@ def save(self, *args, **kwargs):
...
-                        "Trying to save event with deprecated keywords "
-                        + str(self.keywords.filter(deprecated=True).values("id"))
-                        + " or "
-                        + str(self.audience.filter(deprecated=True).values("id"))
-                        + ". Please use up-to-date keywords."
-                    )
-                }
-            )
-
         # if self.location__divisions__ocd_id__endswith == MUNIGEO_MUNI:
```
nicobav, line 1040: while you're at it, could you please 🙏 😇

nicobav (approved):
Magnificent 🎩  

Comments are just adding my humble request to clean up some dust from the corners

charn (approved):
LGTM, just need to fix the tests! :+1: :ship:

PR #740, opened 2023-10-17
LINK-1638 | Make harrastushaku importer support overnight events

nicobav:
If the end time is before start time interpret that as a overnight event.

Refs LINK-1638

### DIFF
--- a/events/importer/harrastushaku.py
+++ b/events/importer/harrastushaku.py
@@ -475,21 +475,28 @@ class HarrastushakuImporter(Importer):
             weekday = int(time_table.get("weekday"))
             start_time = parse_time(time_table.get("starttime"))
             end_time = parse_time(time_table.get("endtime"))
+            end_time_on_next_day = start_time >= end_time
             repetition = int(time_table.get("repetition"))
             if repetition == 0:
                 repetition = 7  # assume repetition 0 and 7 mean the same thing
 
-            if not (weekday and repetition) or start_time >= end_time:
+            if not (weekday and repetition):
                 continue
 
             while current_date.isoweekday() != weekday:
                 current_date += timedelta(days=1)
 
             while current_date <= end_date:
+                ending_date = (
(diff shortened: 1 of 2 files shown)


tuomas777 (approved):
LGTM 👍 👯‍♂️

events/importer/harrastushaku.py
```diff
@@ -475,21 +475,29 @@ def build_sub_event_time_ranges(self, start_date, end_date, time_tables):
...
             while current_date.isoweekday() != weekday:
                 current_date += timedelta(days=1)
 
             while current_date <= end_date:
+                end_datetime = datetime.combine(current_date, end_time).astimezone(
+                    TIMEZONE
+                )
+
+                if end_time_on_next_day:
+                    end_datetime += timedelta(days=1)
```
voneiden, line 495: If this happens just before DST change, then the next day will be calculated in the wrong timezone, eg following test case will fail
```
        (
            "2023-03-25",
            "2023-03-25",
            [make_timetable(starttime="15:00", endtime="15:00", weekday="6")],
            [("2023-03-26T15:00:00+02:00", "2023-03-27T15:00:00+03:00")],
        ),
```

because instead of end value of  `2023-03-27T15:00:00+03:00`  it calculates `2023-03-27T15:00:00+02:00` which in summer time is `2023-03-27T16:00:00+03:00` , so the event will actually appear to end at 16:00 next day.

IMO it should be still 15:00 next day regardless of DST transition. Feel free to argue against this though. There are some utility functions in `events/utils.py` (ex. `def start_of_next_day`) that do similar things by stripping the timezone information before adding/removing days and then restore the timezone afterwards.

voneiden (changes requested):

events/importer/harrastushaku.py
nicobav: I've done some changes regarding this comment

voneiden (approved):
Looks solid!

PR #741, opened 2023-10-17
LINK-1512 | Fine-tune texts in registration email templates

jorilindell:
## Description
- Send different confirmation message to a single signup and to a signup group
- Send different confirmation to waiting list message to a single signup and to a signup group
- Display username in the signup cancellation confirmation email
- Fine-tune texts in custom message to signup

## Closes
[LINK-1512](https://helsinkisolutionoffice.atlassian.net/browse/LINK-1512)

[LINK-1512]: https://helsinkisolutionoffice.atlassian.net/browse/LINK-1512?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


registrations/templates/signup_group_confirmation_to_waiting_list.html
```diff
@@ -0,0 +1,49 @@
...
+      <div class="content-container" style="padding:0 16px;">
+        <h1 style="color:#1a1a1a;font-size:36px;line-height:43px;font-weight:700;margin:0 0 12px;">
+          {% blocktranslate %}Thank you for signing up for the waiting list{% endblocktranslate %}
+        </h1>
+        <div class="divider" style="background:#0000bf;border-radius:0px;height:3px;width:100%;margin:24px 0;"></div>
+        <div style="margin: 56px 0px 40px">
+          <p style="color:#1a1a1a;font-size:16px;font-weight:normal;line-height:24px;margin:0px;">
+            <!-- Event -->
+            {% if event_type_id == "1" %}
+            {% blocktranslate %}The registration for the event <strong>{{event_name}}</strong> waiting list was successful.{% endblocktranslate %}
```
harriris-vincit, line 16: SonarCloud is complaining about duplication due to these. It seems that the event display name, like "event", "course", etc, could be a variable that could be passed to a single `blocktranslate` string. Or the translated block could be passed completely as a variable to the template.

jorilindell: I refactored confirmation email logic little bit and moved confirmation email template texts to a variable instead of having them in the templates. At the same time also removed unnecessary templates to make css and html consistent and easier to maintain.

registrations/models.py
```diff
@@ -636,6 +767,59 @@ def get_registration_message(self, subject, cleaned_body, plain_text_body):
...
+
+        if notification_type == SignUpNotificationType.CANCELLATION:
+            texts["secondary_heading"] = text_options["secondary_heading"][
+                event_type_id
+            ] % {
+                "event_name": event_name,
+                "username": username,
+            }
+        elif notification_type == SignUpNotificationType.CONFIRMATION:
+            if self.signup_group:
```
harriris-vincit, line 793:
```suggestion
            if self.signup_group_id:
```
Evaluating the related object will make Django fetch the whole related object. Using only the related ID is preferred for performance benefits if only the ID is needed (https://docs.djangoproject.com/en/3.2/topics/db/optimization/#use-foreign-key-values-directly)

registrations/models.py
```diff
@@ -636,6 +767,59 @@ def get_registration_message(self, subject, cleaned_body, plain_text_body):
...
+                texts["heading"] = text_options["group"]["heading"]
+                texts["secondary_heading"] = text_options["group"]["secondary_heading"][
+                    event_type_id
+                ] % {"event_name": event_name}
+            else:
+                texts["secondary_heading"] = text_options["secondary_heading"][
+                    event_type_id
+                ] % {"event_name": event_name}
+        elif notification_type == SignUpNotificationType.CONFIRMATION_TO_WAITING_LIST:
+            if self.signup_group:
```
harriris-vincit, line 804:
```suggestion
            if self.signup_group_id:
```

jorilindell: Changed as suggested

registrations/models.py
```diff
@@ -39,6 +39,137 @@ class SignUpNotificationType:
     TRANSFERRED_AS_PARTICIPANT = "transferred_as_participant"
 
 
+signup_email_texts = {
```
harriris-vincit, line 42: I was thinking that we should probably separate the notification-related code to their own module(s) so that `models.py` could be more model-focused. Maybe create a `notifications.py` and optionally a `notifications_utils.py`, and move `signup_email_texts` and functions like `_get_confirmation_template_texts` to those modules?

registrations/notifications.py
```diff
@@ -0,0 +1,249 @@
...
+        },
+    },
+}
+
+
+def get_signup_notification_texts(signup, notification_type: SignUpNotificationType):
+    with (translation.override(signup.get_service_language_pk())):
+        confirmation_message = signup.registration.confirmation_message
+        event_name = signup.registration.event.name
+        event_type_id = signup.registration.event.type_id
```
harriris-vincit, line 151:
```suggestion+1

    event_type_id = signup.registration.event.type_id
    username = signup.first_name
```

registrations/notifications.py
```diff
@@ -0,0 +1,249 @@
...
+        username = signup.first_name
```
harriris-vincit, line 152:
```suggestion

    event_type_id = signup.registration.event.type_id
    username = signup.first_name
```
These don't need to be translated if I've understood everything correctly.

jorilindell: True that. Moved those variables outside of the `with translation.override...`

harriris-vincit (approved):
Looking good 👍

PR #744, opened 2023-10-18
LINK-1618 | Mass Email to Attending Signups Only

harriris-vincit:
### Description
When sending a mass email to all signups, the email will now be sent to only those who have the status "attending". Wait-listed signups shouldn't get the email in this case.
### Closes
[LINK-1618](https://helsinkisolutionoffice.atlassian.net/browse/LINK-1618)

[LINK-1618]: https://helsinkisolutionoffice.atlassian.net/browse/LINK-1618?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/registrations/api.py
+++ b/registrations/api.py
@@ -142,7 +142,9 @@ class RegistrationViewSet(
             responsible_signups = (
                 signup_group.responsible_signups or signup_group.signups.all()
             )
-            for signup in responsible_signups.exclude(email=None):
+            for signup in responsible_signups.exclude(email=None).filter(
+                attendee_status=SignUp.AttendeeStatus.ATTENDING
+            ):
                 message_signups.append(signup)
                 already_included_signups.add(signup.pk)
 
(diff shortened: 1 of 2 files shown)


registrations/api.py
```diff
@@ -142,7 +142,9 @@ def _get_message_signups(signups, signup_groups):
             responsible_signups = (
                 signup_group.responsible_signups or signup_group.signups.all()
             )
-            for signup in responsible_signups.exclude(email=None):
+            for signup in responsible_signups.exclude(
+                Q(email=None) | Q(attendee_status=SignUp.AttendeeStatus.WAITING_LIST)
```
nicobav, line 146: Pondering that should it be better to be explicit with the attendee_status here 🤔 Like if we would have a new status those would be included in the send as well - likely not going to happen but better to explicit imo 🧐 

`.exclude(email=None).filter(attendee_status=SingnUp.AttendeeStatus.ATTENDING)`

nicobav (approved):
Looks a OK 👍 

Added minor suggestion of changing the attendee_status to be explicitly ATTENDING for those who will get the email.

registrations/api.py
harriris-vincit: This is a valid point - I changed it to the suggested explicit form.

jorilindell (approved):
Look good! Nice that you unified some test names as well

PR #747, opened 2023-10-18
LINK-1522 | Command to pseudonymize signups

jorilindell:
## Description
Utils to pseudonymize data
Django manage command to pseudonymize signups and signup groups of past events:
- Signup:
  -  `email`, `membership_number`, `phone_number`, `street_address`, `first_name`, `last_name`, `extra_info` and `date_of_birth` field values are replaced with pseudonyms
  - `created_by` and `last_modified_by` fields are set to null
  - timestamp is stored to`pseudonymization_time`
- Signup group:
  -  `extra_info` field value is replaced with pseudonyms
  - `created_by` and `last_modified_by` fields are set to null
  - timestamp is stored to`pseudonymization_time`

Add pseudonymization to daily jobs

NOTE:
There is an open PR to add new environment variables to Azure: https://dev.azure.com/City-of-Helsinki/linkedevents/_git/linkedevents-pipelines/pullrequest/6358

## Closes
[LINK-1522](https://helsinkisolutionoffice.atlassian.net/browse/LINK-1522) 

[LINK-1522]: https://helsinkisolutionoffice.atlassian.net/browse/LINK-1522?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


registrations/pseudonymize.py
```diff
@@ -0,0 +1,157 @@
...
+
+    Parts of the date which are 0 are kept zero. (e. g. if the day is '00'
+    it is not pseudonymized)
+
+    """
+    if not value:
+        return value
+
+    for part, length, func in (("D", 2, day), ("M", 2, month), ("Y", 4, year)):
+        assert format.count(part) == length
```
nicobav, line 148: handy as the assert is, it can be suppressed with optimization mode. So checking and raising is the way here.

nicobav (changes requested):
Looks good, some minor changes and then we can 🚀

registrations/pseudonymize.py
jorilindell: Changed as suggested

registrations/management/commands/pseudonymize_past_signups.py
```diff
@@ -0,0 +1,43 @@
...
+        "Pseudonymize signups and signup groups for past events. "
+        "The time of pseudonymization can be specified in days in the "
+        "PSEUDONYMIZATION_THRESHOLD_DAYS environment variable."
+    )
+
+    def handle(self, *args, **options):
+        threshold_time = localtime() - timedelta(
+            days=settings.PSEUDONYMIZATION_THRESHOLD_DAYS
+        )
+        signup_groups = SignUpGroup.objects.filter(
```
harriris-vincit, line 21:
```suggestion
        signup_groups = SignUpGroup.objects.prefetch_related("signups").filter(
```
Since a group's pseudonymization will use `signups.all()`

jorilindell: Added the prefetch for signups

registrations/management/commands/pseudonymize_past_signups.py
```diff
@@ -0,0 +1,43 @@
...
+        "Pseudonymize signups and signup groups for past events. "
+        "The time of pseudonymization can be specified in days in the "
+        "PSEUDONYMIZATION_THRESHOLD_DAYS environment variable."
+    )
+
+    def handle(self, *args, **options):
+        threshold_time = localtime() - timedelta(
+            days=settings.PSEUDONYMIZATION_THRESHOLD_DAYS
+        )
+        signup_groups = SignUpGroup.objects.prefetch_related("signups").filter(
```
harriris-vincit, line 21: Maybe consider using [select_for_update](https://docs.djangoproject.com/en/3.2/ref/models/querysets/#select-for-update) for both of the querysets and running the pseydonymization loops inside transactions. This will prevent scenarios where another process is updating signup data concurrently and that updated value ends up as the value instead of the pseudonymized value.

jorilindell:
There is an on-going discussion should we instead of pseudonymization anomymize data. Closing this PR at this point and we can re-open it if we decide to go with pseudonymization

PR #749, opened 2023-10-19
LINK-1629 | Reduce memory footprint of osoite.py

charn:
This PR reduces the amount of memory used by `osoite.py` importer. Using an iterator makes sure the queryset does not cache it's result (which is not needed in this case).

Tested the importer locally. Memory consumption went from roughly 2,3 GB to 1,3 GB. Locally the change had hardly any effect on processing time.

### DIFF
--- a/events/importer/base.py
+++ b/events/importer/base.py
@@ -19,7 +19,6 @@ from events.models import Event, EventLink, Image, Language, Offer, Place
 from .. import utils
 from .util import clean_text, separate_scripts
 
-# Per module logger
 logger = logging.getLogger(__name__)
 
 EXTENSION_COURSE_FIELDS = (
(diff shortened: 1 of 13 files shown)


events/importer/base.py
```diff
@@ -19,7 +19,6 @@
 from .. import utils
 from .util import clean_text, separate_scripts
 
-# Per module logger
```
nicobav, line 22: goodbye Per 👋

events/importer/osoite.py
```diff
@@ -181,7 +181,8 @@ def import_places(self):
         else:
             logger.info("Loading addresses...")
             obj_list = self.pk_get("Address")
-            logger.info("%s addresses loaded" % len(obj_list))
+            logger.info(f"{obj_list.count()} addresses loaded")
```
nicobav, line 184: Formula 1 strings 🏎️  👍

nicobav (approved):
Solid 🪨

PR #752, opened 2023-10-19
LINK-1564 | Send only parent draft event created notification

nicobav:
Do not send the notification considering child events of recurring super events. 
Also do not send draft notifications from events created by admins.

Refs LINK-1564 LINK-1633

### DIFF
--- a/events/models.py
+++ b/events/models.py
@@ -1064,6 +1064,11 @@ class Event(MPTTModel, BaseModel, SchemalessFieldMixin, ReplacedByMixin):
             created
             and self.publication_status == PublicationStatus.DRAFT
             and not self.is_created_with_apikey
+            # Only send super event notification to avoid spamming from child events when recurring event.
+            and not (
+                self.super_event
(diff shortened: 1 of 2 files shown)


danipran (approved):
LGTM :+1:

events/models.py
```diff
@@ -1064,6 +1064,13 @@ def save(self, *args, **kwargs):
             created
             and self.publication_status == PublicationStatus.DRAFT
             and not self.is_created_with_apikey
+            # Only send super event notification to avoid spamming from child events when recurring event.
+            and not (
```
charn, line 1068: From what I can tell the frontend:
1. first creates all sub events (at this stage they will not have a super event) and
2. at the second phase the super event is created with previously created events as sub events.

I'm not saying that this implementation is wrong, but it might not work as expected with the current frontend at least. And I'm not even sure how an user of an API key would create a recurring event. :shrug: So, it also might be required to change how the frontend sends these recurring events.

PR #753, opened 2023-10-20
LINK-1612 | Add translations for objects

charn:
This PR adds translations for keywords added by `add_helsinki_audience` and also for `Language` instances which currently exist in production. For `Language` objects a separate new script `create_languages` is added which can create all instances which currently exists in production and is also able to update translations. For language names I tried to use translations provided by Django, but unfortunately they were not perfect so I decided to hard code the values in the script.

Translations in this PR have already been done in production, but for any new or existing environment these are missing and therefore they should also be part of the codebase.

### DIFF
--- a/docker/django/docker-entrypoint.sh
+++ b/docker/django/docker-entrypoint.sh
@@ -13,6 +13,8 @@ if [[ "$APPLY_MIGRATIONS" = "true" ]]; then
     ./manage.py migrate --noinput
     echo "Applying sync_translation_fields migrations..."
     ./manage.py sync_translation_fields --noinput
+    echo "Create language objects..."
+    ./manage.py create_languages
 fi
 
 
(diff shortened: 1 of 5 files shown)


docker/django/docker-entrypoint.sh
```diff
@@ -13,6 +13,8 @@ if [[ "$APPLY_MIGRATIONS" = "true" ]]; then
     ./manage.py migrate --noinput
     echo "Applying sync_translation_fields migrations..."
     ./manage.py sync_translation_fields --noinput
+    echo "Create languages objects..."
```
danipran, line 16:
```suggestion
    echo "Create language objects..."
```

danipran (approved):
Found a typo, otherwise LGTM :+1:

PR #755, opened 2023-10-23
Relocate free_text filter

voneiden:
free_text filter was originally probably accidentally inserted in the middle of an unrelated if/else clause, breaking filtering of deprecated events.

Ref LINK-1654


danipran (approved):
LGTM :+1:

PR #758, opened 2023-10-25
LINK-1614 | Enable restoring elis events in kulke importer

nicobav:
If the event is marked as deleted (soft deleted) and is found in elis data in the kulke importer, it's going to be restored and marked as not deleted.

Refs LINK-1614

### DIFF
--- a/events/importer/kulke.py
+++ b/events/importer/kulke.py
@@ -957,6 +957,8 @@ class KulkeImporter(Importer):
         if count:
             logger.debug("Deleted %d events", count)
 
+        self._handle_referenced_deleted_events(elis_event_ids, begin_date)
+
         # Find super events that no longer contain at least two events and delete them
         count = (
             Event.objects.exclude(super_event_type__isnull=True)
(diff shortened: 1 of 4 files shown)


events/importer/kulke.py
```diff
@@ -979,6 +979,24 @@ def _handle_removed_events(
                 count,
             )
 
+        self._handle_referenced_deleted_events(elis_event_ids, begin_date)
```
tuomas777, line 982: Is it possible that we will bring back events that in turn should prevent some super events from being deleted in the above super event soft delete thing? Ie. should this `_handle_referenced_deleted_events()` be run before line 961?

nicobav: Very nice catch, sir. I moved the restoring part above the super event deletion and added a test for this case.

tuomas777: To be exact, I was talking about a situation where there is a super event which is erroneously deleted because its sub-events haven't been restored yet. And at least the new test doesn't address that particularly 🙂 

But as said, I'm not sure if that is even a possible scenario in practice 😄

danipran (approved):
LGTM, might wanna add an `alters_data` for `undelete`

PR #760, opened 2023-10-26
Use console email backend if API_MAILGUN_KEY is not set

voneiden:
ref https://helsinkisolutionoffice.atlassian.net/browse/LINK-1657

Originally was intending to add a separate env var EMAIL_TO_CONSOLE and raise ImproperlyConfigured if none of MAILGUN_API_KEY, EMAIL_TO_CONSOLE or DEBUG was set, but that broke management.py.


tuomas777 (approved):
LGTM 👍 📫 

It would be nice if you could change also the PR's title to match the current implementation 🙂

PR #761, opened 2023-10-26
Rename config_dev.toml -> config_dev.env

voneiden:
Since the file is not and never was a toml file by content.

As this may affect your local development setup rather unexpectedly, FYI also 

* @jorilindell
* @harriris-vincit 
* @City-of-Helsinki/ratkaisutoimiston-frontend


charn (approved):
LGTM! :+1: :shipit:

PR #762, opened 2023-10-26
LINK-1658 | XLSX Export for Registration Signups

harriris-vincit:
### Description
Adds XLSX export for registration signups through a new action endpoint.
### Closes
[LINK-1658](https://helsinkisolutionoffice.atlassian.net/browse/LINK-1658)

[LINK-1658]: https://helsinkisolutionoffice.atlassian.net/browse/LINK-1658?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/registrations/api.py
+++ b/registrations/api.py
@@ -8,6 +8,8 @@ from django.contrib.auth.models import AnonymousUser
 from django.db import transaction
 from django.db.models import ProtectedError, Q, Value
 from django.db.models.functions import Concat
+from django.http import HttpResponse
+from django.utils import translation
 from django.utils.translation import gettext as _
 from rest_framework import mixins, status, viewsets
 from rest_framework.decorators import action
@@ -26,6 +28,7 @@ from events.models import Event
 from events.permissions import OrganizationUserEditPermission
 from linkedevents.registry import register_view
 from registrations.exceptions import ConflictException
+from registrations.exports import RegistrationSignUpsExportXLSX
 from registrations.models import (
     Registration,
     RegistrationUserAccess,
@@ -35,12 +38,14 @@ from registrations.models import (
 )
 from registrations.permissions import (
     CanAccessRegistration,
+    CanAccessRegistrationSignups,
     CanAccessSignup,
     CanAccessSignupGroup,
 )
 from registrations.serializers import (
     CreateSignUpsSerializer,
     MassEmailSerializer,
+    RegistrationSignupsExportSerializer,
     RegistrationUserAccessSerializer,
     SeatReservationCodeSerializer,
     SignUpGroupCreateSerializer,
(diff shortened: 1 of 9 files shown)


registrations/tests/test_signups_export.py
```diff
@@ -0,0 +1,323 @@
...
+        "regular_user",
+        "regular_user_without_organization",
+        "weakly_identified_registration_user",
+        "admin_of_another_organization",
+        "user_of_another_organization",
+        "anonymous_user",
+    ],
+)
+@pytest.mark.django_db
+def test_signup_export_forbidden_organization_user_roles(
```
nicobav, line 89: This test has many conditions. It would be good to separate those conditions into different tests rather than have complex testing logic.

nicobav: I see that there's same like situation in other tests as well. So splitting conditions to own tests or where can parametrize its coolio also 😎

registrations/api.py
```diff
@@ -221,6 +227,44 @@ def send_message(self, request, pk=None, version=None):
...
+        methods=["get"],
+        detail=True,
+        permission_classes=[CanAccessRegistration],
+        url_path=r"signups/export/(?P<file_format>xlsx)",
+    )
+    def signups_export(self, request, file_format=None, pk=None, version=None):
+        user = request.user
+        registration = self.get_object()
+
+        if not user.is_authenticated or not (
```
nicobav, line 240: Couldn't we move this checking logic to `CanAccessRegistration`?

nicobav (changes requested):
Looks good, some tiny tidying ups 🧹

registrations/api.py
harriris-vincit: It would be a bit problematic as these checks only apply for the registrations' signups (a user requires additional permissions if they also want to view/manipulate the registrations' signups). Also, a registration is allowed to be fetched even by a non-authenticated user while signups are not.

registrations/tests/test_signups_export.py
harriris-vincit: True - I'll split these up a bit to simplify the logic

registrations/api.py
harriris-vincit: Ah but now I see what you meant: through `view.action` that would of course be possible. Another alternative could be to introduce an additional permission class for this case. I think I'll do some changes here

harriris-vincit: Added a new permission class that is also used in the registration base serializer that has a similar check as this action endpoint.

harriris-vincit:
Closed at least temporarily as there is a need to clarify if printing the information should be allowed in the first place. The changes should not be merged at this point. Once this is clarified, either the PR will be re-opened or the feature branch removed completely.

PR #763, opened 2023-10-27
Add pytest-freezer (freeze time marker), fix broken tests

danipran:
A couple tests are temporarily broken because of DST stuff. Should start working next week... or after this fix. Added xfail tests for the DST stuff; might have an actual DST related bug hiding here.


charn (approved):
LGTM! :+1: :shipit:

PR #764, opened 2023-10-27
LINK-1593: Escape regex input

danipran:
Using unescaped user input (via query strings) in regexes causes errors and is a security vulnerability. This PR escapes the input before using it in regex.

Also wrote some unit tests for `_terms_to_regex`. Apparently, the OR operator is completely busted, which I'm not going to address in this PR, so marked the OR cases with xfail.

### DIFF
--- a/events/api.py
+++ b/events/api.py
@@ -2619,7 +2619,8 @@ def _terms_to_regex(terms, operator):
             e = 1
         else:
             e = 2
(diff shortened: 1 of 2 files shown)


tuomas777 (approved):
I don't think that is an actual security issue in this case, but of course not something we want to allow. Well spotted! 🥇

PR #766, opened 2023-10-27
Increase Language.name to 100 chars

nicobav:
Refs LINK-841


nicobav:
Project decided to bail out from this change, for now at least so closing this.

PR #767, opened 2023-10-30
Full-text search using materialized view PoC

voneiden:
Ref: https://helsinkisolutionoffice.atlassian.net/browse/LINK-1655

I decided to put the SQL in a separate file as the developer experience is much better compared to when it is inlined in a python script. The downside is that I wasn't able to template the weights, so now there's a regex substitution that applies environment variables. 

The qp's are currently `x_full_text` and the supplementary `x_ongoing`.

Additional material: 
* https://helsinkisolutionoffice.atlassian.net/browse/LINK-1536
* https://helsinkisolutionoffice.atlassian.net/browse/LINK-1389


events/management/commands/refresh_full_text.py
```diff
@@ -0,0 +1,140 @@
...
+            else:
+                self.stdout.write(
+                    f"Applied weight override for field {field} to {weight_override}"
+                )
+            sql = replaced_sql
+        return sql
+
+    def handle(self, *args, **options):
+        if options["drop"]:
+            # Print stdout success info
```
charn, line 65: Unnecessary comment?

charn (approved):
LGTM! :+1: :shipit:

events/management/commands/refresh_full_text.py
voneiden: copilot steering attempt :face_with_head_bandage:

voneiden:
Some test tweaking was needed due to main changes. Also changed `language` -> `x_full_text_language`

PR #770, opened 2023-11-01
LINK-1690 | Allow registration admins to update registration

jorilindell:
## Description
At the moment registration admins are not allowed to edit registrations. The reason is that `CanAccessRegistration` `has_object_permission` method calls `Event.can_be_edited_by` instead of `Registration.can_be_edited_by` when `RegistrationViewSet` `perform_create` or `perform_update` methods check permissions by using `self.check_object_permissions(self.request, event)`. 

Fix this by moving event data source permission check to create method of RegistrationSerializer.
Also make event read_only field when updating the registrations. In that case there is no need for the event data source permissions check anymore.

## Closes
[LINK-1690](https://helsinkisolutionoffice.atlassian.net/browse/LINK-1690)

[LINK-1690]: https://helsinkisolutionoffice.atlassian.net/browse/LINK-1690?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/registrations/models.py
+++ b/registrations/models.py
@@ -265,13 +265,8 @@ class Registration(CreatedModifiedBaseModel):
         """Check if current registration can be edited by the given user"""
         return (
             user.is_superuser
-            or user.is_admin_of(self.event.publisher)
-            or user.is_registration_admin_of(self.event.publisher)
-        )
-
-    def is_user_editable_resources(self):
-        return bool(
-            self.event.data_source and self.event.data_source.user_editable_resources
+            or user.is_admin_of(self.publisher)
+            or user.is_registration_admin_of(self.publisher)
         )
 
 
--- a/events/api.py
+++ b/events/api.py
@@ -1844,6 +1844,10 @@ class RegistrationSerializer(LinkedEventsSerializer, RegistrationBaseSerializer)
     ):
         super().__init__(*args, **kwargs)
         context = self.context
+        instance = self.instance
+
+        if instance:
+            self.fields["event"].read_only = True
 
         self.registration_admin_tree_ids = context.get(
             "registration_admin_tree_ids", set()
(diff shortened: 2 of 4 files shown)


registrations/permissions.py
```diff
@@ -48,7 +48,10 @@ def has_object_permission(
             if obj.data_source != user_data_source:
                 return False
 
-        return obj.can_be_edited_by(request.user)
+        # check_object_permissions is called from RegistrationViewSet perform_create and
+        # perform_update methods with event value, so use Registration.can_be_edited_by
+        # instead of obj.can_be_edited_by to use correct can_be_edited_by method
+        return Registration.can_be_edited_by(obj, request.user)
```
charn, line 54: Just a note that the signature of this method is wrong, since `obj` is not necessarily of type `Registration`.

Could the `perform_update` actually be deleted since, since by default `get_object` will call `check_object_permissions` with the `Registration` instance. Or is it possible to update the event field on a registration?

Maybe for the `perform_create` instead of calling `check_object_permissions` we could either try to improve the POST check in `CanAccessRegistration.has_permission` or just do the required check in `perform_create` or in the serializer's `create` method.

nicobav: also some test(s) for the case would be awesome to validate the solution.

jorilindell: Great comment! 

I did following changed to get rid of the hack:
- Check event data source for api key user in `RegistrationSerializer` `create` method
- Make `event` read_only field in `RegistrationSerializer` `__init__` method. In that case there is no need for the permission check of the event object.

registrations/models.py
```diff
@@ -265,14 +265,12 @@ def can_be_edited_by(self, user):
         """Check if current registration can be edited by the given user"""
         return (
             user.is_superuser
-            or user.is_admin_of(self.event.publisher)
-            or user.is_registration_admin_of(self.event.publisher)
+            or user.is_admin_of(self.publisher)
+            or user.is_registration_admin_of(self.publisher)
         )
 
     def is_user_editable_resources(self):
```
harriris-vincit, line 272: This `Registration` method does not seem to be used anywhere - can be removed most probably.

jorilindell: Removed the unused method.

harriris-vincit (approved):
LGTM 👍

PR #773, opened 2023-11-02
LINK-1692 | Event.type_id: index the field and make it an integer

charn:
`Event.type_id` is an un-indexed `CharField`. The values that are saved to it come from `TypeId`, which holds integers. This type casting causes unnecessary database updates in importers since the value usually doesn't actually change. The field is also used in queries for filtering so indexing is a good idea.

Refs LINK-1692


nicobav (approved):
Very nice 🚀

jorilindell (approved):
Very nice improvement

PR #775, opened 2023-11-03
LINK-1660 | Prevent removing the only responsible person of a group

harriris-vincit:
### Description
Prevents deleting or removing through editing the only responsible person of a `SignUpGroup`. Such a person is technically a `SignUp` with `responsible_for_group=True`.
### Closes
[LINK-1660](https://helsinkisolutionoffice.atlassian.net/browse/LINK-1660)

[LINK-1660]: https://helsinkisolutionoffice.atlassian.net/browse/LINK-1660?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/registrations/serializers.py
+++ b/registrations/serializers.py
@@ -208,6 +208,15 @@ class SignUpSerializer(CreatedModifiedBaseSerializer):
                 ):
                     errors["date_of_birth"] = _("The participant is too old.")
 
+        if (
+            data.get("responsible_for_group") is False
+            and self.instance
+            and self.instance.is_only_responsible_signup
+        ):
+            errors["responsible_for_group"] = _(
+                "Cannot set responsible_for_group to False for the only responsible person of a group"
+            )
+
         if errors:
             raise serializers.ValidationError(errors)
 
--- a/registrations/tests/test_models.py
+++ b/registrations/tests/test_models.py
@@ -230,3 +230,18 @@ class TestSignUp(TestCase):
         self.signup.save()
 
         self.assertEqual(self.signup.get_service_language_pk(), "fi")
+
+    def test_is_only_responsible_signup(self):
+        group = SignUpGroupFactory(registration=self.signup.registration)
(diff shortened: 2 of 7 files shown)


registrations/tests/test_models.py
```diff
@@ -230,3 +230,18 @@ def test_get_default_service_language_pk(self):
...
+
+        self.signup.signup_group = group
+        self.signup.save(update_fields=["signup_group"])
+
+        self.assertFalse(self.signup.is_only_responsible_signup)
+
+        self.signup.responsible_for_group = True
+        self.signup.save(update_fields=["responsible_for_group"])
+
+        del self.signup.is_only_responsible_signup
```
charn, line 246: Is the purpose of this the same as doing `self.signup.refresh_from_db()`?

registrations/serializers.py
```diff
@@ -208,6 +208,15 @@ def validate(self, data):
                 ):
                     errors["date_of_birth"] = _("The participant is too old.")
 
+        if (
+            data.get("responsible_for_group") is False
```
nicobav, line 212: PEP8 states this should be written as: `not data.get("responsible_for_group")`

nicobav (approved):
LGTM 👍 

(One tiny pep8 conditional rephrase )

registrations/tests/test_models.py
harriris-vincit: Yes, same purpose: to refresh this cached property

registrations/serializers.py
harriris-vincit: `not data.get("responsible_for_group")` will falsely give the validation error if `responsible_for_group` is not given in the data on purpose

nicobav: Ahh, I buy that 🤝

PR #779, opened 2023-11-06
Ensure last_modified_time is updated when an event is saved

voneiden:
Especially in case of soft deletion by importers
save(update_fields=[...]) would be called, which leaves last_modified_time untouched. This commit changes the updating of last_modified_time as opt-out. The last_modified_time is not updated when n_events is updated.

### DIFF
--- a/events/models.py
+++ b/events/models.py
@@ -366,6 +366,24 @@ class BaseModel(models.Model):
     def is_user_edited(self):
         return bool(self.data_source.user_editable_resources and self.last_modified_by)
 
+    def save(self, update_fields=None, skip_last_modified_time=False, **kwargs):
+        # When saving with update_fields, django will not automatically update
+        # last_modified_time.
+        if (
+            not skip_last_modified_time
+            and update_fields is not None
+            and "last_modified_time" not in update_fields
+        ):
+            update_fields = list(update_fields) + ["last_modified_time"]
+            if (
+                self.last_modified_by_id is not None
+                and "last_modified_by" not in update_fields
+            ):
+                self.last_modified_by = None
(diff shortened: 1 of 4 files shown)


events/models.py
```diff
@@ -366,6 +366,24 @@ def is_user_editable_resources(self):
...
+            not skip_last_modified_time
+            and update_fields is not None
+            and "last_modified_time" not in update_fields
+        ):
+            update_fields = list(update_fields) + ["last_modified_time"]
+            if (
+                self.last_modified_by_id is not None
+                and "last_modified_by" not in update_fields
+            ):
+                self.last_modified_by = None
```
charn, line 382: Is this assignment really needed? Assuming the `pre_save` for the `DateTimeField` gets called this shouldn't be needed, since the field has `auto_now`.

charn: Though why are we touching `last_modified_by`? It's not mentioned in the commit message or in the ticket.

voneiden: Right, consider the following scenario:

1) `User A` modifies an event that occurs in 2023  in 2023 and the fields reflect this
2) 5 years later in 2028 `Admin B` soft  deletes all events from 2023 in shell with Event.objects.filter(end_time__lte=2024).soft_delete()
3) last_modified_time should represent the soft deletion time
4) last_modified_by represents now what?

Ideally of course it would be `Admin B`, but in this case the information is not available. Leaving it as `User A` is IMO misleading, as it implies `User A` has modified the event in 2028

charn: Fair enough. :+1:

charn (approved):
LGTM! :+1: :shipit:

PR #786, opened 2023-11-08
LINK-1707 | Use bulk_create when creating multiple signups

harriris-vincit:
### Description
Use `bulk_create` when creating multiple signups and their protected data in the `signup` and `signup_group` endpoints. 

Additionally, check the attendee and waiting list capacities only once instead of for each signup separately.

### Closes
[LINK-1707](https://helsinkisolutionoffice.atlassian.net/browse/LINK-1707)

[LINK-1707]: https://helsinkisolutionoffice.atlassian.net/browse/LINK-1707?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/registrations/serializers.py
+++ b/registrations/serializers.py
@@ -35,6 +35,29 @@ def _validate_registration_enrolment_times(registration: Registration) -> None:
         raise ConflictException(_("Enrolment is already closed."))
 
 
+def _validate_registration_capacity(
+    registration: Registration, signups_count: int
+) -> tuple[int, int]:
+    attendee_capacity = registration.maximum_attendee_capacity
+    waiting_list_capacity = registration.waiting_list_capacity
+    add_as_attending = add_as_waitlisted = signups_count
+
+    if attendee_capacity is not None:
+        already_attending = SignUp.objects.filter(
+            registration=registration, attendee_status=SignUp.AttendeeStatus.ATTENDING
+        ).count()
+        add_as_attending = attendee_capacity - already_attending
+
+    if waiting_list_capacity is not None:
+        already_waitlisted = SignUp.objects.filter(
+            registration=registration,
+            attendee_status=SignUp.AttendeeStatus.WAITING_LIST,
+        ).count()
+        add_as_waitlisted = waiting_list_capacity - already_waitlisted
+
+    return max(add_as_attending, 0), max(add_as_waitlisted, 0)
+
+
 def _get_protected_data(validated_data: dict, keys: list[str]) -> dict:
     return {key: validated_data.pop(key) for key in keys if key in validated_data}
 
@@ -100,17 +123,11 @@ class SignUpSerializer(CreatedModifiedBaseSerializer):
             validated_data, ["extra_info", "date_of_birth"]
         )
 
-        already_attending = SignUp.objects.filter(
-            registration=registration, attendee_status=SignUp.AttendeeStatus.ATTENDING
-        ).count()
-        already_waitlisted = SignUp.objects.filter(
-            registration=registration,
-            attendee_status=SignUp.AttendeeStatus.WAITING_LIST,
-        ).count()
-        attendee_capacity = registration.maximum_attendee_capacity
-        waiting_list_capacity = registration.waiting_list_capacity
+        add_as_attending, add_as_waitlisted = _validate_registration_capacity(
+            registration, 1
+        )
 
-        if (attendee_capacity is None) or (already_attending < attendee_capacity):
+        if add_as_attending:
             signup = super().create(validated_data)
             self._update_or_create_protected_data(signup, **protected_data)
 
@@ -118,9 +135,7 @@ class SignUpSerializer(CreatedModifiedBaseSerializer):
                 signup.send_notification(SignUpNotificationType.CONFIRMATION)
 
             return signup
-        elif (waiting_list_capacity is None) or (
-            already_waitlisted < waiting_list_capacity
-        ):
+        elif add_as_waitlisted:
             validated_data["attendee_status"] = SignUp.AttendeeStatus.WAITING_LIST
 
             signup = super().create(validated_data)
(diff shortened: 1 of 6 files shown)


registrations/tests/test_signup_post.py
```diff
@@ -706,6 +749,140 @@ def test_group_signup_successful_with_waitlist(user_api_client, registration, us
...
+        (0, 0, 0, 0, 0, status.HTTP_403_FORBIDDEN),
+        (1, 1, 2, 1, 1, status.HTTP_201_CREATED),
+        (1, 0, 1, 1, 0, status.HTTP_201_CREATED),
+        (0, 1, 1, 0, 1, status.HTTP_201_CREATED),
+        (2, 1, 2, 2, 0, status.HTTP_201_CREATED),
+        (0, 2, 2, 0, 2, status.HTTP_201_CREATED),
+        (None, None, 2, 2, 0, status.HTTP_201_CREATED),
+        (None, 1, 2, 2, 0, status.HTTP_201_CREATED),
+        (1, None, 2, 1, 1, status.HTTP_201_CREATED),
+        (0, None, 2, 0, 2, status.HTTP_201_CREATED),
```
voneiden, line 765: Nice parametrization :+1:

registrations/serializers.py
```diff
@@ -35,6 +35,32 @@ def _validate_registration_enrolment_times(registration: Registration) -> None:
...
+        already_waitlisted = SignUp.objects.filter(
+            registration=registration,
+            attendee_status=SignUp.AttendeeStatus.WAITING_LIST,
+        ).count()
+        add_as_waitlisted = waiting_list_capacity - already_waitlisted
+
+    if not (add_as_attending or add_as_waitlisted):
+        raise DRFPermissionDenied(_("The waiting list is already full"))
+
+    return max(add_as_attending, 0), max(add_as_waitlisted, 0)
```
voneiden, line 61: I like the use of max here :+1:  :+1:

registrations/serializers.py
```diff
@@ -35,6 +35,32 @@ def _validate_registration_enrolment_times(registration: Registration) -> None:
...
+
+    if waiting_list_capacity is not None:
+        already_waitlisted = SignUp.objects.filter(
+            registration=registration,
+            attendee_status=SignUp.AttendeeStatus.WAITING_LIST,
+        ).count()
+        add_as_waitlisted = waiting_list_capacity - already_waitlisted
+
+    if not (add_as_attending or add_as_waitlisted):
+        raise DRFPermissionDenied(_("The waiting list is already full"))
```
voneiden, line 59: Two thoughts

1) I suppose `add_as_attending` and `add_as_waitlisted` can, at least in theory, be negative numbers if the number of attendees/waitlisters exceeds the set capacity (can capacity be modified to be less than the number of existing items?). A negative number is truthy, so it would sneak past this if case. Not a big problem, as the create() will not do anything, except maybe result in error 500 instead of 403. 

2) I wouldn't mind keeping this where it was before, at the end of the view. Now the view has `if add_as_attending` and `elif add_as_waitlisted`, but it is not immediately obvious what the "else" behaviour is until you take a look at this function. If that makes sense?

registrations/serializers.py
```diff
@@ -125,8 +143,6 @@ def create(self, validated_data):
 
             return signup
 
-        raise DRFPermissionDenied(_("The waiting list is already full"))
```
voneiden, line 128: Referring to the comment above, I think this was a better place to raise the exception in terms of readability.

registrations/serializers.py
```diff
@@ -433,6 +449,82 @@ def validate(self, data):
...
+
+            cleaned_signup_data = signup_data.copy()
+            cleaned_signup_data["created_by"] = user
+            cleaned_signup_data["last_modified_by"] = user
+
+            extra_info = cleaned_signup_data.pop("extra_info", None)
+            date_of_birth = cleaned_signup_data.pop("date_of_birth", None)
+
+            check_and_set_attendee_status(cleaned_signup_data)
+            capacity_by_attendee_status_map[cleaned_signup_data["attendee_status"]] -= 1
```
voneiden, line 508: Here too, is it possible that one of these two values slips negative and messes up the loop condition above?

registrations/serializers.py
```diff
@@ -433,6 +449,82 @@ def validate(self, data):
...
+
+        signups = []
+        protected_data = []
+
+        for signup_data in validated_data["signups"]:
+            if not (
+                capacity_by_attendee_status_map[SignUp.AttendeeStatus.ATTENDING]
+                or capacity_by_attendee_status_map[SignUp.AttendeeStatus.WAITING_LIST]
+            ):
+                continue
```
voneiden, line 498: break?

voneiden (changes requested):
Looks good, one note on possible issue with readability and a few concerns regarding unexpected negative values.

registrations/tests/test_signup_post.py
harriris-vincit: It's rather pleasing to the eyes

registrations/serializers.py
harriris-vincit: I think this is a very good suggestion and the negative numbers being truthy slipped my mind completely. Now improvements have been made based on these points.

registrations/serializers.py
harriris-vincit: Added "greater than 0" checks to the beginning of the loop to avoid the issue.

registrations/serializers.py
harriris-vincit: Yes indeed, that is better and has been changed as suggested now.

registrations/serializers.py
harriris-vincit: Changed back to how it was.

voneiden (approved):
Excellent, thanks!

voneiden (approved):
Excellent, thanks!

PR #788, opened 2023-11-10
LINK-1708 | Make seat reservation code unique per registration, optimize signup filtering

harriris-vincit:
### Description

- Make `SeatReservationCode` unique per registration
- Add index for `SignUp.attendee_status` as it is used in several places with an exact search (including signup list filtering)
- Optimize the registration-based signup filtering a little bit (use cached registration admin organizations, only do further filtering if the user is not a superuser, don't query registrations in a for loop)

Indexes were not added for the signup text search fields as the queries are done with `icontains` instead of exact or `startswith` searches.
### Closes
[LINK-1708](https://helsinkisolutionoffice.atlassian.net/browse/LINK-1708)

[LINK-1708]: https://helsinkisolutionoffice.atlassian.net/browse/LINK-1708?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


registrations/api.py
```diff
@@ -277,36 +282,28 @@ def filter_queryset(self, queryset):
...
-            try:
-                registration = Registration.objects.select_related(
-                    "event__publisher"
-                ).get(pk=pk)
-            except (ValueError, Registration.DoesNotExist):
-                raise ParseError(
-                    _("Registration with id {pk} doesn't exist.").format(pk=pk)
-                )
+        if not registrations:
+            raise ParseError(_("No registrations found with the given IDs."))
```
voneiden, line 291: Yeaah LE is using ParseError so much, using it qualifies as a code convention at this point, but if we're pedantic, it's meant for situations where the request is malformed and cannot be parsed (like invalid JSON). 

ValidationError might be more appropriate, like 

```python
raise ValidationError({
    "registration": _("No registrations found with the given IDs.")
})
 ```

Double checking that raising an error explicitly is really preferred here over just returning empty, `"count": 0`, page? If yes, should remember to document it in the swagger (if these endpoints are added there ?).

File: registrations/api.py [hunk 2]
```diff
@@ -277,36 +282,28 @@ def filter_queryset(self, queryset):
...
-                )
-                or registration.publisher in registration_admin_organizations
-            ):
-                raise DRFPermissionDenied(
-                    _(
-                        "Only the admins of the organization {organization} have access rights."
-                    ).format(organization=registration.publisher)
-                )
+        if user.is_superuser:
+            return queryset.filter(**{f"{name}__in": registrations})
```
voneiden, line 294: I don't think the dynamic nature of `name` is needed, at least in the current state, here or below on line 308. 

```python
return queryset.filter(registration__in=registrations})
```

would be quite a bit more readable (and searchable!).

registrations/api.py
```diff
@@ -277,36 +282,28 @@ def filter_queryset(self, queryset):
...
+            qs_filter |= Q(registration_user_accesses__email=user.email)
+        registrations = registrations.filter(qs_filter)
 
-            registrations.append(registration)
+        if not registrations:
+            raise DRFPermissionDenied(
+                _(
+                    "Only the admins of the registration organizations have access rights."
+                )
+            )
```
voneiden, line 306: It's unfortunate that this fits so well right here where it is. On a fundamental level I maintain a position that permission checks should be performed before the view in permission class(es). Just food for thought, not requesting this to be changed as it's probably non-trivial. Plus LE api has a tradition of raising PermissionDenied all over the place :laughing:

voneiden (changes requested):
Left couple of questions/thoughts, feel free to address as you best see fit :+1:

registrations/api.py
```diff
@@ -277,36 +282,28 @@ def filter_queryset(self, queryset):
...
         user = self.request.user
 
-        # Get admin organizations and descendants only once instead of using is_admin method
-        registration_admin_organizations = (
-            user.get_registration_admin_organizations_and_descendants()
-        )
+        try:
+            registrations = Registration.objects.filter(pk__in=value)
+        except ValueError:
+            registrations = Registration.objects.none()
```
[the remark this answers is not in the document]
harriris-vincit: Any non-numeric string value such as `/v1/signup/?registration=foo` will result in `ValueError: Field 'id' expected a number but got 'foo'`

registrations/api.py
harriris-vincit: There is actually no proper reason for not returning an empty qs instead of raising the exception other than that I didn't want to touch the existing logic too much. But this makes a lot of sense so I changed this to return the empty qs.

registrations/api.py
harriris-vincit: Changed as suggested

registrations/api.py
harriris-vincit: Agree with both non-triviality and that the perms should be checked in permission classes.

registrations/api.py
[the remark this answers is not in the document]
voneiden: Right you are

voneiden (approved):

voneiden:
LGTM!

PR #789, opened 2023-11-10
LINK-1716 | Add dissolved filter to Organization api

nicobav:
Adds boolean filter for including or excluding (true/false) dissolved organizations from the result.

Refs LINK-1716

### DIFF
--- a/events/api.py
+++ b/events/api.py
@@ -1426,6 +1426,26 @@ class LanguageViewSet(
 register_view(LanguageViewSet, "language")
 
 
+class OrganizationFilter(django_filters.rest_framework.FilterSet):
+    dissolved = django_filters.BooleanFilter(
+        method="filter_dissolved",
+    )
+
+    class Meta:
+        model = Organization
+        fields = ("dissolved",)
+
+    def filter_dissolved(self, queryset, name, value: Optional[bool]):
+        today = timezone.now().date()
+
+        if value:
+            return queryset.filter(dissolution_date__lte=today)
+
+        return queryset.filter(
+            Q(dissolution_date__gt=today) | Q(dissolution_date__isnull=True)
+        )
+
+
 class OrganizationBaseSerializer(LinkedEventsSerializer):
     view_name = "organization-detail"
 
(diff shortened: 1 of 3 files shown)


events/api.py
```diff
@@ -1416,6 +1416,23 @@ class LanguageViewSet(JSONAPIViewMixin, viewsets.ReadOnlyModelViewSet):
 register_view(LanguageViewSet, "language")
 
 
+class OrganizationFilter(django_filters.rest_framework.FilterSet):
+    dissolved = django_filters.BooleanFilter(
+        widget=django_filters.widgets.BooleanWidget(),
```
charn, line 1421: `BooleanWidget` is the default, so there shouldn't any need to add it here.

charn:
The new filter needs to be added to `organization_list.html` probably also to https://github.com/City-of-Helsinki/api-linked-events/

events/api.py
```diff
@@ -1416,6 +1416,23 @@ class LanguageViewSet(JSONAPIViewMixin, viewsets.ReadOnlyModelViewSet):
 register_view(LanguageViewSet, "language")
 
 
+class OrganizationFilter(django_filters.rest_framework.FilterSet):
+    dissolved = django_filters.BooleanFilter(
```
charn, line 1420: I think the name is a bit ambiguous. Now `dissolved=true` returns all organizations, but `dissolved=false` filters them out. Is that the way the filter is supposed to work? A better name in that case would probably be `include_dissolved`, but then again that doesn't sound quite right, since by default they're already included.

Maybe reverse the logic and `hide_dissolved` would be a better filter?

charn (changes requested):
I think rename and reversing the logic on the new filter is required.

events/api.py
tuomas777: While `hide_dissolved` is of course perfectly fine, in cases like this my favourite would be `dissolved=true|false` where `true` means only the dissolved ones. That way all three cases (dissolved, not dissolved, both) would be covered with the single boolean filter. 

With just `hide_dissolved` it is not possible to get only the dissolved ones, and adding support for that later is a bit cumbersome with the `hide_dissolved` filter already being there. And being more general-purpose and easily extendable is a notch more important in these few APIs of ours that actually have also some outside users 😄

nicobav: Yes, was thinking about it to work like @tuomas777 said. I'd say I'll tweak the logic to include only dissolved ones when `true` and exclude dissolved when `false`.
Yes? 👍 👎

charn (approved):
LGTM! :+1: :shipit:

PR #791, opened 2023-11-14
LINK-1522 | Command to anonymize signup and signup group data

jorilindell:
## Description
Django manage command to anonymize signups and signup groups of registrations with past enrolment time
- ContactPerson
  - `email`, `membership_number`, `phone_number`, `first_name` and `last_name` field values are replaced with fake values
- Signup:
  -  `street_address`, `zipcode`, `city`, `first_name`, `last_name`, `extra_info` and `date_of_birth` field values are replaced with fake data
  - `created_by` and `last_modified_by` fields are set to null
  - timestamp is stored to`anonymization_time`
- Signup group:
  -  `extra_info` field value is replaced with fake data
  - `created_by` and `last_modified_by` fields are set to null
  - timestamp is stored to`anonymization_time`

Add anonymization to daily jobs

NOTE: https://github.com/City-of-Helsinki/linkedevents/pull/772 should be merged before this PR

## Closes
[LINK-1522](https://helsinkisolutionoffice.atlassian.net/browse/LINK-1522)


helevents/service_tasks/daily_tasks.sh
```diff
@@ -25,6 +25,11 @@ if [ $? -ne 0 ]; then
     echo "PKS osoiteluettelo update signaled failure"
 fi
 
+timeout --preserve-status -s INT 60m python manage.py anonymize_past_signups
+if [ $? -ne 0 ]; then
+    echo "Past signups anonymization failure"
+fi
+
```
tuomas777, line 32: Actually this whole script [has been moved to the Azure repo](https://dev.azure.com/City-of-Helsinki/linkedevents/_git/linkedevents-pipelines/pullrequest/6825), so this change needs to be added there after this PR is merged. I guess it doesn't hurt to have it in this file as well especially now that it's already added, but unless some issues come up, we are removing this whole file soon 🙂

registrations/models.py
```diff
@@ -328,6 +329,10 @@ class SignUpGroup(CreatedModifiedBaseModel, SignUpMixin, SerializableMixin):
     registration = models.ForeignKey(
         Registration, on_delete=models.PROTECT, related_name="signup_groups"
     )
+    anonymization_time = models.DateTimeField(
+        verbose_name=_("Anonymization time"),
+        null=True,
+    )
```
tuomas777, line 335:
```suggestion
    anonymization_time = models.DateTimeField(
        verbose_name=_("Anonymization time"),
        null=True,
        editable=False,
    )
```
I think it would be safer to have `editable=False` here to prevent editing the value via Django admin.

registrations/models.py
```diff
@@ -345,6 +350,30 @@ def signups_count(self):
     def attending_signups(self):
         return self.signups.filter(attendee_status=SignUp.AttendeeStatus.ATTENDING)
 
+    @transaction.atomic
+    def anonymize(self):
+        faker = Faker("fi_FI")
```
tuomas777, line 355: There might be some business or other reason I'm not aware of, but in general in my opinion it would be better to use values that cannot be mistaken for real values, like `<DELETED>` or something.

registrations/models.py
```diff
@@ -525,6 +554,11 @@ class PresenceStatus:
         default=False,
     )
 
+    anonymization_time = models.DateTimeField(
+        verbose_name=_("Anonymization time"),
+        null=True,
+    )
+
```
tuomas777, line 561:
```suggestion
    anonymization_time = models.DateTimeField(
        verbose_name=_("Anonymization time"),
        null=True,
        editable=False,
    )

```

registrations/models.py
```diff
@@ -345,6 +350,30 @@ def signups_count(self):
...
+        faker = Faker("fi_FI")
+
+        # Allow to anonymize signup only once
+        if self.anonymization_time is None:
+            if contact_person := getattr(self, "contact_person", None):
+                contact_person.anonymize()
+
+            if protected_data := getattr(self, "protected_data", None):
+                protected_data.extra_info = faker.paragraph(nb_sentences=1)
+                protected_data.save()
```
tuomas777, line 364: Hmm, is there a reason why protected data doesn't have `anonymize()` method like other models that can be anonymized? If not, I would like it to be added for consistency reasons.

helevents/service_tasks/daily_tasks.sh
jorilindell: Oh yes, you told in some daily that you guys are moving cron jobs to Azure. I will create a separate PR there after this PR is merged.

registrations/models.py
jorilindell: Good point. Added `editable=False` to `anonymization_time` field

registrations/models.py
jorilindell: Added `editable=False` to `anonymization_time` field

registrations/models.py
jorilindell: I don't see any business reason to have fake data. I changed it to `<DELETED>` to make it easier to see that real data is anonymized.

linkedevents/settings.py
```diff
@@ -658,7 +659,8 @@ def dummy_haystack_connection_for_lang(language_code):
 
 # A list of hex-encoded 32 byte keys used for encrypting sensitive data
 FIELD_ENCRYPTION_KEYS = env("FIELD_ENCRYPTION_KEYS")
-
+# Specify the number of days after which signup and signup groups will be anonymized
```
harriris-vincit, line 662: A minor thing: adding an empty line before and after the new setting and its comment would improve the readability IMO

registrations/models.py
```diff
@@ -345,6 +351,28 @@ def signups_count(self):
     def attending_signups(self):
         return self.signups.filter(attendee_status=SignUp.AttendeeStatus.ATTENDING)
 
+    @transaction.atomic
+    def anonymize(self):
+        # Allow to anonymize signup only once
```
harriris-vincit, line 356:
```suggestion
        # Allow to anonymize signup group only once
```

registrations/models.py
```diff
@@ -345,6 +351,28 @@ def signups_count(self):
...
+        if self.anonymization_time is None:
+            if contact_person := getattr(self, "contact_person", None):
+                contact_person.anonymize()
+
+            if protected_data := getattr(self, "protected_data", None):
+                protected_data.extra_info = anonymize_replacement
+                protected_data.save()
```
harriris-vincit, line 363:
```suggestion
                protected_data.anonymize()
```

registrations/models.py
jorilindell: Updated the comment

registrations/models.py
jorilindell: Thanks, missed that one. Changed as suggested

linkedevents/settings.py
jorilindell: Added empty lines before and after to improve readability

tuomas777 (approved):
Yes 💯

PR #792, opened 2023-11-15
build: remove build trigger from azure-pipelines-test.yml

terovirtanen:
Build pipeline trigger dev and test deployments. This mess up deployments because of LATEST_BUILD_TAG


tuomas777 (approved):
👍

PR #795, opened 2023-11-16
LINK-1658 | Add XLSX export for registration signups

harriris-vincit:
### Description
Adds an XLSX export for a registration's signups. The export can be done through a new action endpoint `/registration/<pk>/signups/export/`.
### Closes
[LINK-1658](https://helsinkisolutionoffice.atlassian.net/browse/LINK-1658)


registrations/permissions.py
```diff
@@ -51,6 +51,30 @@ def has_object_permission(
         return obj.can_be_edited_by(request.user)
 
 
+class CanAccessRegistrationSignups(
+    UserDataFromRequestMixin, permissions.BasePermission
+):
+    def has_permission(self, request: Request, view: APIView) -> bool:
```
voneiden, line 57: Trivial note: `view: APIView` would benefit from a more specific type. The permissions.py has the following mypy errors for APIView:

error: "APIView" has no attribute "action"  [attr-defined]
error: "APIView" has no attribute "user_data_source_and_organization"  [attr-defined]

One workaround could be to define a Union type like

```python
from rest_framework.viewsets import GenericViewSet
from events.api import UserDataSourceAndOrganizationMixin

GenericRegistrationViewSet: Union[UserDataSourceAndOrganizationMixin, GenericViewSet]
```

But yeah anyway, no changes requested.

voneiden (approved):
LGTM. 

xlsxwriter looks way nicer to work with than openpyxl when generating xlsx.

registrations/permissions.py
voneiden: My suggestion apparently does catch fire in practice `ImportError: cannot import name 'UserDataSourceAndOrganizationMixin' from partially initialized module 'events.api'`

PR #796, opened 2023-11-16
Start using conventional commits

tuomas777:
From now on, all new commit messages must adhere to the [conventional commits spec](https://www.conventionalcommits.org/en/v1.0.0/) and their max line length is limited to 72 characters. The rules are enforced with [commitlint](https://commitlint.js.org/#/), which is run as a pre-commit hook and also as a stage in the GitHub CI pipeline.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -98,6 +98,9 @@ jobs:
           linters: isort
           run: isort . --check-only --diff
 
+      - name: Check commitlint
+        uses: wagoid/commitlint-github-action@0d749a1a91d4770e983a7b8f83d4a3f0e7e0874e  # v5.4.4
+
       # With pytest-cov use: --cov=.
       # Some errors, skipping --doctest-modules for now
       - name: Run tests
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,8 @@
 # Keep tool versions in sync with the versions in requirements-dev.txt
 default_language_version:
     python: python3
+default_install_hook_types: [pre-commit, commit-msg]
+default_stages: [pre-commit, manual]
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v4.5.0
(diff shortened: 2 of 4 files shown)


.pre-commit-config.yaml
```diff
@@ -1,6 +1,7 @@
 # Keep tool versions in sync with the versions in requirements-dev.txt
 default_language_version:
     python: python3
+default_stages: [pre-commit]
```
charn, line 4: Don't know about this one, is it needed? If you replace it with `default_install_hook_types: [pre-commit, commit-msg]` the `commit-msg` hook is also installed by default.

tuomas777: `default_install_hook_types: [pre-commit, commit-msg]` is definitely an improvement so that users won't need to install `commit-msg` hook separately, added that 👍 

However, I think `default_stages` is still needed. Without it, many of the hooks are run twice (on pre-commit and commit-msg stages) when committing. That doesn't actually break anything, but it makes pre-commit's output a bit confusing:

```
trim trailing whitespace.................................................Passed
fix end of files.........................................................Passed
check yaml...............................................................Passed
check toml...........................................(no files to check)Skipped
check for added large files..............................................Passed
black....................................................................Passed
flake8...................................................................Passed
isort....................................................................Passed
check yaml...........................................(no files to check)Skipped
check toml...........................................(no files to check)Skipped
black................................................(no files to check)Skipped
flake8...............................................(no files to check)Skipped
isort................................................(no files to check)Skipped
commitlint...............................................................Passed
```

versus 
```
trim trailing whitespace.................................................Passed
fix end of files.........................................................Passed
check yaml...............................................................Passed
check toml...........................................(no files to check)Skipped
check for added large files..............................................Passed
black....................................................................Passed
flake8...................................................................Passed
isort....................................................................Passed
commitlint...............................................................Passed
```

charn: Indeed, didn't really notice that behavior earlier. `default_stages` looks good. :+1:

.github/workflows/ci.yml
```diff
@@ -98,6 +98,9 @@ jobs:
           linters: isort
           run: isort . --check-only --diff
 
+      - name: Check commitlint
+        uses: wagoid/commitlint-github-action@0d749a1a91d4770e983a7b8f83d4a3f0e7e0874e  # v5.4.4
```
charn, line 102: Doesn't wagoid/commitlint-github-action@v5.4.4 work? :thinking:

charn (approved):
LGTM! :+1: :shipit:

PR #798, opened 2023-11-20
Fix major performance issues with include=sub_events

voneiden:
Original implementation fetched sub events without any select/prefetch related, causing thousands of redundant SQL queries.

Second commit cleans up serializer context logic for EventViewSet.

Ref: LINK-1725

### DIFF
--- a/events/api.py
+++ b/events/api.py
@@ -3634,13 +3634,9 @@ class EventViewSet(
             {"headline", "secondary_headline"}
         )
         context["extensions"] = get_extensions_from_request(self.request)
-        (
-            context["keywords"],
-            context["location"],
-            context["image"],
-            context["sub_events"],
-            bulk,
-        ) = self.cache_related_fields(self.request)
+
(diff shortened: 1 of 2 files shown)


nicobav (approved):
LGTM!  🏎️

events/api.py
```diff
@@ -3634,38 +3626,46 @@ def get_serializer_context(self):
...
         context["extensions"] = get_extensions_from_request(self.request)
-        (
-            context["keywords"],
-            context["location"],
-            context["image"],
-            context["sub_events"],
-            bulk,
-        ) = self.cache_related_fields(self.request)
+
+        if self.request.method in ("POST", "PUT", "PATCH"):
```
charn, line 3630: I wonder if `not in permissions.SAFE_METHODS` would be another option for this. That of course would include more than just the mentioned methods, so maybe this is better. :+1:

charn (approved):
LGTM! :+1: :shipit:

PR #802, opened 2023-11-22
perf: improve event listing prefetches (ROUND 2)

voneiden:
Performance on staging was still bad in our test scenario, as the previous optimization reduced the query count by only 1500, leaving another 1400 redundant queries. 

For all events the following were missing:
* helevents_user (missing creators and modifier users of Event and Image)
* munigeo_administrativedivision_translation

Additionally for include=sub_events, the subevent prefetch queries were missing prefetches at least for 

* munigeo_administrativedivisiontype
* munigeo_muncipality
* events_place_divisions
* django_orghierachy_organization

### DIFF
--- a/events/api.py
+++ b/events/api.py
@@ -3576,13 +3576,15 @@ class EventViewSet(
     # sub_event prefetches are handled in get_queryset()
     queryset = (
         Event.objects.all()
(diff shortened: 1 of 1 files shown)


nicobav (approved):
Alrighty 🚤

PR #804, opened 2023-11-24
LINK-1728 | lower LargeResultsSetPagination.max_page_size

charn:
Large values seems to be unused so the page size for LargeResultsSetPagination is limited to the default page size.


nicobav (approved):
LGTM :shipit:

PR #805, opened 2023-11-24
LINK-1741 | Organization admins can manage registrations & signups created by them

harriris-vincit:
### Description
Allows organization admins (also known as event admins or regular admins) to view, edit and delete registrations and signups if the registration is created by them.

Creation of new registrations and signups has already been possible for them, and no changes have been done for that functionality.

### Closes
[LINK-1741](https://helsinkisolutionoffice.atlassian.net/browse/LINK-1741)

[LINK-1741]: https://helsinkisolutionoffice.atlassian.net/browse/LINK-1741?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/registrations/permissions.py
+++ b/registrations/permissions.py
@@ -71,6 +71,7 @@ class CanAccessRegistrationSignups(
         return (
             user.is_superuser
             or user.is_registration_admin_of(obj.publisher)
+            or (user.is_admin_of(obj.publisher) and obj.created_by_id == user.id)
             or user.is_registration_user_access_user_of(obj.registration_user_accesses)
         )
 
@@ -81,8 +82,13 @@ class CanAccessSignup(UserDataFromRequestMixin, permissions.BasePermission):
 
     @staticmethod
     def _has_object_update_permission(request: Request, obj: SignUp) -> bool:
-        if request.user.is_superuser or request.user.is_registration_admin_of(
-            obj.publisher
+        if (
+            request.user.is_superuser
+            or request.user.is_registration_admin_of(obj.publisher)
+            or (
+                request.user.is_admin_of(obj.publisher)
+                and obj.registration.created_by_id == request.user.id
+            )
         ):
             return True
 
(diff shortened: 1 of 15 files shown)


registrations/permissions.py
```diff
@@ -125,8 +130,11 @@ def has_object_permission(
 class CanAccessSignupGroup(CanAccessSignup):
     @staticmethod
     def _has_object_update_permission(request: Request, obj: SignUpGroup) -> bool:
-        if request.user.is_superuser or request.user.is_registration_admin_of(
-            obj.publisher
+        if (
+            request.user.is_superuser
+            or request.user.is_registration_admin_of(obj.publisher)
+            or request.user.is_admin_of(obj.publisher)
+            and obj.registration.created_by_id == request.user.id
```
jorilindell, line 137: I would use parentheses here as well:
```
if (
            request.user.is_superuser
            or request.user.is_registration_admin_of(obj.publisher)
            or (
                request.user.is_admin_of(obj.publisher)
                and obj.registration.created_by_id == request.user.id
            )
        )
```

jorilindell (commented):
Looks really good. I added couple of comments about parentheses. for me they make the code little bit more readable. But it's matter of taste.

You could add a test to `test_send_message.py` file and test that admin user who create the registration can send a message to a signup's or signup group's contact person.

registrations/permissions.py
charn: Based on https://docs.python.org/3/reference/expressions.html#operator-precedence parentheses shouldn't be required, but I guess they improve readability so might be a nice addition.

harriris-vincit: Yes, I think it's a nice readability improvement. Changed as suggested now.

jorilindell (approved):
Looks good to me

PR #807, opened 2023-11-27
LINK-1727 | Fix internal server error on image upload when external user sets an arbitrary publisher

charn:
This PR fixes an internal server error being thrown instead of returning a validation error when an external user (a user which doesn't belong to an organization) sets an arbitrary organization. Since the user doesn't belong to any organization the publisher field should be left blank.


danipran (approved):
LGTM :+1:

PR #808, opened 2023-11-30
LINK-1718 | Add email notification to contact persons on event cancellation

harriris-vincit:
### Description
When an event is cancelled (`Event.event_status` is set to `EventCancelled`), an email notification about the cancellation will be sent to the contact persons of all the related signups and signup groups.

Due to the possibility of large amounts of emails being sent in the worst case, the `django-mailer` package is added and taken into use to queue the emails for later sending (which can be easily further controlled in terms of sending delay and batch size, for example). Using `django-mailer` also allows non-blocking emailing so this should bring a general improvement too for code that uses the Django emailing functions.

Scheduling for the `django-mailer` management commands `send_mail` (send queued emails), `retry_deferred` (retry failed emails) and `purge_mail_log` (clean successfully sent and/or failed emails from DB) has it's own PR in DevOps: https://dev.azure.com/City-of-Helsinki/linkedevents/_git/linkedevents-pipelines/pullrequest/7082

### Closes
[LINK-1718](https://helsinkisolutionoffice.atlassian.net/browse/LINK-1718)

[LINK-1718]: https://helsinkisolutionoffice.atlassian.net/browse/LINK-1718?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,9 +224,9 @@ Note that search tests will fail unless you configure [search](#search)
 
 
 ## Sending email
-The project uses [django-mailer](https://github.com/pinax/django-mailer) for queuing and sending email. When a normal Django email function like `send_mail` is called, the email will be stored into an email queue that exists in the database for later sending.
+The project uses [django-mailer](https://github.com/pinax/django-mailer) for queuing and sending email. When a normal Django email function like `send_mail` is called, the email will be stored into a queue that exists in the database for later sending.
 
-`django-mailer` comes with a number of management commands for interacting with the email queue. Out of those, the `send_mail` command is for us perhaps the most notable one as it is responsible for the actual sending of the queues messages.
+`django-mailer` comes with a number of management commands for interacting with the email queue. Out of those, the `send_mail` command is for us perhaps the most notable one as it is responsible for the actual sending of the queued messages.
 
 To send messages locally, you can run the `send_mail` command:
 `./manage.py send_mail`
--- a/README.md
+++ b/README.md
@@ -223,6 +223,17 @@ py.test events
 Note that search tests will fail unless you configure [search](#search)
 
 
+## Sending email
+The project uses [django-mailer](https://github.com/pinax/django-mailer) for queuing and sending email. When a normal Django email function like `send_mail` is called, the email will be stored into an email queue that exists in the database for later sending.
+
+`django-mailer` comes with a number of management commands for interacting with the email queue. Out of those, the `send_mail` command is for us perhaps the most notable one as it is responsible for the actual sending of the queues messages.
+
+To send messages locally, you can run the `send_mail` command:
+`./manage.py send_mail`
+
+For more information about `django-mailer` and the management commands, you can refer to the [usage documentation](https://github.com/pinax/django-mailer/blob/master/docs/usage.rst).
+
+
 ## Requirements
 
 Linked Events uses two files for requirements. The workflow is as follows.
--- a/events/signals.py
+++ b/events/signals.py
@@ -1,16 +1,7 @@
 import logging
 
-from django.contrib.auth import get_user_model
-from django.core.mail import send_mail
 from django.db.models.signals import post_save
 from django.dispatch import receiver
-
-from notifications.models import (
-    NotificationTemplateException,
-    NotificationType,
-    render_notification_template,
-)
-from registrations.utils import get_email_noreply_address
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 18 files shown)


registrations/signals.py
```diff
@@ -17,6 +29,22 @@ def _signup_or_group_post_delete(instance: Union[SignUp, SignUpGroup]) -> None:
...
 
 
+def _send_event_cancellation_notification(event):
+    registration_ids = Registration.objects.filter(event_id=event.pk).values_list(
+        "pk", flat=True
+    )
+
+    for contact_person in SignUpContactPerson.objects.filter(
+        Q(email__isnull=False)
+        & ~Q(email__exact="")
```
tuomas777, line 39:
```suggestion
        & ~Q(email="")
```
No need for `__exact` because it is used when no lookup type is specified.

registrations/signals.py
```diff
@@ -36,3 +64,31 @@ def signup_group_post_delete(
...
+@receiver(
+    post_save,
+    sender=Event,
+    dispatch_uid="notify_signups_on_event_cancelled_post_save",
+)
+def notify_signups_on_event_cancelled_post_save(
+    sender: type[Event], instance: Event, **kwargs: dict
+) -> None:
+    if getattr(instance, "_send_event_cancellation_notification", False):
+        transaction.on_commit(lambda: _send_event_cancellation_notification(instance))
```
tuomas777, line 94: Now that we have django-mailer, I think we want to bind notification sending (= actually adding notifications to be sent) exactly to event saving, and therefore `Event.save()` should be wrapped inside a transaction. And when that is the case, I think we can call `_send_event_cancellation_notification()` in the `pre_save` handler, because the actual sending will be then rolled back if the event saving fails. And then there is no need for this `post_save` handler at all. Maybe. 🤔

registrations/models.py
```diff
@@ -697,33 +697,40 @@ def get_registration_message(self, subject, cleaned_body, plain_text_body):
...
 
-    def send_notification(self, notification_type):
+    def get_notification_message(self, notification_type):
         [_, linked_registrations_ui_locale] = get_ui_locales(self.service_language)
 
+        if notification_type == SignUpNotificationType.EVENT_CANCELLATION:
+            email_template = "event_cancellation_confirmation.html"
+        elif notification_type == SignUpNotificationType.CANCELLATION:
+            email_template = "cancellation_confirmation.html"
+        else:
```
tuomas777, line 707: IMO in cases like this, it's better to be defensive and check explicitly all valid cases, and then raise an exception in the `else` block for possible invalid types, so calling the function incorrectly won't be silently ignored.

registrations/models.py
```diff
@@ -697,33 +697,40 @@ def get_registration_message(self, subject, cleaned_body, plain_text_body):
...
         try:
-            send_mail(
-                get_signup_notification_subject(self, notification_type),
-                rendered_body,
-                get_email_noreply_address(),
-                [self.email],
-                html_message=rendered_body,
-            )
+            send_mail(*message, html_message=rendered_body)
         except SMTPException:
```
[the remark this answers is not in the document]
harriris-vincit: They cannot at this point when the mail is being saved into the DB queue - only when the mail is actually being sent. I'll change these accordingly a bit later today.

events/signals.py
```diff
@@ -47,26 +46,26 @@ def user_created_notification(sender, instance, created, **kwargs):
...
-        except SMTPException as e:
-            logger.error(e, exc_info=True, extra={"user": instance.username})
+
+        send_mail(
+            rendered_notification["subject"],
+            rendered_notification["body"],
+            get_email_noreply_address(),
+            recipient_list,
+            html_message=rendered_notification["html_body"],
+        )
```
tuomas777, line 71: This `user_created_notification` is disabled, and should be removed altogether now (I got a confirmation on that). But I guess that can be done separately from this PR.

tuomas777 (approved):
I guess the README could mention those management commands that need to be called now to get emails sent. Or maybe at least a link to django-mailer's documentation?

Looks good to me, good job! 👍 📫

jorilindell (approved):

events/signals.py
harriris-vincit: I'll actually remove it in this PR, but in a separate commit.

README.md
```diff
@@ -223,6 +223,17 @@ py.test events
 Note that search tests will fail unless you configure [search](#search)
 
 
+## Sending email
+The project uses [django-mailer](https://github.com/pinax/django-mailer) for queuing and sending email. When a normal Django email function like `send_mail` is called, the email will be stored into an email queue that exists in the database for later sending.
+
+`django-mailer` comes with a number of management commands for interacting with the email queue. Out of those, the `send_mail` command is for us perhaps the most notable one as it is responsible for the actual sending of the queues messages.
```
harriris-vincit, line 229:
```suggestion
`django-mailer` comes with a number of management commands for interacting with the email queue. Out of those, the `send_mail` command is for us perhaps the most notable one as it is responsible for the actual sending of the queued messages.
```

jorilindell (approved):

PR #811, opened 2023-12-05
fix: Remove phone_number from mandatory fields

jorilindell:
## Description
Phone_number is removed from the SignUp model and moved to ContactPerson model. So having phone_number in the mandatory fields is misleading and can cause some validation errors for the old registrations where the phone_number is still set to be mandatory.

Remove the phone_number field from the mandatory fields of existing registration and don't allow phone_number as a mandatory field when updating or creating a registration.


harriris-vincit (approved):
Looks great!

PR #816, opened 2023-12-12
feat: add price group models and price group admin for registrations

harriris-vincit:
### Description
Adds new models for managing and storing price groups (will be "Asiakasryhmät" in the UI) for registrations. These will be used later when taking Talpa payments for registrations.

`PriceGroup` instances are pricing categories without prices (basically they contain information only about the organization and the name of the price group / pricing category). Prices are given when creating a `Registration` and selecting which price groups are selectable on the registration form - the model for these is `RegistrationPriceGroup`.

`PriceGroup` instances with `publisher=None` are created by the system in a data migration and are considered default price groups available to all publishers / organizations.

`RegistrationPriceGroup` instances are meant to be selectable by the end-users when signing up to a registration. When a selection is made and the registration form is saved, a `SignUpPriceGroup` is created - that one keeps the historic information about `PriceGroup.description` and `RegistrationPriceGroup.price` that existed at the moment the signup was made.

Note: most of the described business logic will be added in later PRs. This PR adds only the new models and the possibility to manage `PriceGroups` and select them for a `Registration` through the Django admin site.
### Closes
[LINK-1214](https://helsinkisolutionoffice.atlassian.net/browse/LINK-1214)
[LINK-1684](https://helsinkisolutionoffice.atlassian.net/browse/LINK-1684)
[LINK-1834](https://helsinkisolutionoffice.atlassian.net/browse/LINK-1834)

[LINK-1214]: https://helsinkisolutionoffice.atlassian.net/browse/LINK-1214?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LINK-1684]: https://helsinkisolutionoffice.atlassian.net/browse/LINK-1684?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


voneiden (approved):
:ok:

registrations/admin.py
```diff
@@ -87,3 +100,32 @@ def get_readonly_fields(self, request, obj=None):
...
+        "description",
+    )
+    list_filter = (PublisherFilter,)
+    search_fields = ("description",)
+
+    def save_model(self, request, obj, form, change):
+        if obj.pk is None:
+            obj.created_by = request.user
+        else:
+            obj.last_modified_by = request.user
```
jorilindell, line 123: /events/admin.py file has BaseAdmin class. Could this be inherited from there?

harriris-vincit: `BaseAdmin` uses a `modified_by` field, but in registration models we use `last_modified_by` instead. Because of this it might not be worth it to inherit.

jorilindell: Ah, very true. The naming difference is a little bit confusing. But there is no reason to do anything to it right now.

jorilindell: There is a similar save_model method in RegistrationAdmin as well in you like to remove some duplicate code.

voneiden (approved):
Checked the force-pushed changes... seems good. The PR has a good description, I wouldn't mind having it included in a commit message body. There's of course going to be a reference to this PR in the merge commit, but yeah, it's one step further away.

Especially this wasn't immediately obvious to me from the code: 

> PriceGroup instances with publisher=None are created by the system in a data migration and are considered default price groups available to all publishers / organizations.

registrations/admin.py
```diff
@@ -87,3 +100,73 @@ def get_readonly_fields(self, request, obj=None):
...
+            yield {
+                "selected": self.lookup_val == lookup,
+                "query_string": changelist.get_query_string(
+                    {self.lookup_kwarg: lookup}
+                ),
+                "display": title,
+            }
+
+
+class PriceGroupAdmin(VersionAdmin):
```
jorilindell, line 126: You could add `TranslationAdmin` class (you can import it from `modeltranslation.admin`) to display language specific fields for the description in Django admin as well. 

After adding `TranslationAdmin` you have to move "modeltranslation" before "helusers.apps.HelusersAdminConfig" in INSTALLED_APPS to avoid `The model "Registration" is not registered for translation.` errors.

`TranslationAdmin` should be added to RegistrationAdmin class as well. But that can be done in a separate PR.

jorilindell (changes requested):
Description field should be translated field in Django admin as well. Otherwise looks very good

registrations/admin.py
harriris-vincit: I created a new base class in `registrations/admin.py`

registrations/admin.py
harriris-vincit: A very important point - I've now implemented these changes (also for `RegistrationAdmin`)

jorilindell (approved):
Looks very good!

PR #823, opened 2023-12-28
LINK-1685 | Add registration price groups to API

harriris-vincit:
### Description
Adds three new registration price group related features to the API:

- The possibility to create, read, update and delete `PriceGroup` instances through a new endpoint `price_group`. Filtering by the `publisher` and `description` fields and ordering by the `description` field are also supported when listing the instances. Create, update and delete operations are restricted to superusers and financial admins only, but reading is allowed to all authenticated users.
- The possibility to create, read, update and delete `RegistrationPriceGroup` instances for `Registrations` through the `registration` endpoint (so essentially as part of the `Registration` payload). This is allowed to all users who are allowed to do the same operations for `Registration` instances.
- The possibility to create, read, update and delete `SignUpPriceGroup` instances for `SignUps` through the `signup` and `signup_group` endpoints (so essentially as part of the `SignUp` payload). This is allowed to all users who are allowed to do the same operations for `SignUp` instances.

The features are all behind a feature flag called `WEB_STORE_INTEGRATION_ENABLED`.

**Note:** Due to the large amount of changes it might be best to do the review on a commit basis (commit-by-commit) as the three new features are each contained in a commit of their own.
### Closes
[LINK-1685](https://helsinkisolutionoffice.atlassian.net/browse/LINK-1685)
[LINK-1857](https://helsinkisolutionoffice.atlassian.net/browse/LINK-1857)

[LINK-1685]: https://helsinkisolutionoffice.atlassian.net/browse/LINK-1685?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


registrations/filters.py
```diff
@@ -0,0 +1,193 @@
...
+        if value == ["None"] or value == ["none"]:
+            return queryset.filter(publisher=None)
+
+        publishers = Organization.objects.filter(pk__in=value).values_list(
+            "pk", flat=True
+        )
+        if not publishers:
+            return self.Meta.model.objects.none()
+
+        return queryset.filter(publisher_id__in=publishers)
```
jorilindell, line 181: It would be nice to filter default and publisher specific price group with a single request e.g. `/price_group/?publisher=ahjo:00001,none`

At the moment `/price_group/?publisher=ahjo:00001,ahjo:u021600` and `/price_group/?publisher=none` works as expected but price groups without publisher are excluded when trying to use `/price_group/?publisher=ahjo:00001,none` filter

registrations/serializers.py
```diff
@@ -975,3 +1130,25 @@ class RegistrationSignupsExportSerializer(serializers.Serializer):
         choices=["en", "sv", "fi"],
         default="fi",
     )
+
+
+class PriceGroupSerializer(CreatedModifiedBaseSerializer):
```
jorilindell, line 1135: At the moment only default language value for the description is returned e.g.
```
{
  "id": 9,
  "created_time": "2024-01-04T10:55:35.243078Z",
  "last_modified_time": "2024-01-04T10:55:35.243164Z",
  "created_by": "Admin User - jori.lindell@gmail.com",
  "last_modified_by": null,
  "publisher": "ahjo:00001",
  "is_free": false,
  "description": "Helsingin kaupungin hintaryhmä"
},
```
but description should be returned similar way as in other responses e.g.
```
{
  "id": 9,
  "created_time": "2024-01-04T10:55:35.243078Z",
  "last_modified_time": "2024-01-04T10:55:35.243164Z",
  "created_by": "Admin User - jori.lindell@gmail.com",
  "last_modified_by": null,
  "publisher": "ahjo:00001",
  "is_free": false,
  "description": {
    "fi": "Helsingin kaupungin hintaryhmä"
  }
},
```
This could be achieved by inheriting the behaviour from TranslatedModelSerializer class e.g.
```
class PriceGroupSerializer(TranslatedModelSerializer, CreatedModifiedBaseSerializer):
```
Issue with that is that `from events.api import TranslatedModelSerializer` will cause circular import problem. So maybe some refactoring is needed as well

registrations/migrations/0033_registration_price_group_vat.py
```diff
@@ -0,0 +1,85 @@
...
+                decimal_places=4, default=Decimal("0"), max_digits=19
+            ),
+        ),
+        migrations.AddField(
+            model_name="registrationpricegroup",
+            name="vat",
+            field=models.DecimalField(
+                decimal_places=4, default=Decimal("0"), max_digits=19
+            ),
+        ),
```
jorilindell, line 26: This is not super critical since price groups are not in the production, but in theory these values would be nice to migrate for the existing price groups. At the moment the default value is used for them

registrations/forms.py
```diff
@@ -0,0 +1,31 @@
...
+class RegistrationPriceGroupAdminForm(forms.ModelForm):
+    price = forms.DecimalField(
+        required=False, max_digits=19, decimal_places=2, min_value=0
+    )
+    price_without_vat = forms.DecimalField(
+        required=False, disabled=True, max_digits=19, decimal_places=2, min_value=0
+    )
+    vat = forms.DecimalField(
+        required=False, disabled=True, max_digits=19, decimal_places=2, min_value=0
+    )
```
jorilindell, line 15: The form is initialized with 4 decimals and limiting decimal amount to 2 causes validation problems. Basically user is not able to edit price groups because `price_without_vat` and `vat` fields are disabled

Screenshot 2024-01-04 at 14 43 15

registrations/serializers.py
```diff
@@ -339,6 +421,63 @@ class Meta:
...
+    )
+    vat_percentage = serializers.DecimalField(
+        required=False,
+        max_digits=4,
+        decimal_places=2,
+        min_value=0,
+        max_value=Decimal("99.99"),
+    )
+
+    def validate(self, data):
```
jorilindell, line 446: API could also check that unique constraint "unique_registration_price_group" is not violated. At the moment 500 error is returned if user tries to save 2 registration price groups with same price group id. e.g.
```
[
    {
        "id": 6,
        "price": "12.00",
        "price_group": 2,
        "vat_percentage": "14.00"
    },
    {
        "price": 12,
        "price_group": 2,
        "vat_percentage": "0.00"
    }
]
```

registrations/serializers.py
```diff
@@ -339,6 +421,63 @@ class Meta:
...
+
+        errors = {}
+
+        vat_percentage = validated_data.get("vat_percentage")
+        if vat_percentage not in [
+            vat[0] for vat in RegistrationPriceGroup.VAT_PERCENTAGES
+        ]:
+            errors["vat_percentage"] = _("{value} is not a valid choice.") % {
+                "value": vat_percentage
+            }
```
jorilindell, line 457: This returns `{value} is not a valid choice.` as validation message

Should this be 
```
errors["vat_percentage"] = _("%(value)s is not a valid choice.") % {
                "value": vat_percentage
            }
```

registrations/forms.py
```diff
@@ -0,0 +1,31 @@
...
+
+    class Meta:
+        model = RegistrationPriceGroup
+        fields = (
+            "price_group",
+            "price",
+            "vat_percentage",
+            "price_without_vat",
+            "vat",
+        )
```
jorilindell, line 31: Would it be possible to return PriceGroup `description` for the RegistrationPriceGroup as well?

At the moment price group description is only available via signup_group and requires extra requests to show in the UI. It makes the UI implementation a little bit more complicated. Also price_group endpoint requires authentication so price group description is not possible to show for unauthenticated users

registrations/filters.py
harriris-vincit: Filtering by default and specific publishers at the same time hows now been implemented.

registrations/forms.py
harriris-vincit: Should be fixed now by allowing only two decimal places for all price fields in the database (as the Euro currency uses only two decimal places)

registrations/serializers.py
harriris-vincit: Should be fixed now

registrations/serializers.py
harriris-vincit: Added this kind of a validation to `RegistrationSerializer`.

registrations/serializers.py
harriris-vincit: Fixed

registrations/migrations/0033_registration_price_group_vat.py
harriris-vincit: Added a data migration for VAT prices to this migration file

registrations/serializers.py
```diff
@@ -975,3 +977,45 @@ class RegistrationSignupsExportSerializer(serializers.Serializer):
...
+
+
+class PriceGroupSerializer(TranslatedModelSerializer, CreatedModifiedBaseSerializer):
+    def _description_is_valid(self, data, validated_data):
+        description_fields = ["description_fi", "description_en", "description_sv"]
+
+        return (
+            self.partial
+            and not any([field in data.keys() for field in description_fields])
+        ) or any([validated_data.get(field) for field in description_fields])
```
tuomas777, line 1160: I think this part could use a comment. With data and validated_data mixed in with different conditions, it is surprisingly difficult to figure out what we are exactly checking here 🙂

registrations/tests/test_price_group_delete.py
```diff
@@ -0,0 +1,267 @@
...
+
+    return response
+
+
+def assert_delete_default_price_group(api_client: APIClient, price_group: PriceGroup):
+    assert PriceGroup.objects.count() == 9
+
+    assert_delete_price_group(api_client, price_group.pk)
+
+    assert PriceGroup.objects.count() == 8
```
tuomas777, line 42: Hmm, not sure what actually ties this function to default price groups, and thus why this is even needed. Couldn't the count checks be added to `assert_delete_price_group()` and only that function used? 🙂

registrations/permissions.py
```diff
@@ -170,3 +172,31 @@ def get_signups_keys():
...
+                user_organization,
+            ) = self.user_data_source_and_organization_from_request(request)
+            return request.user.is_financial_admin_of(user_organization)
+
+        return True
+
+    def has_object_permission(
+        self, request: Request, view: APIView, obj: PriceGroup
+    ) -> bool:
+        if view.action == "retrieve":
```
tuomas777, line 197: When checking whether the request is a read or a write operation in custom permissions, [the preferred way in DRF is checking against `permissions.SAFE_METHODS`](https://www.django-rest-framework.org/api-guide/permissions/#custom-permissions) .

registrations/models.py
```diff
@@ -170,6 +171,43 @@ class PriceGroup(CreatedModifiedBaseModel):
...
 
+    @cached_property
+    def old_instance(self):
+        if not self.pk:
+            return None
+
+        return PriceGroup.objects.get(pk=self.pk)
+
+    @cached_property
+    def publisher_change_allowed(self):
```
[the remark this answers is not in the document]
harriris-vincit: Changed to use just `property`

registrations/models.py
```diff
@@ -170,6 +171,43 @@ class PriceGroup(CreatedModifiedBaseModel):
...
+        if self.old_instance and not self.old_instance.publisher_id:
+            raise DRFPermissionDenied(_("You may not delete a default price group."))
+
+        super().delete(*args, **kwargs)
+
+    def save(self, *args, **kwargs):
+        if self.old_instance and not self.old_instance.publisher_id:
+            raise DRFPermissionDenied(_("You may not edit a default price group."))
+
+        if not self.publisher_change_allowed:
```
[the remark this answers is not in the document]
harriris-vincit: I renamed this to `publisher_is_valid` as it is more descriptive of what the method actually does

registrations/models.py
```diff
@@ -170,6 +171,43 @@ class PriceGroup(CreatedModifiedBaseModel):
...
+            return True
+
+        return (
+            self.old_instance.publisher_id == self.publisher_id
+            or not self.registration_price_groups.exists()
+        )
+
+    def delete(self, *args, **kwargs):
+        if self.old_instance and not self.old_instance.publisher_id:
+            raise DRFPermissionDenied(_("You may not delete a default price group."))
```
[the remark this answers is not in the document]
harriris-vincit: Changed this  to use a custom exception that is based on `ValidationError`

registrations/serializers.py
harriris-vincit: Refactored the code to be more easily understandable AND added a comment too

registrations/tests/test_price_group_delete.py
harriris-vincit: Improved this and similar functions in the other price group test files

registrations/permissions.py
harriris-vincit: Changed as suggested

registrations/serializers.py
```diff
@@ -339,6 +425,77 @@ class Meta:
...
+    price = serializers.DecimalField(
+        required=False, max_digits=19, decimal_places=2, min_value=0
+    )
+    price_without_vat = serializers.DecimalField(
+        required=False, read_only=True, max_digits=19, decimal_places=2
+    )
+    vat = serializers.DecimalField(
+        required=False, read_only=True, max_digits=19, decimal_places=2
+    )
+    vat_percentage = serializers.DecimalField(
```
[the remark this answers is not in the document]
harriris-vincit: That's what I though too, but because `ChoiceField` normalizes the value to a string and the API user can provide values like "10.00" or "10.0", the out-of-the-box validation with a `ChoiceField` will fail in those cases and extra code is required to make it work. Thus, I thought this would after all be the simplest solution that I could come up with.

tuomas777 (approved):
Quite a massive addition 😅 Didn't properly test it manually myself, but cannot find anything else to complain about the code, so LGTM 👍 Great job! 🥇

PR #825, opened 2023-12-29
LINK-1698 | Additions to gdpr api user data fetch

nicobav:
Adds to gdpr-api/ GET data:
 - Organizations where user is admin or regular user
 - Events created by user 
 
 Refs LINK-1698

### DIFF
--- a/events/models.py
+++ b/events/models.py
@@ -34,7 +34,9 @@ from django.db.models.signals import m2m_changed
 from django.dispatch import receiver
 from django.utils import timezone
 from django.utils.translation import gettext_lazy as _
+from helsinki_gdpr.models import SerializableMixin
 from image_cropping import ImageRatioField
+from mptt.managers import TreeManager
 from mptt.models import MPTTModel, TreeForeignKey
 from mptt.querysets import TreeQuerySet
 from munigeo.models import AdministrativeDivision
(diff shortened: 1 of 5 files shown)


helevents/models.py
```diff
@@ -228,11 +229,31 @@ class User(AbstractUser, UserModelPermissionMixin, SerializableMixin):
...
+        publishers = Organization.objects.filter(
+            Q(admin_users=self) | Q(regular_users=self)
+        )
+
+        return [
+            org.serialize()
+            for org in SerializablePublisher.objects.filter(
+                id__in=publishers.values_list("id", flat=True)
+            )
+        ]
```
danipran, line 255: Would this work as well? 🤔 
```suggestion
        return [
            org.serialize()
            for org in SerializablePublisher.objects.filter(
                Q(admin_users=self) | Q(regular_users=self)
            )
        ]
```

danipran (approved):
LGTM! Left one suggestion.

helevents/models.py
nicobav: yes it would 🙃  
There might have been readability / self documenting reasoning but I don't see it anymore 😄

danipran (approved):
LGTM, left one comment which might need addressing

events/models.py
```diff
@@ -783,9 +838,50 @@ class Meta:
...
+        {"name": "short_description"},
+        {"name": "start_time"},
+        {"name": "end_time"},
+        {"name": "images"},
+        {"name": "keywords"},
+        {
+            "name": "publisher",
+            "accessor": lambda x: (f"{x.id} - {x.name}") if x else "",
+        },
+        {"name": "user_organization"},
```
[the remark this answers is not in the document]
nicobav: nope! Nice catch 🎣 
I'll remove that 👍

PR #826, opened 2024-01-03
chore: support multiple OIDC providers

voneiden:
Additionally, changes the test token scope structure to match keycloak.


linkedevents/settings.py
```diff
@@ -135,11 +135,11 @@ def sentry_anonymize_user_repr(obj, hint):
...
     SUPPORT_EMAIL=(str, ""),
     SYSTEM_DATA_SOURCE_ID=(str, "system"),
-    TOKEN_AUTH_ACCEPTED_AUDIENCE=(str, "https://api.hel.fi/auth/linkedevents"),
+    TOKEN_AUTH_ACCEPTED_AUDIENCE=(list, "https://api.hel.fi/auth/linkedevents"),
     TOKEN_AUTH_ACCEPTED_SCOPE_PREFIX=(str, "linkedevents"),
-    TOKEN_AUTH_AUTHSERVER_URL=(str, "https://api.hel.fi/sso/openid"),
-    TOKEN_AUTH_FIELD_FOR_CONSENTS=(str, "https://api.hel.fi/auth"),
-    TOKEN_AUTH_REQUIRE_SCOPE_PREFIX=(bool, True),
+    TOKEN_AUTH_AUTHSERVER_URL=(list, "https://api.hel.fi/sso/openid"),
+    TOKEN_AUTH_FIELD_FOR_CONSENTS=(list, "https://api.hel.fi/auth"),
```
danipran, line 141: Default values should be lists... I think? 🤔 not sure if environ automatically casts strings into lists or not.

danipran (changes requested):

linkedevents/settings.py
voneiden: true true

PR #829, opened 2024-01-04
LINK-1841: remove daily_tasks.sh, hourly_tasks.sh

danipran:
No longer needed, moved to Azure Pipelines.

Refs: LINK-1841


tuomas777 (approved):
👍

PR #830, opened 2024-01-05
LINK-1644 feat: add event user data removal to gdpr delete

nicobav:
In gdpr api delete operation, null  users's event user fields when the `user_email` field matches to user's email.

Refs LINK-1644


danipran (changes requested):
Looks fine, but `helevents/utils.py` seems to have Windows EOLs (CR LF) instead of Linux EOLs (LF)... which explains why Git shows that the whole file was edited.

danipran (approved):
LGTM! 👍

PR #833, opened 2024-01-09
feat: remove spammy debug logs from importers

danipran:
E.g. on an hourly cronjob, there were 14.7k lines of logs generated. These two logging lines generated approximately 13.7k lines. Which is about 93% of the total lines.

These aren't very informative either, so decided to remove them. It's a lot of noise.


tuomas777 (approved):
Yes please 👍

PR #834, opened 2024-01-10
LINK-1382 | feat: support keycloak in User.token_amr_claim

charn:
Tunnistamo returns amr-claim as a string. Keycloak return the claim as a list of strings, which is according to specification. This changes the logic in linked events to expect a list for User.token_amr_claim.


danipran (approved):
LGTM! :+1:

PR #835, opened 2024-01-11
fix: disable file-based lock for django-mailer

harriris-vincit:
### Description
The `django-mailer` uses a lockfile to ensure that an email is not sent twice. By default, the path of that lockfile is the current working directory which seems to cause a permission issue with LE cron jobs that try to create the lockfile. Since `django-mailer` also has a DB-based lock that is more reliable than the file-based one when available, this PR disables the file-based lock completely.
### Closes
[LINK-1852](https://helsinkisolutionoffice.atlassian.net/browse/LINK-1852)

[LINK-1852]: https://helsinkisolutionoffice.atlassian.net/browse/LINK-1852?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/linkedevents/settings.py
+++ b/linkedevents/settings.py
@@ -577,6 +577,7 @@ DEFAULT_FROM_EMAIL = env(
     "DEFAULT_FROM_EMAIL"
 )  # Email address used as default "from" email
 EMAIL_BACKEND = "mailer.backend.DbBackend"
+MAILER_USE_FILE_LOCK = False
 MAILER_EMAIL_MAX_RETRIES = 5
 if EMAIL_HOST and EMAIL_PORT:
     MAILER_EMAIL_BACKEND = "django.core.mail.backends.smtp.EmailBackend"


nicobav:
I haven't used mailer but what I found from the [docs](https://github.com/pinax/django-mailer/blob/master/docs/usage.rst#locking):

> The send_all command uses a filesystem-based lock file in case clearing the queue takes longer than the interval between calling send_all(). This works to stop multiple workers on a single machine from processing the messages multiple times.
>
>To stop workers processes on different machines from sending the same mail multiple times, it also uses database-level locking where possible. Where available this is more reliable than filesystem-based locks.
> ....
>
> If you want to disable the file-based locking, you can set the MAILER_USE_FILE_LOCK setting to False.

I'm thinking that should we just disable the file based lock? `MAILER_USE_FILE_LOCK = True`

harriris-vincit:
> I haven't used mailer but what I found from the [docs](https://github.com/pinax/django-mailer/blob/master/docs/usage.rst#locking):
> 
> > The send_all command uses a filesystem-based lock file in case clearing the queue takes longer than the interval between calling send_all(). This works to stop multiple workers on a single machine from processing the messages multiple times.
> > To stop workers processes on different machines from sending the same mail multiple times, it also uses database-level locking where possible. Where available this is more reliable than filesystem-based locks.
> > ....
> > If you want to disable the file-based locking, you can set the MAILER_USE_FILE_LOCK setting to False.
> 
> I'm thinking that should we just disable the file based lock? `MAILER_USE_FILE_LOCK = True`

This is a good catch. I think the file based lock can also lead to problems in case a process crashes, etc, so the DB-based lock would likely be better that way too.

tuomas777 (approved):
👍

PR #838, opened 2024-01-15
LINK-1045 | Creating an Order in Talpa

harriris-vincit:
### Description
Adds the following new features:
- API clients and settings for the Talpa web store API (currently only the Order Experience API is supported).
- A new model `SignUpPayment` that has the payment information for signups and tracks the status of the Talpa order (status updates will be done through webhook notifications in another PR).
- Two new serializers for creating an order for a `SignUp` or a `SignUpGroup`. When giving the boolean `create_payment` in the payload of a signup or a group, the order API request will be done provided that signups have price groups selected and that the validation otherwise passes. When the API request is done successfully, a `SignUpPayment` is created.

### Closes
[LINK-1045](https://helsinkisolutionoffice.atlassian.net/browse/LINK-1045)

[LINK-1045]: https://helsinkisolutionoffice.atlassian.net/browse/LINK-1045?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jorilindell:
How is the precision of prices defined in the Tapla. Some of the prices are now using 4 decimals and total price doesn't have decimals at all.
Screenshot 2024-01-24 at 9 10 35

Maybe we could use 2 decimals for all the prices?. e.g. like this:
Screenshot 2024-01-24 at 9 13 54

harriris-vincit:
> How is the precision of prices defined in the Tapla. Some of the prices are now using 4 decimals and total price doesn't have decimals at all. Screenshot 2024-01-24 at 9 10 35
> 
> Maybe we could use 2 decimals for all the prices?. e.g. like this: Screenshot 2024-01-24 at 9 13 54

You seem to have the older migrations for price groups still applied - that is likely causing the four decimal places to be displayed for the item rows. The total sum seems to be sent with decimal places from Linked Events & Registrations, but appears without them in Talpa so it's a formatting that Talpa applies to the value. There doesn't appear to be an obvious way in the item row data to change how many decimal places are shown in Talpa, but I'll experiment a bit with the meta data and some other item attributes.

locale/fi/LC_MESSAGES/django.po
```diff
@@ -745,150 +745,203 @@ msgstr ""
...
 msgstr ""
 
-#: registrations/models.py:210
+#: registrations/models.py:225
 msgid ""
 "You may not change the publisher of a price group that has been used in "
 "registrations."
-msgstr "Ilmoittautumisissa käytetyn alennusryhmän julkaisija-kenttää ei voi vaihtaa."
+msgstr ""
+"Ilmoittautumisissa käytetyn alennusryhmän julkaisija-kenttää ei voi vaihtaa."
```
jorilindell, line 799: Should this be
"Ilmoittautumisissa käytetyn asiakasryhmän julkaisija-kenttää ei voi vaihtaa."?

locale/sv/LC_MESSAGES/django.po
```diff
@@ -743,151 +743,204 @@ msgstr ""
...
 msgstr ""
 
-#: registrations/models.py:210
+#: registrations/models.py:225
 msgid ""
 "You may not change the publisher of a price group that has been used in "
 "registrations."
-msgstr "Du får inte ändra utgivare för en prisgrupp som har använts vid "
+msgstr ""
+"Du får inte byta utgivare för en prisgrupp som har använts vid "
```
jorilindell, line 797: Should this be "Du får inte byta utgivare för en kundgrupp som har använts vid "
"registreringar."

jorilindell (approved):
Really good. Added couple of comments about price group Finnish and Swedish translations. You can ignore if they are not relevant.

locale/fi/LC_MESSAGES/django.po
harriris-vincit: It should indeed. I will fix this in the upcoming push

locale/sv/LC_MESSAGES/django.po
harriris-vincit: This should also use the customer group term -> will fix soon

nicobav (commented):
Looks great, I had some minor things in mind but was too slow 😆 🐌

PR #839, opened 2024-01-16
fix: Use event start time for signup birth of date validation

jorilindell:


registrations/tests/test_signup_group_post.py
```diff
@@ -761,20 +761,53 @@ def test_signup_group_date_of_birth_is_mandatory_if_audience_min_or_max_age_spec
...
+            status.HTTP_400_BAD_REQUEST,
+            "The participant is too old.",
+            True,
+        ),
+        ("2000-02-29", status.HTTP_201_CREATED, None, False),
     ],
 )
 @freeze_time("2023-03-14 03:30:00+02:00")
 @pytest.mark.django_db
 def test_signup_group_age_has_to_match_the_audience_min_max_age(
```
harriris-vincit, line 795: There is a similar test for the `signup` endpoint (`test_signup_age_has_to_match_the_audience_min_max_age`) - that one should be updated as well

jorilindell: Updated also tests for the `signup` endpoint

harriris-vincit (approved):
LGTM 👍

PR #840, opened 2024-01-18
LINK-1849 | Add Substitute User Permission

harriris-vincit:
### Description
Makes it possible to add substitute users for registrations to allow administration of the registrations while the user who created the registrations is on vacation, etc. Substitute users have the same permissions as a registration admin or the creator of a registration: they can fully view, edit or delete the registration's or related signups' information. Only users with a "hel.fi" email address can be marked as substitutes.
### Closes
[LINK-1849](https://helsinkisolutionoffice.atlassian.net/browse/LINK-1849)

[LINK-1849]: https://helsinkisolutionoffice.atlassian.net/browse/LINK-1849?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


registrations/models.py
```diff
@@ -389,6 +392,7 @@ def can_be_edited_by(self, user):
             user.is_superuser
             or user.is_admin_of(self.publisher)
             or user.is_registration_admin_of(self.publisher)
+            or user.is_substitute_user_of(self.publisher)
```
jorilindell, line 395: Should this be `or user.is_substitute_user_of(self.registration_user_accesses)`?

jorilindell:
Could `/registration/?admin_user=true` request return also registrations where current user is substitute user. In that case we could display also those registrations in the registration table in Linked Events UI.

Maybe you could modify `get_editable_events_for_registration` method in `/helevents/models.py` module a little bit to achieve that. e.g. 
```
def get_editable_events_for_registration(self, queryset):
        publishers = (
            self.get_admin_organizations_and_descendants()
            | self.get_registration_admin_organizations_and_descendants()
        )
        return queryset.filter(publisher__in=publishers) | queryset.filter(
            registration__registration_user_accesses__email=self.email,
            registration__registration_user_accesses__is_substitute_user=True,
        )
 ```

**Comment by jorilindell** [comment]
Should `only_admin_visible_fields` fields in `RegistrationSerializer` also be returned to substitute users?

You could achieve this e.g. by updating `are_only_admin_visible_fields_allowed` method to 
```
def are_only_admin_visible_fields_allowed(self, obj):
        user = self.user

        return not user.is_anonymous and (
            obj.publisher.tree_id in self.admin_tree_ids
            or obj.publisher.tree_id in self.registration_admin_tree_ids
            or user.is_substitute_user_of(obj.registration_user_accesses)
        )
```

**Comment by jorilindell** [comment]
Could UserSerializer return `is_substitute_user` or similar boolean value which display is current user substitute user in any registration. 
One way to achieve this is to add 
```
@property
    def is_substitute_user(self):
        """Check if the user is substitute user in any registration"""

        return RegistrationUserAccess.objects.filter(
            email=self.email, is_substitute_user=True
        ).exists()
```
property to `UserModelPermissionMixin` and add `is_substitute_user` field in `UserSerializer`

In that way we could show registration tab to users who are substitute user in any registration but are not admin users in any organization

File: registrations/filters.py [hunk 2]
```diff
@@ -62,6 +62,13 @@ def filter_queryset(self, queryset):
             Q(registration__event__publisher__in=self._user_admin_organizations)
             & Q(registration__created_by=user)
         )
+
+        if user.email and user.email.endswith("@hel.fi"):
```
jorilindell, line 66: Maybe required domain ("hel.fi") could be defined in the environment variables. That would make local testing and development easier. At least I don't have hel.fi email account any more.

registrations/filters.py
```diff
@@ -88,8 +95,16 @@ def filter_registration(self, queryset, name, value: list[str]):
         ) | (
             Q(event__publisher__in=self._user_admin_organizations) & Q(created_by=user)
         )
+
+        if user.email and user.email.endswith("@hel.fi"):
```
jorilindell, line 99: Maybe required domain could be defined in the environment variables.

registrations/serializers.py
```diff
@@ -457,9 +457,25 @@ class RegistrationUserAccessCreateSerializer(serializers.ModelSerializer):
         allow_null=True,
     )
 
+    def validate(self, data):
+        validated_data = super().validate(data)
+
+        errors = {}
+
+        email = validated_data["email"]
+        if validated_data.get("is_substitute_user") and not email.endswith("@hel.fi"):
```
jorilindell, line 466: Maybe required domain could be defined in the environment variables? In that case the error message should use that variable as well.

jorilindell:
Link to attendance list link should be different for the registration user access with or without `is_substitute_user` value.
The link should be 
- `{{linked_registrations_ui_url}}/{{linked_registrations_ui_locale}}/registration/{{registration_id}}/attendance-list/` if `is_substitute_user` is `False`
- `{{linked_events_ui_url}}/{{linked_events_ui_locale}}/registrations/{{registration_id}}/attendance-list/` if `is_substitute_user` is `True`

Note that in the route it's **registration** in the Linked Registration UI and **registrations** in the Linked Events UI

registrations/models.py
harriris-vincit: Indeed it should; this is now fixed.

harriris-vincit:
> Could `/registration/?admin_user=true` request return also registrations where current user is substitute user. In that case we could display also those registrations in the registration table in Linked Events UI.
> 
> Maybe you could modify `get_editable_events_for_registration` method in `/helevents/models.py` module a little bit to achieve that. e.g.
> 
> ```
> def get_editable_events_for_registration(self, queryset):
>         publishers = (
>             self.get_admin_organizations_and_descendants()
>             | self.get_registration_admin_organizations_and_descendants()
>         )
>         return queryset.filter(publisher__in=publishers) | queryset.filter(
>             registration__registration_user_accesses__email=self.email,
>             registration__registration_user_accesses__is_substitute_user=True,
>         )
> ```

Implemented this suggestion so that the additional filtering is done if the user's email has one of the accepted domains.

harriris-vincit:
> Should `only_admin_visible_fields` fields in `RegistrationSerializer` also be returned to substitute users?
> 
> You could achieve this e.g. by updating `are_only_admin_visible_fields_allowed` method to
> 
> ```
> def are_only_admin_visible_fields_allowed(self, obj):
>         user = self.user
> 
>         return not user.is_anonymous and (
>             obj.publisher.tree_id in self.admin_tree_ids
>             or obj.publisher.tree_id in self.registration_admin_tree_ids
>             or user.is_substitute_user_of(obj.registration_user_accesses)
>         )
> ```

I believe it should as the substitute user is essentially a temporary admin. Implemented this suggestion now.

harriris-vincit:
> Could UserSerializer return `is_substitute_user` or similar boolean value which display is current user substitute user in any registration. One way to achieve this is to add
> 
> ```
> @property
>     def is_substitute_user(self):
>         """Check if the user is substitute user in any registration"""
> 
>         return RegistrationUserAccess.objects.filter(
>             email=self.email, is_substitute_user=True
>         ).exists()
> ```
> 
> property to `UserModelPermissionMixin` and add `is_substitute_user` field in `UserSerializer`
> 
> In that way we could show registration tab to users who are substitute user in any registration but are not admin users in any organization

Implemented now.

registrations/filters.py
harriris-vincit: I added an environment variable and a setting that accepts a list as the value so multiple accepted domains are now supported. Default element in the list will be "@hel.fi". There is also a shared util function now that checks a given email against the acceptable domains (`has_allowed_substitute_user_email_domain`).

harriris-vincit:
> Link to attendance list link should be different for the registration user access with or without `is_substitute_user` value. The link should be
> 
>     * `{{linked_registrations_ui_url}}/{{linked_registrations_ui_locale}}/registration/{{registration_id}}/attendance-list/` if `is_substitute_user` is `False`
> 
>     * `{{linked_events_ui_url}}/{{linked_events_ui_locale}}/registrations/{{registration_id}}/attendance-list/` if `is_substitute_user` is `True`
> 
> 
> Note that in the route it's **registration** in the Linked Registration UI and **registrations** in the Linked Events UI

@jorilindell This change was also implemented (took note of the singular-plural difference as well).

jorilindell:
At the moment it's not possible to update registrations by a substitute user if the user doesn't have any organization.

I think that 
```
if request.method in ("PUT", "PATCH", "DELETE"):
            (
                __,
                user_organization,
            ) = self.user_data_source_and_organization_from_request(request)
            return bool(user_organization)

        return request.method in permissions.SAFE_METHODS
``` 
in `CanAccessRegistration` prevents that.

harriris-vincit:
> At the moment it's not possible to update registrations by a substitute user if the user doesn't have any organization.
> 
> I think that
> 
> ```
> if request.method in ("PUT", "PATCH", "DELETE"):
>             (
>                 __,
>                 user_organization,
>             ) = self.user_data_source_and_organization_from_request(request)
>             return bool(user_organization)
> 
>         return request.method in permissions.SAFE_METHODS
> ```
> 
> in `CanAccessRegistration` prevents that.
> 
> Maybe somethink like
> 
> ```
> if request.method in ("PUT", "PATCH", "DELETE"):
>             (
>                 __,
>                 user_organization,
>             ) = self.user_data_source_and_organization_from_request(request)
>             return bool(user_organization) or request.user.is_substitute_user
> ```
> 
> could solve the issue

This one was just fixed

jorilindell (approved):
Looks very good to me

PR #843, opened 2024-01-25
LINK-1856 | One-time Access Code for Contact Person

harriris-vincit:
### Description
Makes it possible for strongly identified contact persons to read, update and delete their own signups and signup groups. These actions are permitted if the contact person uses their one-time access code and their user account is linked to the `SignUpContactPerson`.
### Closes
[LINK-1856](https://helsinkisolutionoffice.atlassian.net/browse/LINK-1856)

[LINK-1856]: https://helsinkisolutionoffice.atlassian.net/browse/LINK-1856?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jorilindell:
Could API return info does current user has permissions to edit signup or signup group. At the moment UI allows user to edit signup or signup group only if instance is create by the user (`is_created_by_current_user` field). But after this change user should be able to edit signup or signup group also if user is contact person.

API could return e.g. `can_be_edited_by_current_user` field for this purpose.
```
can_be_edited_by_current_user = serializers.SerializerMethodField()

    def get_can_be_edited_by_current_user(self, obj):
        if not (request := self.context.get("request")):
            return False
        
        return obj.can_be_edited_by(
            request.user, access_code=request.GET.get("access_code")
        )
```

In case that `can_be_edited_by_current_user` is too heavy in the performance wise API could return `is_contact_person` field and UI could allow editing if `is_created_by_current_user` or `is_contact_person` field is True

harriris-vincit:
Converted back to draft to implement a more secure one-time access code.

registrations/tests/test_models.py
```diff
@@ -475,6 +476,56 @@ def test_signup_group_registration(self):
...
+
+        # Email address is missing.
+        self.assertIsNone(self.contact_person.email)
+        self.assertFalse(self.contact_person.can_create_access_code(self.user))
+
+        # Email address is set and user's email doesn't match it or user doesn't already have
+        # editing and deletion rights for the signup.
+        self.contact_person.email = "test2@test.com"
+        self.contact_person.save(update_fields=["email"])
+        self.assertNotEquals(self.contact_person.email, self.user.email)
```
nicobav, line 490: these are quite hard to follow whats going on. Perhaps restructure/split or add  inline comments whats happening 🤔

registrations/tests/test_signup_group_put.py
```diff
@@ -182,6 +183,112 @@ def test_registration_admin_can_update_signup_group(
...
+    ) as mocked:
+        assert_update_signup_group(
+            api_client,
+            signup_group.id,
+            signup_group_data,
+        )
+        assert mocked.called is True
+
+    signup_group.refresh_from_db()
+    del signup_group.extra_info
```
nicobav, line 229: should this be here? and if should then why?

nicobav (changes requested):

registrations/models.py
```diff
@@ -433,6 +435,13 @@ def data_source(self):
     def publisher(self):
         return self.registration.publisher
 
+    @cached_property
+    def actual_contact_person(self):
+        if getattr(self, "signup_group_id", None):
```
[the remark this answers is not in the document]
harriris-vincit: A signup that belongs to a group shouldn't have a contact person of their own as a group of signups always have a single shared contact person. I like the walrus operator suggestion so will implement it (though will do `getattr(self, "signup_group", None)` instead)

registrations/tests/test_models.py
harriris-vincit: Restructured to individual tests

registrations/tests/test_signup_group_put.py
```diff
@@ -182,6 +183,112 @@ def test_registration_admin_can_update_signup_group(
...
+        response = update_signup_group(
+            api_client,
+            signup_group.id,
+            signup_group_data,
+        )
+        assert mocked.called is True
+    assert response.status_code == status.HTTP_403_FORBIDDEN
+
+    signup_group.refresh_from_db()
+    del signup_group.extra_info
```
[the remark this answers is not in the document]
harriris-vincit: This is to refresh the cached property to ensure the correct value is being checked

registrations/models.py
```diff
@@ -858,6 +877,39 @@ class SignUpContactPerson(SerializableMixin):
         },
     )
 
+    def can_create_access_code(self, current_user: User) -> bool:
+        if not self.email:
+            return False
+
+        if self.email == current_user.email:
+            signup_or_group = self.signup_group if self.signup_group_id else self.signup
+            if signup_or_group.can_be_edited_by(
```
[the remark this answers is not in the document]
harriris-vincit: Refactored to a single if condition

registrations/models.py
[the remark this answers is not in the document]
harriris-vincit: Implemented the walrus operator change now

registrations/permissions.py
```diff
@@ -7,6 +9,30 @@
 from registrations.models import PriceGroup, Registration, SignUp, SignUpGroup
 
 
+def _has_signup_admin_permissions(request: Request, obj: Union[SignUp, SignUpGroup]):
```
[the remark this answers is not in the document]
harriris-vincit: Created a mixin with these two functions as static methods

registrations/tests/test_signup_group_put.py
harriris-vincit: This is to refresh the cached property and ensure that the correct value is being checked

jorilindell (approved):
Looks very good to me

registrations/tests/test_signup_group_put.py
nicobav: Oh right! Should we still invalidate the property in the logic level if that is something of a feature we want to have. Instead of invalidating it here in the test just to get the test working 🤔 
(btw I ran the test locally and it passed without the cache but  I guess since cache it would flicker in some cases. )

nicobav (approved):
LGTM 👍 

Left a comment in the previous discussion but leave it up to you. Awesome 📦

registrations/tests/test_signup_group_put.py
harriris-vincit: It could be useful as a general feature to add the invalidation to the `save` method of the objects that use these cached properties or maybe add it to `refresh_from_db`. In the API views we generally don't mind if we have a cached value as long as we are aware of this in the rest of the logic, and `get_object`will fetch a fresh instance without the cached value(s) AFAIK. So perhaps for this case it's OK to invalidate it in the test as it is only to check that the DB value has changed correctly?

nicobav: Sounds reasonable 🤝

PR #844, opened 2024-01-25
LINK-1382 | Vantaa AD user is not an external user 

charn:
Add Vantaa AD as non-external authentication method. Also change the default value of `HELUSERS_BACK_CHANNEL_LOGOUT_ENABLED` since it's enabled by default on all environments.


danipran (approved):
Check commitlint's error message

PR #848, opened 2024-01-31
feat: support images in Espoo importer

voneiden:
This does complicate the importer slightly as Image object does not have origin_id field and a decision was made to not add one. Support has been added to depend on custom field for mapping, although the default behaviour is still:

 (hel) origin_id <-> (espoo) id

 For images we use simply

 (hel) url <-> (espoo) url

 It's important to note that url is not unique, so imported events may
 end up linking to same hel Image even if they link to different Images
 in the Espoo API.

Ref LINK-1871


nicobav (approved):
LOKTM! 🚀

PR #850, opened 2024-02-02
fix: whitelist web_store in the dockerignore file

harriris-vincit:
Fixes broken build due to the new `web_store` app not being mentioned in the `.dockerignore` file.


jorilindell (approved):
Great!

PR #862, opened 2024-02-27
fix: use fallback languages in ICS file generation

harriris-vincit:
### Description
If an event doesn't have a translated name in the language provided as an argument to the `create_event_ics_file_content` function, use a suitable fallback language in the function. The first language that has a translation for the event name will be used; otherwise, the first fallback option is used.
### Closes
[LINK-1910](https://helsinkisolutionoffice.atlassian.net/browse/LINK-1910)


terovirtanen:
**LINKEDEVENTS-API branch is deployed to platta: https://linkedevents-pr862.api.dev.hel.ninja** :rocket::rocket::rocket:

registrations/tests/test_utils.py
```diff
@@ -67,10 +73,72 @@ def test_create_ics_file_content_with_start_time_as_end_time():
...
+        ("sv", "Tapahtuman nimi", "fi"),
+        ("sv", None, "en"),
+        ("en", "Tapahtuman nimi", "fi"),
+        ("en", "Evenemangets namn", "sv"),
+        ("en", None, "fi"),
+    ],
+)
+@pytest.mark.freeze_time("2024-01-01")
+@pytest.mark.django_db
+def test_create_ics_file_using_fallback_language(
```
nicobav, line 105: There's a lot of logic 🖖 in this test. I'd suggest to split those in separate tests even if it would seem silly.

harriris-vincit: Split this into two tests.

nicobav (approved):

registrations/utils.py
```diff
@@ -104,6 +121,14 @@ def create_event_ics_file_content(event, language="fi"):
     cal.add("prodid", "-//linkedevents.hel.fi//NONSGML API//EN")
     cal.add("version", "2.0")
 
+    # If event name isn't translated for the given language,
+    # try to fall back to a language that has a translation.
+    if (
```
charn, line 126: I guess modeltranslation would have a builtin way to define fallback languages, but I'm guessing we don't want to add `fallback_languages` to `EventTranslationOptions`?

harriris-vincit: Thinking about this ICS generation function, here it could lead to `event.short_description` being shown in a different language compared to `event.name` (instead of having an empty `event.short_description` in case there is no translation for the same language). It's difficult to say for sure if there are similar risks in other use cases, but it's quite possible due to the global nature of the change.

harriris-vincit: So I tested this for a single event and unlike what I had stated in my previous comment, it seems to work just fine. If there is a field that has value in a fallback language and another field that has a value in another / different fallback language, the first fallback language seems to be used also for the other field (so no chance for having fields in different languages due to fallback differences). So perhaps we just use the built-in `fallback_languages`. Will push that change soon.

terovirtanen:
**LINKEDEVENTS-API branch is deployed to platta: https://linkedevents-pr862.api.dev.hel.ninja** :rocket::rocket::rocket:

PR #867, opened 2024-03-08
LINK-1926 | send invitation is is_substitute_user is changed

jorilindell:
## Description
Send invitation to registration user access email also when is_substitute_user values is changes. 
Move logic to send invitation to signals to avoid duplicate code in admin side and RegistrationViewSet

## Closes
[LINK-1926](https://helsinkisolutionoffice.atlassian.net/browse/LINK-1926)

[LINK-1926]: https://helsinkisolutionoffice.atlassian.net/browse/LINK-1926?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/registrations/tests/test_registration_put.py
+++ b/registrations/tests/test_registration_put.py
@@ -325,6 +325,7 @@ def test_send_email_to_new_registration_user_access(
     RegistrationUserAccessFactory(registration=registration, email="delete1@email.com")
     RegistrationUserAccessFactory(registration=registration, email="delete2@email.com")
     assert len(registration.registration_user_accesses.all()) == 2
+    mail.outbox.clear()
 
     registration_data = {
         "event": {"@id": get_event_url(registration.event.id)},
@@ -366,6 +367,7 @@ def test_email_is_not_sent_if_registration_user_access_email_is_not_updated(
         email=user_email,
         is_substitute_user=is_substitute_user,
     )
+    mail.outbox.clear()
 
     registration_data = {
         "event": {"@id": get_event_url(registration.event.id)},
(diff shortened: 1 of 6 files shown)


terovirtanen:
**LINKEDEVENTS-API branch is deployed to platta: https://linkedevents-pr867.api.dev.hel.ninja** :rocket::rocket::rocket:

registrations/signals.py
```diff
@@ -148,6 +149,42 @@ def registration_post_save(
...
+) -> None:
+    if not instance.pk:
+        return
+
+    old_instance = RegistrationUserAccess.objects.filter(pk=instance.pk).first()
+    if old_instance and (
+        old_instance.email != instance.email
+        or old_instance.is_substitute_user != instance.is_substitute_user
+    ):
+        instance.send_invitation()
```
harriris-vincit, line 170: The signals code here is overall fine, but I was thinking that since `pre_save` is called before actually saving, the actual saving might fail after the invitation has been sent. Perhaps both the `pre_save` and `post_save` could be done by overriding the `Model.save()` method:
```
def save(self, *args, **kwargs):
    if self.pk:
        old_instance = RegistrationUserAccess.objects.filter(pk=self.pk).first()
    else:
        old_istance = None
         
    super().save(*args, **kwargs)
     
    if not old_instance or old_instance.email != self.email or old_instance.is_substitute_user != self.is_substitute_user:
        self.send_invitation()
```

registrations/tests/test_registration_put.py
```diff
@@ -366,6 +366,8 @@ def test_email_is_not_sent_if_registration_user_access_email_is_not_updated(
         email=user_email,
         is_substitute_user=is_substitute_user,
     )
+    # An email is sent when registration user access is created
+    assert len(mail.outbox) == 1
```
tuomas777, line 370: In addition to these mail count asserts, I guess it would be a bit more robust to verify that the mail that was sent is indeed the correct one. Isn't there already `assert_invitation_email_is_sent()` for that, or is there some problem with it?

registrations/signals.py
jorilindell: Changed as suggested and added `@transaction.atomic` for the `save` method

registrations/tests/test_registration_put.py
jorilindell: The problem was that the desired email content was compered to a wrong email. Solved it by clearing mail.outbox before updating registration user access object.

tuomas777 (approved):
LGTM 👍 🐴

PR #910, opened 2024-05-10
LINK-1997 | Merchant improvements

harriris-vincit:
### Description
Makes the following changes:
- A new merchant is created in Talpa every time `WebStoreMerchant.paytrail_merchant_id` value is changed. This is because Talpa needs to create a new merchant secret mapping when the Paytrail merchant ID is changed and it is recommended by the Talpa devs to create a new merchant in that case.
- `WebStoreMerchant.url` is no longer editable by users and will default to `setttings.LINKED_EVENTS_UI_URL`. The URL is used when the "Return to service" button is used in the Talpa checkout UI and therefore it should always be the service's URL.
- `RegistrationWebStoreProductMapping` references merchant via `external_merchant_id` (= Talpa's merchant ID) to allow updating it when a merchant has been re-created in Talpa.
- `RegistrationWebStoreProductMapping` is updated or created when necessary when a payment is created for a signup or signup group. This ensures that the product mapping is up to date in case the merchant has been re-created in Talpa.
### Closes
[LINK-1997](https://helsinkisolutionoffice.atlassian.net/browse/LINK-1997)

[LINK-1997]: https://helsinkisolutionoffice.atlassian.net/browse/LINK-1997?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


terovirtanen:
**LINKEDEVENTS-API branch is deployed to platta: https://linkedevents-pr910.api.dev.hel.ninja** :rocket::rocket::rocket:

terovirtanen:
**LINKEDEVENTS-API branch is deployed to platta: https://linkedevents-pr910.api.dev.hel.ninja** :rocket::rocket::rocket:

terovirtanen:
**LINKEDEVENTS-API branch is deployed to platta: https://linkedevents-pr910.api.dev.hel.ninja** :rocket::rocket::rocket:

registrations/models.py
```diff
@@ -241,7 +241,10 @@ class WebStoreMerchant(CreatedModifiedBaseModel):
         verbose_name=_("Phone number"),
         max_length=18,
     )
-    url = models.URLField(verbose_name=_("URL"))
+    url = models.URLField(
```
nicobav, line 244: Do we have any use of this field? Since its always set to `LINKED_EVENTS_UI_URL` it then could (and should) be removed.

harriris-vincit: If the value of `settings.LINKED_EVENTS_UI_URL` is changed, then without this field we would need to check against the Talpa API if the value has changed to determine if a merchant needs to be updated. When saving to the local DB, we can check against that without an API request.

nicobav (approved):
LGTM! 🚀

PR #929, opened 2024-06-05
LINK-2080 | Add lang parameter to checkout URLs

harriris-vincit:
### Description
Adds a "lang" GET parameter after the Talpa checkout URLs.
### Closes
[LINK-2080](https://helsinkisolutionoffice.atlassian.net/browse/LINK-2080)

[LINK-2080]: https://helsinkisolutionoffice.atlassian.net/browse/LINK-2080?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/registrations/utils.py
+++ b/registrations/utils.py
@@ -168,6 +168,13 @@ def move_waitlisted_to_attending(registration, count: int):
             registration.move_first_waitlisted_to_attending(first_on_list=first_on_list)
 
 
+def get_checkout_url_with_lang_param(checkout_url: str, lang_code: str) -> str:
+    if "?" in checkout_url:
+        return f"{checkout_url}&lang={lang_code}"
+    else:
+        return f"{checkout_url}?lang={lang_code}"
+
+
 def get_access_code_for_contact_person(contact_person, user):
     if not contact_person:
         return None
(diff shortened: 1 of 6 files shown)


terovirtanen:
**LINKEDEVENTS-API branch is deployed to platta: https://linkedevents-pr929.api.dev.hel.ninja** :rocket::rocket::rocket:

registrations/utils.py
```diff
@@ -168,6 +168,15 @@ def move_waitlisted_to_attending(registration, count: int):
             registration.move_first_waitlisted_to_attending(first_on_list=first_on_list)
 
 
+def get_checkout_url_with_lang_param(checkout_url, lang_code):
+    if "?" in checkout_url:
+        checkout_url += f"&lang={lang_code}"
+    else:
+        checkout_url += f"?lang={lang_code}"
+
+    return checkout_url
```
danipran, line 177: Missing unit tests.

registrations/utils.py
```diff
@@ -168,6 +168,15 @@ def move_waitlisted_to_attending(registration, count: int):
             registration.move_first_waitlisted_to_attending(first_on_list=first_on_list)
 
 
+def get_checkout_url_with_lang_param(checkout_url, lang_code):
```
danipran, line 171: Type hints would be nice since this (most likely) assumes that the URL is `str`. Not sure about `lang_code`.
```suggestion
def get_checkout_url_with_lang_param(checkout_url: str, lang_code) -> str:
```

registrations/utils.py
danipran, line 177: Currently mutates `checkout_url`, I think 🤔 
```suggestion
    if "?" in checkout_url:
        return f"{checkout_url}&lang={lang_code}"
    else:
        return f"{checkout_url}?lang={lang_code}"
```

danipran (changes requested):

registrations/utils.py
harriris-vincit: Apparently the result of `+=` is a new string, but can have the same memory address which is due to optimizations in CPython. But I still changed this as suggested to not leave anything based on assumptions.

harriris-vincit: Added a proper unit test.

danipran (approved):
LGTM 👍

PR #939, opened 2024-06-14
LINK-2096 | Use GeoModelSerializer in analytics API

harriris-vincit:
### Description
Ensures that coordinates are returned using `DEFAULT_SRS` from the data analytics API.
### Closes
LINK-2096


jorilindell (approved):
Very good

PR #952, opened 2024-06-27
fix: allow to patch presence_status regardless of price groups

harriris-vincit:
### Description
Due to faulty validation for the signup price group selection, `presence_status` cannot be patched for a signup if the registration has price group selections. This PR fixes the validation.
### Closes
LINK-2105


charn (approved):
LGTM! :+1: :shipit:

PR #957, opened 2024-07-30
fix(events): fix error with x_ongoing_OR_set

danipran:
If no events were found, the filter caused a crash (TypeError, set.intersection needs values). This fixes the issue.

This also fixes another issue; this error was encountered with the query param `local_ongoing_OR_set=xxxxxxx`, which shouldn't even trigger the filter (according to the implementation) since it requires an ascending number at the end, starting from 1. E.g. `blabla_OR_set1=x,blabla_OR_set2=y` is valid.

There's also something terribly wrong with the regex that the ongoing OR set filters use. Apparently, this was encountered in https://github.com/City-of-Helsinki/linkedevents/commit/f581b16a2b9df1bf0f21a404f29e87a4766bee40, since there are tests marked as xfail for the function.

Added some test cases for the filter (there were none before), but marked them as xfail because the filter doesn't seem to work as described in documentation.

### DIFF
--- a/events/api.py
+++ b/events/api.py
@@ -2946,7 +2946,7 @@ def _filter_event_queryset(queryset, params, srs=None):  # noqa: C901
         )
         queryset = queryset.filter(Exists(kw_qs) | Exists(audience_qs))
 
-    if "local_ongoing_OR_set" in "".join(params):
+    if "local_ongoing_OR_set1" in "".join(params):
         count = 1
         all_ids = []
         while f"local_ongoing_OR_set{count}" in params:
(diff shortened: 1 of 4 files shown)


terovirtanen:
**LINKEDEVENTS-API branch is deployed to platta: https://linkedevents-pr957.api.dev.hel.ninja** :rocket::rocket::rocket:

danipran:
Unfortunately, the PR environment doesn't quite work with this PR due to cache issues (doesn't have a proper cache and the cache needs to be populated in order to use the x_ongoing_OR_set filters). Should work locally if local Redis is used.

charn (approved):
Nice gradual improvement even though it doesn't fully fix everything wrong with the filter. :+1:

LGTM! :+1: :shipit:

PR #966, opened 2024-08-23
fix: GDPR disable feature

voneiden:
The helsinki_gdpr implementation expects that the delete function may return an ErrorResponse object. Returning a Response object was actually equal to returning None, and it did not prevent the deletion from proceeding in Profile and Keycloak.

Refs: LINK-2116

### DIFF
--- a/helevents/utils.py
+++ b/helevents/utils.py
@@ -1,7 +1,8 @@
+from typing import Optional
+
 from django.conf import settings
 from django.contrib.auth import get_user_model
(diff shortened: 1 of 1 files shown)


terovirtanen:
**LINKEDEVENTS-API branch is deployed to platta: https://linkedevents-pr966.api.dev.hel.ninja** :rocket::rocket::rocket:

charn (approved):
LGTM! :+1: :shipit:

PR #969, opened 2024-08-27
LINK-2130 | Faster registration and GDPR delete transaction tests

harriris-vincit:
### Description
Make some tests in the `registrations` and `helevents` apps faster by using `django_capture_on_commit_callbacks` instead of `django_db(transaction=True)`.
### Closes
[LINK-2130](https://helsinkisolutionoffice.atlassian.net/browse/LINK-2130)

[LINK-2130]: https://helsinkisolutionoffice.atlassian.net/browse/LINK-2130?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/registrations/tests/test_registration_put.py
+++ b/registrations/tests/test_registration_put.py
@@ -1364,27 +1364,31 @@ def test_update_registration_with_product_accounting_api_exception(
     assert RegistrationWebStoreProductMapping.objects.count() == 0
 
 
-@pytest.mark.django_db(transaction=True)
+@pytest.mark.django_db
 def test_update_registration_increase_attendee_capacity_move_signups_to_attending(
-    api_client, event
+    api_client, event, django_capture_on_commit_callbacks
 ):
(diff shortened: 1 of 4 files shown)


registrations/tests/test_registration_put.py
```diff
@@ -1,4 +1,5 @@
 from decimal import Decimal
+from distutils.util import execute
```
charn, line 2: I believe this import can be removed. :point_up:

charn (approved):
Noticed one redundant import, but otherwise LGTM! :+1: :shipit:

nicobav (approved):
LGTM 👍

registrations/tests/test_registration_put.py
harriris-vincit: Very much so - now removed.

PR #974, opened 2024-09-03
LINK-2052 | feat(espoo-importer): don't import older events

nicobav:
Espoo event importer has run into timeouts quite often recently due to the large data mass it is querying from the Espoo LE api. To fix the above, this introduces filtering the incoming (espoo le api) data by using `start` filter which can be assigned to specific time through settings.

Refs LINK-2052


terovirtanen:
**LINKEDEVENTS-API branch is deployed to platta: https://linkedevents-pr974.api.dev.hel.ninja** :rocket::rocket::rocket:

events/importer/espoo.py
```diff
@@ -356,8 +362,13 @@ def _import_origin_objs(
     def origin_id(instance):
         return getattr(instance, instance_id_field)
 
+    queryset = model.objects.filter(data_source=data_source)
+
+    if model == Event:
+        queryset = queryset.filter(start_time__gte=EVENT_START)
```
charn, line 368: Just a theoretical question, but if the `end_date >= datetime.date.today()` then that event would not be updated?

nicobav: the `end_time` is irrelevant in the solution. Only `start_time` matter in this context so it will be updated if the `start_time` is greater or equal to the specified `EVENT_START`. If  `end_time >= datetime.date.today()` then the `start_time` would most likely also be and if start would be before `EVENT_START` it would not be updated regardless of the end time.

terovirtanen:
**LINKEDEVENTS-API branch is deployed to platta: https://linkedevents-pr974.api.dev.hel.ninja** :rocket::rocket::rocket:

PR #982, opened 2024-09-10
LINK-1463: Only superusers can add or remove organization users

harriris-vincit:
### Description
Allow only superusers to add or remove organizations' users.
### Closes
[LINK-1463](https://helsinkisolutionoffice.atlassian.net/browse/LINK-1463)

[LINK-1463]: https://helsinkisolutionoffice.atlassian.net/browse/LINK-1463?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/events/permissions.py
+++ b/events/permissions.py
@@ -159,13 +159,46 @@ class DataSourceResourceEditPermission(
         return True
 
 
-class DataSourceOrganizationEditPermission(BasePermission):
+class OrganizationEditPermission(BasePermission):
+    @staticmethod
+    def _is_editing_web_store_merchants_or_accounts(request_data):
+        return settings.WEB_STORE_INTEGRATION_ENABLED and (
+            request_data.get("web_store_merchants")
+            or request_data.get("web_store_accounts")
+        )
+
+    @staticmethod
+    def _is_editing_organization_users(request_data, obj=None):
+        user_keys = (
+            "admin_users",
+            "registration_admin_users",
+            "financial_admin_users",
+            "regular_users",
+        )
+
+        if not obj or not obj.pk:
+            # New organization => can only add users.
+            return any([request_data.get(key) for key in user_keys])
+
+        # Existing organization => can add or remove users
+        # => check if usernames are equal between request data and the organization's user relations.
+        return any(
+            [
+                key in request_data
+                and set(request_data[key])
+                != set(getattr(obj, key).values_list("username", flat=True))
+                for key in user_keys
+            ]
+        )
+
     def has_permission(self, request, view):
         if request.method == "POST":
-            if settings.WEB_STORE_INTEGRATION_ENABLED and (
-                request.data.get("web_store_merchants")
-                or request.data.get("web_store_accounts")
-            ):
+            if self._is_editing_organization_users(request.data):
+                # User must be a superuser to add users to a new organization.
+                return request.user.is_authenticated and request.user.is_superuser
+            elif self._is_editing_web_store_merchants_or_accounts(request.data):
+                # User must be a superuser or a financial admin with organization admin rights to
+                # add web store merchants or accounts to a new organization.
                 return request.user.is_authenticated and (
                     request.user.is_superuser
                     or request.user.admin_organizations.exists()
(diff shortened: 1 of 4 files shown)


events/permissions.py
```diff
@@ -113,6 +113,51 @@ def has_permission(self, request, view):
 )
 
 
+class OrganizationAddRemoveUsersPermission(BasePermission):
+    @staticmethod
+    def _has_added_or_removed_users(request_data, obj):
```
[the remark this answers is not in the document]
harriris-vincit: Basically the conditions here check if the username sets between the request data and the existing object's user relations are equal or not. If there are removals, they won't be equal anymore. Same for additions.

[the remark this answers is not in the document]
nicobav: ahh, yes 👍

events/permissions.py
```diff
@@ -113,6 +113,51 @@ def has_permission(self, request, view):
...
+
+        return True
+
+    def has_object_permission(self, request, view, obj):
+        if request.method in ("PUT", "PATCH") and self._has_added_or_removed_users(
+            request.data, obj
+        ):
+            return request.user.is_authenticated and request.user.is_superuser
+
+        return True
```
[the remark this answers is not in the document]
harriris-vincit: I combined this permission with `DataSourceOrganizationEditPermission` and since that existing permission didn't actually have anything to do with data sources, I renamed it to `OrganizationEditPermission`.

events/permissions.py
```diff
@@ -185,26 +230,26 @@ def has_object_permission(self, request: Request, view, obj):
...
+            elif self._is_editing_web_store_merchants_or_accounts(request.data):
+                # User must be a superuser or a financial admin with organization admin rights to
+                # edit an existing organization's web store merchants or accounts.
+                data_keys = request.data.keys()
+
+                if len(data_keys) > 1:
+                    return (
+                        request.user.is_superuser
+                        or request.user.is_financial_admin_of(obj)
+                        and utils.organization_can_be_edited_by(obj, request.user)
```
nicobav, line 246: clarifying to myself; so if user is modifying both; the merchants and accounts, the user needs to have organization admin rights AND the financial admin rights?
But when modifying only other (merch or account) then financial admin is sufficient?

events/tests/test_organization_api.py
```diff
@@ -183,34 +189,47 @@ def assert_account_fields_exist(data):
...
+
+def assert_cannot_see_organization_users(data):
+    for field in organization_user_fields:
+        assert data.get(field) is None
+
+
 @pytest.mark.django_db
-def test_admin_user_can_see_organization_users(organization, user, user_api_client):
-    organization.regular_users.add(user)
+def test_superuser_or_admin_can_see_organization_users(api_client, organization):
```
nicobav, line 203: Looks like this is testing only `super_user` not the organization admin. I suggest changing the name of the test to `test_super_user_can_see_organization_users`

nicobav (changes requested):

events/permissions.py
harriris-vincit: The logic here should go like:

- if editing merchants and/or accounts with other organization data given in the request, the user must be either a superuser or a financial admin with also organization editing (org admin) rights
- if editing only merchants and/or accounts (without other organization data), the user must be either a superuser or a financial admin

But actually, this check doesn't appear to work properly because if the user gives both merchants and accounts, then `if len(data_keys) > 1` will be `True` and the organization admin rights are required unnecessarily :thinking: 

Fix needed here, BRB.

events/tests/test_organization_api.py
harriris-vincit: Still forgot a wrong name for this after splitting the test cases -> fixing

events/permissions.py
harriris-vincit: Now fixed

events/tests/test_organization_api.py
harriris-vincit: Fixed

events/permissions.py
```diff
@@ -159,13 +159,46 @@ def has_object_permission(self, request: Request, view, obj):
         return True
 
 
-class DataSourceOrganizationEditPermission(BasePermission):
+class OrganizationEditPermission(BasePermission):
+    @staticmethod
+    def _is_editing_web_store_merchants_or_accounts(request_data):
+        return settings.WEB_STORE_INTEGRATION_ENABLED and (
+            request_data.get("web_store_merchants")
+            or request_data.get("web_store_accounts")
```
nicobav, line 167: 👍

events/permissions.py
```diff
@@ -159,13 +159,46 @@ def has_object_permission(self, request: Request, view, obj):
...
+            "regular_users",
+        )
+
+        if not obj or not obj.pk:
+            # New organization => can only add users.
+            return any([request_data.get(key) for key in user_keys])
+
+        # Existing organization => can add or remove users
+        # => check if usernames are equal between request data and the organization's user relations.
+        return any(
```
nicobav, line 185: Nice improvement 👍

nicobav (approved):
Looking awesome 🚀

PR #997, opened 2024-10-04
LINK-2187 | fix(registration): order signups by id

nicobav:
Order signups by the id in the signup endpoint to achieve correct ordering in the UI for the attendee list and waitlist listing.
Reason for use the ID for ordering is that the waitlist logic uses the id for ordering waitlist queue. Other solution would be to change it to be using created_time.

Refs LINK-2187

### DIFF
--- a/registrations/api.py
+++ b/registrations/api.py
@@ -582,8 +582,8 @@ class SignUpViewSet(
         ActionDependingBackend,
         filters.OrderingFilter,
     ]
-    ordering_fields = ("first_name", "last_name")
-    ordering = ("first_name", "last_name")
+    ordering_fields = ("id", "first_name", "last_name")
+    ordering = ("id",)
     filterset_class = SignUpFilter
     permission_classes = [CanAccessSignup]
 
--- a/registrations/tests/test_signup_get.py
+++ b/registrations/tests/test_signup_get.py
@@ -802,6 +802,83 @@ def test_signup_list_assert_attendee_status_filter(api_client, registration):
 
 
 @pytest.mark.django_db
+def test_signup_list_orders_by_id_by_default(api_client, registration):
+    user = create_user_by_role("registration_admin", registration.publisher)
+    api_client.force_authenticate(user)
+
+    signup_first = SignUpFactory(
+        registration=registration,
+        first_name="The First",
+        last_name="Attendee",
+        attendee_status=SignUp.AttendeeStatus.ATTENDING,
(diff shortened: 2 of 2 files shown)


registrations/api.py
```diff
@@ -583,7 +583,7 @@ class SignUpViewSet(
         filters.OrderingFilter,
     ]
     ordering_fields = ("first_name", "last_name")
-    ordering = ("first_name", "last_name")
+    ordering = ("id",)
```
voneiden, line 586: In previous uses of `ordering` the field value(s) have always been included in `ordering_fields` and I think that's preferable for predictable behaviour. Is the deviation intentional? 

With the current approach `ordering=id` would seemingly work as it's the default `ordering`,  `ordering=-id` wouldn't work and `ordering=id,first_name` would actually be equivalent to `ordering=first_name` since I believe `OrderingFilter` just drops tokens that are not in ordering_fields.

voneiden (changes requested):

registrations/api.py
nicobav: Very good point. I think for now the UI does not use ordering request params to get the signups so this would work indeed but perhaps I'll add those first_name and last_name back to avoid the confusion you mentioned above.

terovirtanen:
**LINKEDEVENTS-API branch is deployed to platta: https://linkedevents-pr997.api.dev.hel.ninja** :rocket::rocket::rocket:

registrations/api.py
nicobav: I changed it to have the names and added tests for sort

terovirtanen:
**LINKEDEVENTS-API branch is deployed to platta: https://linkedevents-pr997.api.dev.hel.ninja** :rocket::rocket::rocket:

registrations/api.py
tuomas777: hmm, shouldn't this rather be 
```python
    ordering_fields = ("id", "first_name", "last_name")
    ordering = ("id",)
```

voneiden: > hmm, shouldn't this rather be
> 
> ```python
>     ordering_fields = ("id", "first_name", "last_name")
>     ordering = ("id",)
> ```

:point_up: yeah, I meant this

nicobav: Thank you for clarification. Once again updated.

tuomas777 (approved):
LGTM 👍 🦑

PR #999, opened 2024-10-09
fix: remove matko importer

voneiden:
matko importer has not been receiving new data for many years and the time has come to remove it.

Leaving the existing conversion logic from matko -> tprek in place

refs: LINK-1651


tuomas777 (approved):
Yes please 👍

PR #1023, opened 2025-01-10
test(registrations): fix flaky test

danipran:
`test_price_group_list_ordering` failed in certain circumstances, e.g. `pytest -k order`. This PR fixes that issue.

PriceGroupViewSet orders the price groups by `description`, `description` is a translated field, and `PriceGroupFactory` overrides all languages with `"<lang> Price Group <seq>"`, where `<lang>` is the language code and the `<seq>` is a growing sequence of numbers. In certain circumstances, this causes the order to become something unexpected, e.g. with the descriptions `EN Price Group 9` and `EN Price Group 10`, the latter price groups would be placed first (because `"10"` > `"9"`).

Solved the issue by setting the translation directly and making sure that the correct translation is used.

Also refactored the test a bit. I'm not sure if I agree with the test using paging since we're testing whether the order of the price groups is correct or not and it adds unnecessary complexity, but decided to keep it as it is.

Refs: LINK-2210

### DIFF
--- a/registrations/tests/test_price_group_get.py
+++ b/registrations/tests/test_price_group_get.py
@@ -1,3 +1,4 @@
+import math
 from collections import Counter
 from typing import Optional, Union
 
@@ -25,13 +26,13 @@ def get_detail(api_client: APIClient, price_group_pk: Union[str, int]):
     return api_client.get(url)
 
 
-def get_list(api_client: APIClient, query: Optional[str] = None):
+def get_list(api_client: APIClient, query: Optional[str] = None, data=None):
     url = reverse("pricegroup-list")
 
     if query:
         url = "%s?%s" % (url, query)
 
-    return api_client.get(url)
+    return api_client.get(url, data=data)
 
 
 def assert_get_detail(api_client: APIClient, price_group_pk: Union[str, int]):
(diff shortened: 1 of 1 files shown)


terovirtanen:
**LINKEDEVENTS-API branch is deployed to platta: https://linkedevents-pr1023.api.dev.hel.ninja** :rocket::rocket::rocket:

registrations/tests/test_price_group_get.py
```diff
@@ -243,31 +245,42 @@ def test_price_group_list_ordering(api_client):
...
     ]
 
-    page_size = 3
-    page_start = 0
-    page_end = page_size
-    for page in range(1, 4):
-        response = get_list(api_client, query=f"page={page}&page_size={page_size}")
+    # Price groups are ordered by description, so need to activate the correct
+    # language.
+    activate("fi")
```
tuomas777, line 263: Using `django.utils.translation.override()` would be better to make sure other tests aren't affected.

terovirtanen:
**LINKEDEVENTS-API branch is deployed to platta: https://linkedevents-pr1023.api.dev.hel.ninja** :rocket::rocket::rocket:

tuomas777 (approved):
Great 👍 🦜

PR #1046, opened 2025-03-21
test: remove db collation dep from test_price_group_list_ordering

voneiden:
This test would previously fail if DB collation was not fi_FI.utf8. Specific DB collation should not be a unit test requirement.


danipran (changes requested):
Should probably change these as well, just to be consistent: https://github.com/City-of-Helsinki/linkedevents/blob/f74b8beca0e4fa35e037505ea5e7749523a91376/registrations/tests/test_signup_get.py#L1071-L1085

PR #1053, opened 2025-04-07
LINK_2261 | Add weekday filter for events

nicobav:
Adds filter weekdays which results events occuring on given weekday. Filter supports giving multiple days like 1,3,5

Refs LINK-2261


mhieta (approved):
Great! 💯

PR #1055, opened 2025-04-15
LINK-2269: Rebuild event search index in batches

mhieta:
Use queryset batches in search index rebuild process to improve performance.
Add configurable batch size with default of 500.

This will result to significant performance improvement as rebuilding the search index with 500 batch-size the time with total of 47547 events is approximately 3 times faster than without batching.

Also overall memory consumption should be much lower due to the reduced bulk-update sizes.

Refs: LINK-2269


terovirtanen:
**LINKEDEVENTS-API branch is deployed to platta: https://linkedevents-pr1055.api.dev.hel.ninja** :rocket::rocket::rocket:

nicobav (approved):

PR #1065, opened 2025-05-13
LINK-2305 | Delete EventSearchIndex in batches

nicobav:
Delete the EventSearchIndex in batches to make it a bit faster.

Refs: LINK-2305


mhieta (approved):
Good! 💯

PR #1094, opened 2025-09-03
LINK-2373: Cache get_word_bases-function results

mhieta:
Optimize cache to use `get_word_bases` instead of `analyze_word`.
This should improve overall caching performance.

Also use `lru_cache` instead of `cache`.

Refs: LINK-2373

### DIFF
--- a/events/search_index/utils.py
+++ b/events/search_index/utils.py
@@ -1,6 +1,6 @@
 import logging
 import re
-from functools import cache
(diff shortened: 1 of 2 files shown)


terovirtanen:
**LINKEDEVENTS-API branch is deployed to platta: https://linkedevents-pr1094.api.dev.hel.ninja** :rocket::rocket::rocket:

PR #1101, opened 2025-09-12
LINK-2379: Add PostgreSQL index recommendations

mhieta:
Add more indexes to database according to Azure Postgres Index recommendations.

Event, Place and Keyword indexes are configured through model `Meta.indexes` for a better maintainability and Django integration.

Event `in_language` indexes are added through manual migration because it is the least disruptive option. Also migration via a separate through-model causes M2M-type issues and would need a separate multi-step data migration process for existing data because the migration cannot be applied directly to existing fields. Thus manual migration is much simpler alternative here.

Refs: LINK-2379


terovirtanen:
**LINKEDEVENTS-API branch is deployed to platta: https://linkedevents-pr1101.api.dev.hel.ninja** :rocket::rocket::rocket:

PR #1184, opened 2026-03-30
feat: use django-helsinki-health-endpoints for readiness and healthz

nicobav:
Make use of django-helsinki-health-endpoints library to publish readiness and healthz endpoints.
Add SENTRY_RELEASE to django settings from env for the library and make sure that pyproject.toml isn't excluded from the container. Those are requisite for the readiness data.

Refs: KEH-242


tuomas777 (approved):
LGTM 👍 🐊

PR #1195, opened 2026-04-21
Espoo importer orphan fixes + perf improvements

mhieta:
Related objects (images, places, keywords, organizations) were deleted when their events were removed by a previous import run but the objects themselves remained in the DB.

The old approach only protected objects linked to events outside the import window (start_time < EVENT_START). Objects linked to events that were once within the window but later deleted by event_syncher became orphaned and accumulated until ModelSyncher.finish() exceeded the 20% deletion safety threshold.

Fix: replace old_event_ids with surviving_event_ids, which covers only events that will exist after the import run completes:
- Events outside the import window (start_time < EVENT_START) never touched by event_syncher
- Events present in the current API response being (re-)imported in this run

Events inside the window that are absent from the API response
will be deleted by event_syncher, so related objects referenced
exclusively by them are now correctly cleaned up.

Refs: [LINK-2522](https://helsinkisolutionoffice.atlassian.net/browse/LINK-2522)

[LINK-2522]: https://helsinkisolutionoffice.atlassian.net/browse/LINK-2522?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/events/importer/espoo.py
+++ b/events/importer/espoo.py
@@ -8,8 +8,8 @@ from urllib.parse import urljoin, urlparse
 
 import requests
 from django.conf import settings
-from django.db import transaction
-from django.db.models import Model
+from django.db import models, transaction
+from django.db.models import Exists, Model, OuterRef
 from django.utils import timezone
 from django_orghierarchy.models import Organization
 from requests.adapters import HTTPAdapter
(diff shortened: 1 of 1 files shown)


events/importer/espoo.py
```diff
@@ -617,11 +633,20 @@ def import_events(self):
...
+            Organization.objects.filter(data_source=self.data_source)
+            .filter(
+                Exists(
+                    Event.objects.filter(
+                        publisher=OuterRef("pk"),
+                        id__in=surviving_event_ids,
+                    )
+                )
+            )
+            .iterator()
```
voneiden: The performance in prod for the old one is `1.293 ms` and the performance for the new one is `1.480 ms`. Both resolve to an identical plan.

events/importer/espoo.py
```diff
@@ -639,10 +664,20 @@ def import_events(self):
...
+            .filter(
+                Exists(
+                    Event.objects.filter(
+                        location=OuterRef("pk"),
+                        id__in=surviving_event_ids,
+                    )
+                )
+            )
+            .iterator()
+        ):
```
voneiden: Old: `1.363 ms`
New: `1.138 ms`

voneiden (approved):
The fix commit seems reasonable. The perf commit seems unnecessary, it refers to removing DISTINCT but there is no DISTINCT used in the first place. Checked the perf difference between old and new and it seems to be +- 0 and the old  query performance in prod is around 1ms already.

Approved, but consider merging without the perf commit.

voneiden:
Moving the perf improvement in separate PR. In hindsight I misunderstood the intent a bit.

PR #1239, opened 2026-08-12
Optimize indexing for multi-event create and update

tuomas777:
Suppress per-event search-index signals during multi-event POST and PUT requests to avoid the overhead of indexing each event separately, then update PostgreSQL and Haystack indexes in bulk.

Also, PostgreSQL search-vector refreshing is now performed as part of the search-index update operations instead of as a separate step.

Refs: LINK-2516

## Summary by CodeRabbit

* **New Features**
  * Improved search indexing for bulk event creation and updates.
  * Synchronized PostgreSQL and Haystack search indexes for more consistent results.
  * Added bulk rebuilding of event search indexes.
* **Bug Fixes**
  * Reduced duplicate and unnecessary indexing during bulk operations.
  * Limited search results to eligible public, non-deleted events.
  * Removed search entries when published events become drafts.
  * Improved updates to search data for affected events.

### DIFF
--- a/events/search_index/signals.py
+++ b/events/search_index/signals.py
@@ -1,4 +1,6 @@
 import logging
+from contextlib import contextmanager
+from contextvars import ContextVar
 
 from django.conf import settings
 from django.db.models.signals import m2m_changed, post_save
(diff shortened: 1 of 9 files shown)


events/search_index/signals.py
```diff
@@ -1,21 +1,61 @@
...
+
+def search_index_signals_enabled() -> bool:
+    return settings.EVENT_SEARCH_INDEX_SIGNALS_ENABLED and not (
+        search_index_signals_suppressed()
+    )
+
+
+def search_index_signals_suppressed() -> bool:
+    return _search_index_signals_suppressed.get()
+
```
voneiden, line 26: These names are somehow a bit tough to swallow, but I do realize they kinda represent a different thing. But it's still somehow funky that index signals can be disabled but not supressed. Not necessarily a problem though.

voneiden (approved):
Left one comment, but it's more semantic. Not a blocker for merging.

events/search_index/signals.py
tuomas777: Tweaked the names a bit.

PR #1247, opened 2026-08-24
Link-2600 | Sonar maintainability issues

jorilindell:
## Summary

Fixed 11 open Sonar maintainability issues across the events, registrations, and notifications modules to improve code quality and reduce technical debt. Added unit test coverage for event export command and eliminated N+1 query issues in payment notifications.

### events

- Remove wildcard imports from importer modules (S1192, S1066, S3457)
- Fix conditional logic in sync and import utilities (S1066, S1192)
- Simplify exception handling and variable selection in importers (S112, S8517)
- Add implicit concatenation escaping in import string handling (S5799)
- Fix regex character class and simplify complex method logic (S6035, S3776)
- Add missing exception logging for payment failures (S1186)
- Add unit tests for event export management command (197 lines of coverage)
- Add type hints and docstrings to serializers and API filters

### registrations

- Optimize payment notification query to avoid N+1 lookups
- Remove unused parameters and dead code branches (S1172, S5797)
- Fix redundant variable assignments (S8512)

### notifications

- Simplify management command exception handling (S1186)

### linkedevents

- Add type hints to schema utilities and settings

### multilingual_haystack

- Add docstring to search backend method

### docker

- Add quotes around Dockerfile environment variables (S6570)

### other

- Add sonar.tests property to exclude test files from analysis
- Update .gitignore for VS Code Copilot skills files

## Sonar

- S1066 — Redundant conditionals merged
- S1172 — Unused parameters removed
- S1186 — Empty exception handlers addressed with logging
- S1192 — Duplicate string literals consolidated
- S3457 — String formatting improved
- S3776 — Complex command logic reduced
- S5797 — Dead code branches removed
- S5799 — Implicit string concatenation escaped
- S6035 — Regex character classes fixed
- S8512 — Redundant assignments eliminated
- S8517 — Min/max selection simplified
- S6570 — Docker environment variables quoted

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved importer error logging with full diagnostic details.
  * Fixed paragraph-break detection and place replacement handling.
  * Preserved archive paths reliably during container builds.
  * Improved payment notification query efficiency.

* **Improvements**
  * Refined event import and export command selection and validation.
  * Improved Enkora event metadata and offer information.
  * Standardized category, address, translation, and configuration handling.
  * Added clearer setup and backend behavior documentation.

* **Tests**
  * Expanded coverage for event export command behavior and validation.

### DIFF
--- a/events/importer/utils.py
+++ b/events/importer/utils.py
@@ -134,6 +134,21 @@ def address_eq(a, b):
     return True
 
 
+def _get_single_place(name, data_source, include_deleted=False):
+    filters = {
+        "name__iexact": name,
+        "data_source": data_source,
+    }
+    if not include_deleted:
+        filters["deleted"] = False
+
+    matches = Place.objects.filter(**filters)
+    if matches.count() == 1:
+        return matches[0]
+
+    return None
+
+
 def replace_location(
     replace=None, from_source="tprek", by=None, by_source="matko", include_deleted=False
 ):
@@ -149,31 +164,15 @@ def replace_location(
     :return: Boolean that determines whether a new location was found for the hapless events
     """  # noqa: E501
     if not by:
-        replacements = Place.objects.filter(
-            name__iexact=replace.name, data_source=by_source, deleted=False
-        )
-        if replacements.count() == 1:
-            by = replacements[0]
-        else:
-            # the backup is to look for deleted locations and reinstate them
-            if include_deleted:
-                replacements = Place.objects.filter(
-                    name__iexact=replace.name, data_source=by_source
-                )
-                if replacements.count() == 1:
-                    by = replacements[0]
-                else:
-                    # no replacement whatsoever was found, this may result in an
-                    # exception
-                    return False
-            else:
-                return False
+        by = _get_single_place(replace.name, by_source)
+        if not by and include_deleted:
+            by = _get_single_place(replace.name, by_source, include_deleted=True)
+        if not by:
+            # no replacement whatsoever was found, this may result in an
+            # exception
+            return False
     if not replace:
-        to_be_replaced = Place.objects.filter(
-            name__iexact=by.name, data_source=from_source
-        )
-        if to_be_replaced.count() == 1:
-            replace = to_be_replaced[0]
+        replace = _get_single_place(by.name, from_source, include_deleted=True)
         # if no place to be replaced was found, it's alright, we might have a
         # brand new location here!
     by.deleted = False
--- a/events/importer/utils.py
+++ b/events/importer/utils.py
@@ -59,7 +59,7 @@ def separate_scripts(text, scripts):
     last_language = scripts[0]
     last_paragraph = ""
     for paragraph in paragraphs:
-        if paragraph in (r"</p><p>", r"</p>" r"\n", r"<p>", r"<br><br><br>"):
+        if paragraph in (r"</p><p>", r"</p>", r"\n", r"<p>", r"<br><br><br>"):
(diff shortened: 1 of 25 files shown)


events/importer/utils.py
```diff
@@ -59,7 +59,7 @@ def separate_scripts(text, scripts):
     last_language = scripts[0]
     last_paragraph = ""
     for paragraph in paragraphs:
-        if paragraph in (r"</p><p>", r"</p>" r"\n", r"<p>", r"<br><br><br>"):
+        if paragraph in (r"</p><p>", r"</p>", r"\n", r"<p>", r"<br><br><br>"):
```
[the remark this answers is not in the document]
jorilindell: Can this cause any real issues

events/tests/test_event_export_command.py
```diff
@@ -0,0 +1,182 @@
...
+from django.core.management.base import CommandError
+
+from events.exporter.base import Exporter, exporters, register_exporter
+from events.management.commands.event_export import Command
+
+
+@pytest.fixture(autouse=True)
+def cleanup_exporters():
+    """Clear exporters before each test to ensure test isolation."""
+    exporters.clear()
```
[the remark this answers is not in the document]
jorilindell: Is this still an issue

events/models.py
```diff
@@ -58,6 +58,7 @@
 PLACE_NAME_FORMAT = "location__name_%s"
 SHORT_DESCRIPTION_FORMAT = "short_description_%s"
 DESCRIPTION_FORMAT = "description_%s"
+ORGANIZATION_MODEL = "django_orghierarchy.Organization"
```
danipran: Can't you just import the model instead? E.g.
```py
from django_orghierarchy import Organization
...
foo = models.ForeignKey(Organization, ...)
```

registrations/models.py
```diff
@@ -92,6 +92,12 @@
     VatPercentage.VAT_0.value: "4U",
 }
 
+FIRST_NAME_LABEL = _("First name")
+LAST_NAME_LABEL = _("Last name")
+PHONE_NUMBER_LABEL = _("Phone number")
+STREET_ADDRESS_LABEL = _("Street address")
+ZIP_CODE_LABEL = _("ZIP code")
```
danipran: While this _should_ work, I'm not sure it's a good idea to use constants for lazy translations. Seems like it might get a bit flaky.

Not sure if it helps with maintainability/consistency either. Sure, it reduces duplication, but now the translations are either inside the model fields or they're in a constant. So you have to jump around. Also changing these values doesn't affect the choice labels only but also verbose names.

Lastly, this change doesn't seem to have too many benefits since these are translations only. These are not something that you'd actually need to use in logic or something, e.g. accessing keys, finding a certain value and so on, which would actually help with understanding the code and keeping things consistent.

Personally, I'd revert this change, but differing opinions are welcome since I'm not 100% sure on whether this is an improvement or not.

danipran (changes requested):

events/models.py
jorilindell: I tested the direct import as well, and it seems to work fine. I originally assumed the string reference was needed to avoid circular imports, but since that doesn’t appear to be an issue here, I agree that importing the model directly is cleaner.

registrations/models.py
jorilindell: I was thinking the same, and your comment confirms it. These changes are now reverted

danipran (approved):
LGTM! 👍

